PR #8, opened 2015-12-18
add a FileSystem DigitalObject::Base subclass, fixes HYACINTH-69

barmintor:
@elo2112 is this what you had in mind? If so will add the db migration in another PR.

### DIFF
--- a/app/models/digital_object/base.rb
+++ b/app/models/digital_object/base.rb
@@ -21,7 +21,7 @@ class DigitalObject::Base
   
   def require_subclass_override!; raise 'This method must be overridden by a subclass'; end
   
-  def initialize(*args)
+  def initialize
     raise 'The DigitalObject::Base class cannot be instantiated.  You can only instantiate subclasses like DigitalObject::Item' if self.class == DigitalObject::Base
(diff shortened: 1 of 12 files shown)


app/models/digital_object/base.rb
```diff
@@ -21,7 +21,7 @@ class DigitalObject::Base
   
   def require_subclass_override!; raise 'This method must be overridden by a subclass'; end
   
-  def initialize()
+  def initialize(*args)
```
elo2112: Why add (*args) if this method doesn't take any arguments?

barmintor: Because the call to super in the subclasses bombs if you pass the args hash.

elo2112: Do you mind if I just remove all cases where we pass the args hash?  This class and its subclasses aren't supposed to accept any args in the initializer.  I don't want the method signature to imply that they do.

barmintor: I'm a little suspicious of the subclass part, but you can explicitly call
super with an empty argument. Fine by me.

PR #31, opened 2016-11-11
Add specs for ezid metadata retrieval

fcd1:


elo2112:
Other than my comment about "return nil" vs "return", this looks good!  I'm okay with merging once we hear back from @barmintor.

lib/hyacinth/ezid/hyacinth_metadata_retrieval.rb
```diff
@@ -21,9 +21,10 @@ def initialize(digital_object_data_arg)
     # Following returns the title of an item. NOTE: if ever an item contains
     # multiple titles, it will only return the first one.
     def title
+      return unless @dfd.key? 'title'
```
barmintor: I'm going to send a PR to this branch with some recommendations.

elo2112:
Thanks for those additional commits, @barmintor.  Looks good.  Assigning ticket to @fcd1 for final approval, but I'm happy to see this merged into master at any time.

barmintor:
Starting to pull YARD in more generally, while good in theory, is a big step. But I do think it would be wise to start incorporating some of it (especially in situations like `@return`) while we decide how much we want to use. We can talk about it this week.

PR #32, opened 2016-11-12
Ezid pull

barmintor:
Demonstrate use of YARD to clarify void and nil returns
Demonstrate use of described_class in specs to facilitate rename and limit scope
Rename a class that sounds like a static procedure


barmintor:
@elo2112 @fcd1 if you install the yardstick gem:
`gem install yardstick`

you will be able to run YARD's documentation checks against a class- like RuboCop for docs. E.g.
`yardstick lib/hyacinth/ezid/hyacinth_metadata.rb`

PR #35, opened 2016-12-06
Changes to ApiSession, add Ezid concern to DigitalObject::Base

fcd1:


elo2112:
This looks good!  A couple of minor suggestions:

In app/models/concerns/digital_object/ezid.rb, I'd change:
`ezid_api_session.timeout`
to:
`ezid_api_session.timed_out?`

And along with that, Hyacinth::Ezid::ApiSession's attr_reader and instance variable :timeout/@timeout would become :timed_out/@timed_out, and you'd add a getter method with a question mark at the end.

The reason I'm suggesting this is that I've frequently seen a "timeout" property used to actually set or get the timeout duration for a session, rather than check to see if the session has timed out.  The question mark in the getter is also nice because it reads more like a sentence (i.e. "has this session timed out?"), which is something that you commonly see in Ruby.

My other suggestion is to concatenate your multiple logger.info calls in app/models/concerns/digital_object/ezid.rb, lines 26 through 30, into one call.  Reason is, if multiple concurrent requests are hitting Hyacinth at the same time, and those requests log anything, your logger messages will stay together.

elo2112:
Thanks for the update!  Looks great!

PR #38, opened 2016-12-20
Ezid publish integration

elo2112:


elo2112:
ClassLength now back to previous value of 250.

PR #47, opened 2017-03-21
Enable insertion of DOI (minus 'doi:' label) into MODS

fcd1:


app/models/concerns/digital_object/xml_datastream_rendering.rb
```diff
@@ -156,6 +156,9 @@ def value_for_field_name(field_name, df_data)
         created_at
       elsif field_name == '$updated_at'
         updated_at
+      elsif field_name == '$doi'
+        # slice off the "doi:" label
+        doi[4..-1] unless doi.nil?
```
elo2112: Would you mind changing line 161 to `doi.gsub(/^doi:/, '') unless doi.nil?`

That'll be more future-compatible when we remove the "doi:" prefix from values.

elo2112: Actually, I think this code should always return a value, even if doi is nil.  I'd recommend this instead:

`doi.present? ? doi.gsub(/^doi:/, '') : ''`

PR #52, opened 2017-08-03
Minting DOI before publish

cgalarza:
Minting DOI before publishing an object. Updating DOI after request from publish target returns.

### DIFF
--- a/app/models/digital_object/publish_target.rb
+++ b/app/models/digital_object/publish_target.rb
@@ -145,7 +145,11 @@ class DigitalObject::PublishTarget < DigitalObject::Base
       Authorization: "Token token=#{publish_target_field('api_key')}"
     )
     if do_ezid_update && response.code == 200 && response.headers[:location].present?
-      digital_object.do_ezid_publish(response.headers[:location])
+      # By this point, all records should have an ezid. Let's update the status of
+      # that ezid to :public, and send the latest published_object_url in case it changed.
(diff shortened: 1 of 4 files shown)


barmintor:
@fcd1 @elohanlon I want y'all to check in over here to understand the change we're proposing here. to prevent the double-publish, the DOI is being reserved/assigned to metadata - the object is published - the DOI is updated and made public. That's a big change, and I want to make sure we are all on the same page about it.

elohanlon:
@cgalarza and I talked about this change a while back and I like it!  👍

app/models/digital_object/publish_target.rb
```diff
@@ -145,7 +145,11 @@ def publish_digital_object(digital_object, do_ezid_update)
       Authorization: "Token token=#{publish_target_field('api_key')}"
     )
     if do_ezid_update && response.code == 200 && response.headers[:location].present?
-      digital_object.do_ezid_publish(response.headers[:location])
+      # By this point, all records should have an ezid. Let's update the status of
+      # that ezid to :public, and send the latest published_object_url in case it changed.
+      unless update_doi_metadata(response.headers[:location])
```
cgalarza: @elohanlon Should this line have been `digital_object.do_ezid_publish(response.headers[:location])`

elohanlon: @cgalarza You're right.  Fixing that now and will commit/push change very soon.

PR #53, opened 2017-08-22
Published Target Facet

cgalarza:
- Added byebug
- Can facet by publish target name and [None] (if item is not published)
- HYACINTH-197 #resolve


elohanlon:
Awesome changes.  Thanks, @cgalarza!

PR #55, opened 2017-11-21
Hyacinth 206

barmintor:
Backporting to rack 1.6.x, see also https://github.com/rack/rack/pull/1192

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@ image_server.yml
 uri_service.yml
 term_additional_fields.yml
 resque.yml
+repository_cache.yml
 ezid.yml
 
 *_user_accounts.yml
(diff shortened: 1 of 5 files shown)


elohanlon:
@barmintor would you mind updating this PR to only include the config/application.rb and config/boot.rb changes?  Gemfile has already been updated in master because of other changes I made recently, and repository_cache.yml isn't used by the app anymore and can be deleted in your local copy.  No need to ignore.

barmintor:
@elohanlon rack/rack#1192 has been reverted here:
https://github.com/rack/rack/commit/37c1160b2360074d20858792f23a7eb3afeabebd
with the underlying performance issue addressed here:
https://github.com/rack/rack/pull/1179

@botimer suggests this is released in rackm 2.0.6

PR #56, opened 2017-11-21
short circuit to false if checking basic editor permissions on a nil controlled vocabulary

barmintor:
This is to clarify the otherwise confusing server error when metadata editors look at the controlled_vocabularies index view.


elohanlon:
Travis seems to be back to normal now.  Re-ran the tests for the above commit and everything passed!  Will accept and merge.  Thanks for the PR!

PR #57, opened 2017-12-14
Ordered child import via row number dependency

elohanlon:


Gemfile
```diff
@@ -92,20 +92,20 @@ end
...
 gem 'devise', '>= 3.4.1'
 
 # CUL Fedora Dependencies and Content Models
 gem 'cul_hydra', '~> 1.4.7'
-#gem 'cul_hydra', :path => '../cul_hydra'
+#gem 'cul_hydra', path: '../cul_hydra'
 gem 'jettywrapper', '>= 1.5.1'
 # Temporarily use specific commit because new version of gem hasn't been released yet.  Latest is 1.1.3.
-gem 'rdf-rdfxml', :github => 'ruby-rdf/rdf-rdfxml', :ref => '78c13fe5dbcecaf1f56abe9535d00f16c670a764'
+gem 'rdf-rdfxml', git: 'https://github.com/ruby-rdf/rdf-rdfxml', ref: '78c13fe5dbcecaf1f56abe9535d00f16c670a764'
```
barmintor: should be in a separate PR, but: this pin is 3 years old, and has definitely been released :)

db/migrate/20171110194020_add_prerequisite_csv_row_number_to_digital_object_imports.rb
```diff
@@ -0,0 +1,11 @@
+class AddPrerequisiteCsvRowNumberToDigitalObjectImports < ActiveRecord::Migration
+  def change
+    change_table(:digital_object_imports) do |t|
+      t.text :prerequisite_csv_row_numbers, null: true
```
cgalarza: Does this have to be text because it's keeping and array under the hood?

app/jobs/process_digital_object_import_job.rb
```diff
@@ -1,82 +1,149 @@
...
     else
-      digital_object = new_object(digital_object_data, user, digital_object_import)
+      digital_object = new_object(digital_object_data['digital_object_type']['string_key'], user, digital_object_import)
     end
 
-    status = assign_data(digital_object, digital_object_data, digital_object_import)
+    result = assign_data(digital_object, digital_object_data, digital_object_import)
 
-    return if status == :requeued
+    if result == :parent_not_found
```
cgalarza: I wonder if we can retry more and wait less so that we aren't holding up the processing for 30sec if it ends up processing in 10sec. Have you done any testing around this? I don't think its a reason to hold up the pull request, but we might need to revisit later once this new processing is used in production.

app/jobs/process_digital_object_import_job.rb
```diff
@@ -1,82 +1,149 @@
...
     digital_object.set_digital_object_data(digital_object_data, true)
     :success
   rescue Hyacinth::Exceptions::ParentDigitalObjectNotFoundError => e
     digital_object_import.digital_object_errors << e.message
-    requeue_job(digital_object_import)
-  rescue Hyacinth::Exceptions::NotFoundError, Hyacinth::Exceptions::MalformedControlledTermFieldValue, UriService::Error => e
-    digital_object_import.digital_object_errors << e.message
+    :parent_not_found
+  rescue Exception => e
+    digital_object_import.digital_object_errors << e.message + '<span class="backtrace">' + "Backtrace:\n\t#{e.backtrace.join("\n\t")}" + '</span>'
```
cgalarza: I feel like the <span> should be moved into a template of some sort, but that would probably mean an update to how errors are stored. Again, not a deal breaker.

lib/hyacinth/utils/csv_import_export_utils.rb
```diff
@@ -161,11 +164,16 @@ def self.create_import_job_from_csv_data(csv_data_string, import_filename, user)
...
+
       Hyacinth::Utils::CsvImportExportUtils.csv_to_digital_object_data(csv_data_string) do |digital_object_data, csv_row_number|
+        prerequisite_csv_row_numbers = prerequisite_row_map[csv_row_number].present? ? prerequisite_row_map[csv_row_number].map { |row_number| row_number + 1 } : [] # CSV row numbers are 1-indexed
+
         digital_object_import = DigitalObjectImport.create!(
           import_job: import_job,
           digital_object_data: JSON.generate(digital_object_data),
-          csv_row_number: csv_row_number + 1
+          csv_row_number: csv_row_number + 1, # CSV row numbers are 1-indexed
+          prerequisite_csv_row_numbers: prerequisite_csv_row_numbers
```
cgalarza: Instead of saving prerequisite_csv_row_numbers, I wonder if we could store the id of the digital object import. For what I can tell it would mean that the jobs would have to be queued in a certain order in order to retrieve an id. This might add more complexity than its worth, but i figured i would throw it out there.

lib/hyacinth/utils/csv_import_export_utils.rb
```diff
@@ -175,6 +183,117 @@ def self.create_import_job_from_csv_data(csv_data_string, import_filename, user)
...
+
+    identifiers_to_row_numbers = identifiers_to_row_numbers_for_csv_data(csv_data_string)
+
+    # Return if no identifiers were found in the csv data, since that means that processing order doesn't matter
+    return {} if identifiers_to_row_numbers.blank?
+
+    prerequisite_rows_based_on_headers(csv_data_string, row_number_for_header_row, index_of_parent_headers, identifiers_to_row_numbers)
+  end
+
+  def self.prerequisite_rows_based_on_headers(csv_data_string, row_number_for_header_row, index_of_parent_headers, identifiers_to_row_numbers)
```
cgalarza: This method is starting to raise warning flags because of the all parameters required, but I totally see why it's needed. I almost wonder if a intermediate data structure would be helpful to make the code easier to read.

barmintor: I'm 👍  @cgalarza on this - the class is getting a little bulky in general. Maybe out of scope for this PR, but I'd definitely be in favor of whittling it down a little.

lib/hyacinth/utils/csv_import_export_utils.rb
```diff
@@ -175,6 +183,117 @@ def self.create_import_job_from_csv_data(csv_data_string, import_filename, user)
...
+        next
+      end
+      indexes_of_identifier_headers = indexes_of_identifier_headers_for_row(row)
+      break
+    end
+
+    return {} if indexes_of_identifier_headers.blank?
+
+    row_counter = -1
+    CSV.parse(csv_data_string) do |row|
```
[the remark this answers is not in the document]
barmintor: there is a lot of syntax overhead from supporting the ignorable "human readable" label row, I think - otherwise we could just tell `CSV` to use the headers as keys.

app/jobs/process_digital_object_import_job.rb
elohanlon: Good suggestion.  I'll do a loop in 10-second increments instead!

app/jobs/process_digital_object_import_job.rb
elohanlon: I'll make a template method to format this string.  Good idea!

lib/hyacinth/utils/csv_import_export_utils.rb
elohanlon: I'll have to think more about this, but it would be nice not to have another field.  We should revisit this idea later on.

lib/hyacinth/utils/csv_import_export_utils.rb
elohanlon: This many-param method comes from me having to refactor around rubocop's Cyclomatic Complexity rule.  Too many conditionals/branches in a single method.  I wish the variables were just in an outer scope and didn't need to be passed.  I'm also in favor of whittling down later on.

lib/hyacinth/utils/csv_import_export_utils.rb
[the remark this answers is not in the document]
elohanlon: I am 10,000% in favor of getting rid of that "human readable" label row.  It's not nearly as useful as we thought it would be, and is actually more trouble than it's worth for the people who originally asked for it.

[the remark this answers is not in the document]
elohanlon: @cgalarza I wish that worked, and I thought it would.  Unfortunately, it gives me this error:
```undefined method `with_index' for #<Array:0x007fd0f3a095d0>```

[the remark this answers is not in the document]
elohanlon: actually, this works: ```CSV.parse(csv_data_string).each_with_index do |row, csv_row_number|```

But does that mean I'm technically calling ```each.each_with_index```?  I'm using CSV.parse in order to avoid keeping the entire CSV in memory and only working with one line at a time.  If I call ```each.each_with_index```, does that mean that I'm aggregating the entire set of results from each and keeping them in memory in order pass them to each_with_index?  @barmintor, do you have any thoughts about that too?

elohanlon:
Closing this PR in favor of refactored version: https://github.com/cul/ldpd-hyacinth/pull/59

PR #62, opened 2017-12-28
Unpublish on delete, HYACINTH-186

cgalarza:
Unpublishing from all publish targets when an item is deleted.

I am not sure the tests I added are helpful, but it helped me understand the logic better. I have a feeling that I may be missing something else that needs to happen as part of this process. @elohanlon, let me know if you can think of anything. Should we add something to the delete dialog that says something along the lines of "Deleting this object unpublishes it from all publish targets"?


elohanlon:
@cgalarza I like the sound of that.  In `digital_objects_controller.js`, could you change this:

```
var confirmResponse = confirm('Are you sure you want to delete this Digital Object?');
```

to this?

```
var confirmResponse = confirm('Are you sure you want to delete this Digital Object and unpublish it from all publish targets?');
```

PR #64, opened 2018-01-04
Tests for XML Generation

cgalarza:
- Small fixes to xml generation code and a little clean up
- Test for the overall creation of a basic xml document
- Unit test for methods used within the larger recursive method

### DIFF
--- a/spec/models/concerns/digital_object/xml_datastream_rendering_spec.rb
+++ b/spec/models/concerns/digital_object/xml_datastream_rendering_spec.rb
@@ -112,7 +112,6 @@ describe DigitalObject::XmlDatastreamRendering do
       end
 
       it 'generates correct xml' do
-        # doc = Nokogiri::XML::Document.new
         digital_object.render_xml_translation_with_data(doc, doc, base_xml_translation, dynamic_field_data, xml_translation_map)
         expect(doc).to be_equivalent_to expected_mods
       end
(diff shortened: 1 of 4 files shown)


spec/models/concerns/digital_object/xml_datastream_rendering_spec.rb
```diff
@@ -0,0 +1,363 @@
...
+      expect(digital_object.value_for_field_name("name_term.last_name", name_df_data)).to eql ""
+    end
+  end
+
+  describe "#render_output_of_ternary" do
+    let(:name_df_data) { dynamic_field_data["name"][0] }
+
+    it 'returns "true" value if field present' do
+      arr = ["name_term.value", "True", "False"]
+      expect(digital_object.render_output_of_ternary(arr, name_df_data)).to eql "True"
```
elohanlon: Would you mind changing the "True" and "False" string values in the render_output_of_ternary tests to "Yes" and "No" (or something else), just so when looking at these tests, there's no chance that someone would incorrectly assume that render_output_of_ternary actually returns boolean values as strings?  I think it might make the functionality clearer at first glance.

spec/models/concerns/digital_object/xml_datastream_rendering_spec.rb
```diff
@@ -0,0 +1,363 @@
...
+            </mods:role>
+            <mods:role>
+              <mods:roleTerm type="text" valueURI="http://id.loc.gov/roles/789">Editor</mods:roleTerm>
+            </mods:role>
+         </mods:name>
+        </mods:mods>'
+      end
+
+      it 'generates correct xml' do
+        # doc = Nokogiri::XML::Document.new
```
elohanlon: accidental leftover comment?

spec/models/concerns/digital_object/xml_datastream_rendering_spec.rb
cgalarza: Will do

PR #65, opened 2018-01-08
XML Generation Refactor and HYACINTH-210

cgalarza:
- Refactored XML rendering code into two classes.
- Broke-up rendering code into smaller method
- Fixed/added test after refactor
- Added not_equals as a conditional that can be used
- Checking render_if statement before rendering attributes
- Resolves HYACINTH-210


lib/hyacinth/xml_generator.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    def initialize(digital_object_data, base_xml_translation, dynamic_fields_groups_map, internal_fields = {})
+      @dynamic_fields_groups_map = dynamic_fields_groups_map # Map of dynamic field group keys to their xml_translation
+      @base_xml_translation = base_xml_translation
+      @digital_object_data = digital_object_data
+      @internal_fields = internal_fields
+    end
+
+    def generate
+      @document = Nokogiri::XML::Document.new
```
elohanlon: @cgalarza You use `@document` here, but then call the `document` method in three cases that immediately follow.  I know that this is completely valid and works as intended, but part of me wanted to use the instance variable in all four cases.  I saw this as an opportunity to ask: In cases like this, is it clearer/better to reference an instance variable directly or the same-name method created by an attr_reader?  Also interested in what @barmintor thinks about this.

cgalarza: I would be curious as to what @barmintor thinks. In cases like this I usually only use the @ reference to the instance variable to set the value of the variable, otherwise I use method definition. If we were to override the getter method, then we would be correctly using our new method. That being said, I don't feel particularly strongly about this.

PR #66, opened 2018-01-15
Safe shutdown of resque workers (to avoid dirty exit)

elohanlon:


lib/tasks/resque.rake
```diff
@@ -50,10 +52,31 @@ namespace :resque do
...
+      puts "\nTelling workers to finish current jobs, but not process any new jobs..."
+      syscmd = "kill -s USR2 #{pids.join(' ')}"
+      puts "$ #{syscmd}"
+      `#{syscmd}`
+      puts "\n"
+      puts "Waiting for workers to finish current jobs..."
+      start_time = Time.now
+      while (Time.now - start_time) < MAX_WAIT_TIME_TO_KILL_WORKERS do
+        sleep 1
+        num_workers_working = Resque.workers.select { |w| w.working? }.length
```
cgalarza: Just a style thing, I like to use `Resque.workers.select(&:working?).length`.

lib/tasks/resque.rake
```diff
@@ -50,10 +52,31 @@ namespace :resque do
...
+        puts "#{num_workers_working} workers still working..."
+        break if num_workers_working == 0
```
cgalarza: Another style thing, I like using `num_workers_working.zero?`

lib/tasks/resque.rake
```diff
@@ -50,10 +52,31 @@ namespace :resque do
...
       puts "No workers to kill"
     else
+      # First tell workers to stop accepting new work by sending USR2 signal
+      puts "\nTelling workers to finish current jobs, but not process any new jobs..."
+      syscmd = "kill -s USR2 #{pids.join(' ')}"
+      puts "$ #{syscmd}"
+      `#{syscmd}`
+      puts "\n"
+      puts "Waiting for workers to finish current jobs..."
+      start_time = Time.now
```
cgalarza: It probably won't matter too much in this case because you are consistently using `Time.now` but for rails app `Time.current` is better because it takes into consideration the timezone set by the rails application.

elohanlon:
Great suggestions, thank you!  Pushing a new commit with changes in a sec and then will merge.

PR #71, opened 2018-02-24
Csv import priority

elohanlon:


app/jobs/process_digital_object_import_job.rb
```diff
@@ -195,7 +195,7 @@ def self.handle_remaining_prerequisite_case(digital_object_import, queue_long_jo
       # If prerequisite are still pending, then re-queue this import
       digital_object_import.digital_object_errors = [] # clear earlier errors if we're re-queueing
       digital_object_import.save!
-      Hyacinth::Queue.process_digital_object_import(digital_object_import.id)
+      Hyacinth::Queue.process_digital_object_import(digital_object_import.id, digital_object_import.import_job.priority.to_sym)
```
cgalarza: Why not just pass the digital_object_import object since you're pulling out to bits of information from the same object?

elohanlon: @cgalarza That is a really good question.  I'll make that change now, thanks!

elohanlon:
Tests actually did pass.  It's that weird jetty connection error again.  Merging PR!

PR #80, opened 2019-02-20
Cloning with SSH

cgalarza:
I had to reset the origin when I cloned it with https. It kept asking for my username and password. This might be how my git configuration it set up though. Do you remember having this problem?


elohanlon:
Oh, good catch.  Definitely want the ssh version rather than https, otherwise it'll do exactly what you're describing.  I didn't have this problem because I cloned the repo via ssh before writing the README.

PR #86, opened 2019-03-14
More updates to group and users

cgalarza:


cgalarza:
@elohanlon I missed these two commit in my previous pull request

PR #88, opened 2019-03-18
Controlled Vocabulary API

cgalarza:
- API endpoint for controlled vocabulary within Hyacinth
- Permissions are required to use this endpoint
- URI Service credentials have to be added to the hyacinth.yml file

Note: This pull request is missing tests for the controlled vocabulary endpoint


lib/uri_service.rb
```diff
@@ -0,0 +1,15 @@
+module URIService
+  thread_mattr_accessor :active_connection
+
+  def self.connection
+    self.active_connection = new_connection unless active_connection
+    active_connection
+  end
+
+  def self.new_connection
+    config = Rails.application.config_for(:hyacinth)['uri_service']
```
elohanlon: Rather than re-loading hyacinth.yml, you can re-use the `HYACINTH` constant that's made available in the `00_hyacinth.rb` initializer.

PR #89, opened 2019-03-18
Project Core API

cgalarza:
- Basic API to CRUD projects
- This endpoint does not contain authorization yet

### DIFF
--- a/spec/requests/api/v1/projects_spec.rb
+++ b/spec/requests/api/v1/projects_spec.rb
@@ -188,7 +188,7 @@ RSpec.describe 'Projects Requests', type: :request do
       end
     end
 
-    context 'when deleting a project that does not exists' do
+    context 'when deleting a project that does not exist' do
       before { delete '/api/v1/projects/not-valid' }
 
       it 'returns errors' do
--- a/spec/requests/api/v1/projects_spec.rb
+++ b/spec/requests/api/v1/projects_spec.rb
@@ -94,7 +94,7 @@ RSpec.describe 'Projects Requests', type: :request do
       end
     end
 
-    context 'when creating requires string_key' do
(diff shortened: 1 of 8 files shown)


spec/requests/api/v1/projects_spec.rb
```diff
@@ -0,0 +1,205 @@
...
+          "project": {
+            "display_label": "Best Project",
+            "project_url": "https://best_project.com",
+            "string_key": "best_project"
+          }
+        }))
+      end
+    end
+
+    context 'when creating requires string_key' do
```
elohanlon:
```suggestion
    context 'when create request is missing string_key' do
```

spec/requests/api/v1/projects_spec.rb
```diff
@@ -0,0 +1,205 @@
...
+      it 'returns error' do
+        expect(response.body).to be_json_eql(%({
+          "errors": [
+            { "title": "String key can't be blank" }
+          ]
+        }))
+      end
+    end
+
+    context 'when creating requires display_label' do
```
elohanlon:
```suggestion
    context 'when create request is missing display_label' do
```

spec/requests/api/v1/projects_spec.rb
```diff
@@ -0,0 +1,205 @@
...
+      it 'returns 204' do
+        expect(response.status).to be 204
+      end
+
+      it 'deletes record from database' do
+        expect(Project.find_by(string_key: string_key)).to be nil
+      end
+    end
+
+    context 'when deleting a project that does not exists' do
```
elohanlon:
```suggestion
    context 'when deleting a project that does not exist' do
```

PR #91, opened 2019-03-26
Dynamic Fields

cgalarza:
- Adding database migrations and models for dynamic_field, dynamic_field_group and dynamic_field_category
- Adding tests for all three models


app/models/concerns/dynamic_field_structure/string_key.rb
```diff
@@ -0,0 +1,20 @@
+module DynamicFieldStructure
+  module StringKey
+    extend ActiveSupport::Concern
+
+    RESERVED_STRING_KEYS = ['uri', 'value']
```
elohanlon: Is 'value' reserved for a reason that's specific to your code design, or is this a carry-over from Hyacinth 2?  I remember why we'd want 'uri' to be reserved, since that makes it easier to find controlled term fields in a deeply nested hash.  I think that in Hyacinth 2, 'value' was also reserved for a similar reason, when URI Service used 'value' as its pref label field name.  If we still want to preserve that ease of lookup in a nested hash (which may be a good idea), I think we should change 'value' above to 'pref_label'.

app/models/concerns/dynamic_field_structure/string_key.rb
```diff
@@ -0,0 +1,20 @@
...
+      validates :string_key, presence: true,  string_key: true, exclusion: { in: RESERVED_STRING_KEYS }
+      validate :unique_string_key
+    end
+
+    private
+
+      # Validate that the string_key is unique between all siblings.
+      def unique_string_key
+        sibling_string_keys = siblings.map(&:string_key)
+        errors.add(:string_key, 'has already been taken') if sibling_string_keys.include?(string_key)
```
elohanlon:
```suggestion
        errors.add(:string_key, 'has already been taken by a sibling field') if sibling_string_keys.include?(string_key)
```

app/models/dynamic_field.rb
```diff
@@ -0,0 +1,71 @@
...
+    TEXTAREA = 'textarea'
+    INTEGER = 'integer'
+    BOOLEAN = 'boolean'
+    SELECT = 'select'
+    DATE = 'date'
+    CONTROLLED_TERM = 'controlled_term'
+  end
+
+  TYPES_TO_LABELS = {
+    DynamicField::Type::STRING => 'String',
```
elohanlon: It might be possible to just use `STRING => 'String'` here rather than the full `DynamicField::Type::STRING`, but I'm not 100% sure.  By the way, I realize that this may be carry-over code from Hyacinth 2, so this and other comments to come may be me proposing a revision to something that I originally wrote.

db/migrate/20190325175848_create_dynamic_field.rb
```diff
@@ -0,0 +1,31 @@
...
+      t.string  :string_key,    null: false
+      t.string  :display_label, null: false
+      t.string  :field_type,    null: false, default: 'string'
+      t.integer :sort_order,    null: false
+
+      t.boolean :is_facetable,  null: false, default: false
+      t.string  :filter_label
+      t.string  :controlled_vocabulary
+      t.text    :select_options
+      t.text    :additional_data_json
```
elohanlon: Do we want to get rid of additional_data_json since we have select_options?

app/validators/valid_json_validator.rb
```diff
@@ -0,0 +1,11 @@
+class ValidJsonValidator < ActiveModel::EachValidator
+  ALPHANUMERIC_UNDERSCORE_KEY_REGEX = /\A[a-z]{1}[a-z0-9_]{0,240}\z/
```
elohanlon: Not needed?

db/migrate/20190325175848_create_dynamic_field.rb
```diff
@@ -0,0 +1,31 @@
+class CreateDynamicField < ActiveRecord::Migration[5.2]
```
elohanlon: For consistency, this should be plural (`CreateDynamicFields`).

spec/factories/dynamic_field.rb
```diff
@@ -0,0 +1,21 @@
+FactoryBot.define do
```
elohanlon: For consistency, this factory file should be `dynamic_fields.rb`.

spec/models/dynamic_field_group_spec.rb
```diff
@@ -0,0 +1,339 @@
...
+      let(:dynamic_field_group) { FactoryBot.build(:dynamic_field_group, xml_translation: 'randomstring') }
+
+      it 'does not save' do
+        expect(dynamic_field_group.save).to be false
+      end
+
+      it 'returns correct error' do
+        dynamic_field_group.save
+        expect(dynamic_field_group.errors.full_messages).to include 'Xml translation does not validate as JSON'
+      end
```
elohanlon: We should probably also add a validation rule someday that validates the the xml translation rules against a json schema for valid xml translation rule properties, but that can wait until later.  Maybe worth putting a TODO or "skip" test in here for that?

spec/models/dynamic_field_group_spec.rb
```diff
@@ -0,0 +1,339 @@
...
+        let(:parent) { FactoryBot.create(:dynamic_field_category) }
+        let(:dynamic_field_group) { FactoryBot.build(:dynamic_field_group, parent: parent) }
+
+        before do
+          FactoryBot.create(:dynamic_field_group, string_key: 'name', parent: parent)
+          parent.reload
+        end
+
+        it 'does not save' do
+          expect(dynamic_field_group.save).to be false
```
elohanlon: The only thing I don't like about this test is that it assumes knowledge of the factory.  It assumes that you know that the factory for :dynamic_field_group defaults to having a string_key of 'name'.  Maybe that's an okay practice though.  Not sure, since I only recently started using FactoryBot.

spec/models/dynamic_field_group_spec.rb
```diff
@@ -0,0 +1,339 @@
...
+    context 'when setting DynamicFieldGroup as parent' do
+      let(:parent) { FactoryBot.create(:dynamic_field_group) }
+
+      it 'adds parent' do
+        expect(dynamic_field_group.parent).to eq parent
+      end
+
+      it 'adds dynamic field group to parent dynamic field group' do
+        expect(parent.dynamic_field_groups).to match_array [dynamic_field_group]
+      end
```
elohanlon: We should add another test here that ensures that a dynamic_field_group cannot be added as its own parent (because that would lead to infinite loops elsewhere in the code as we traverse the field hierarchy).

spec/models/dynamic_field_group_spec.rb
```diff
@@ -0,0 +1,339 @@
...
+    end
+
+    context 'when setting parent to invalid parent type' do
+      let(:parent) { FactoryBot.create(:user, email: 'random_user@example.com') }
+
+      it 'does save' do
```
elohanlon:
```suggestion
      it 'does not save' do
```

spec/models/dynamic_field_group_spec.rb
```diff
@@ -0,0 +1,339 @@
...
+  end
+
+  describe '#dynamic_fields' do
+    let(:dynamic_field_group) { FactoryBot.create(:dynamic_field_group) }
+    let(:dynamic_field_1) { FactoryBot.build(:dynamic_field, dynamic_field_group: nil) }
+    let(:dynamic_field_2) { FactoryBot.build(:dynamic_field, dynamic_field_group: nil, string_key: 'usage_primary') }
+
+    it 'can add a dynamic field' do
+      dynamic_field_group.dynamic_fields << dynamic_field_1
+      dynamic_field_1.save
```
elohanlon: As it is, `dynamic_field_1.save` doesn't add anything to this test.  You could either remove this line or call `.reload` on the next line.

spec/models/dynamic_field_group_spec.rb
```diff
@@ -0,0 +1,339 @@
...
+      expect(dynamic_field_group.dynamic_fields).to match_array [dynamic_field_1]
+    end
+
+    it 'can add multiple dynamic fields' do
+      dynamic_field_group.dynamic_fields << dynamic_field_1
+      dynamic_field_group.dynamic_fields << dynamic_field_2
+      dynamic_field_1.save
```
elohanlon: Same comment here as above, about `dynamic_field_1.save` not adding anything to this test.  I think the test would still function the same way without this `.save` call and the one below it.

spec/models/dynamic_field_spec.rb
```diff
@@ -0,0 +1,186 @@
+require 'rails_helper'
+
+RSpec.describe DynamicField, type: :model do
+  describe '#new' do
+    context 'when parameters are correct' do
+      subject { FactoryBot.create(:dynamic_field) }
```
elohanlon: Just as a heads up, rubocop will be complaining about not having a "named subject" here once my bixby changes are merged in.  I'm fine with fixing this later, but it might be faster to fix now while this set of tests is fresh in your mind.

https://www.rubydoc.info/gems/rubocop-rspec/1.6.0/RuboCop/Cop/RSpec/NamedSubject

spec/models/dynamic_field_spec.rb
```diff
@@ -0,0 +1,186 @@
...
+    context 'when missing string_key' do
+      let(:dynamic_field) { FactoryBot.build(:dynamic_field, string_key: nil) }
+
+      it 'does not save' do
+        expect(dynamic_field.save).to be false
+      end
+
+      it 'returns correct error' do
+        dynamic_field.save
+        expect(dynamic_field.errors.full_messages).to include 'String key can\'t be blank'
```
elohanlon: Optional change: For readability, I tend to prefer double quotes when a single quote is in a string, even when we're not interpolating anything.  Not sure if rubocop would complain about that though.

spec/models/dynamic_field_spec.rb
```diff
@@ -0,0 +1,186 @@
...
+    context 'when field_type invalid' do
+      let(:dynamic_field) { FactoryBot.build(:dynamic_field, field_type: 'not-valid') }
+
+      it 'does not save' do
+        expect(dynamic_field.save).to be false
+      end
+
+      it 'returns correct error' do
+        dynamic_field.save
+        expect(dynamic_field.errors.full_messages).to include 'Field type is not included in the list'
```
elohanlon:
```suggestion
        expect(dynamic_field.errors.full_messages).to include 'Field type is not included in the list of allowed types'
```

spec/models/dynamic_field_group_spec.rb
```diff
@@ -0,0 +1,339 @@
...
+    context 'when setting parent to invalid parent type' do
+      let(:parent) { FactoryBot.create(:user, email: 'random_user@example.com') }
+
+      it 'does save' do
+        expect(dynamic_field_group.save).to be false
+      end
+
+      it 'returns correct error' do
+        dynamic_field_group.save
+        expect(dynamic_field_group.errors.full_messages).to include 'Parent type is not included in the list'
```
elohanlon:
```suggestion
        expect(dynamic_field_group.errors.full_messages).to include 'Invalid parent type'
```

spec/models/dynamic_field_spec.rb
```diff
@@ -0,0 +1,186 @@
...
+    context 'when missing dynamic_field_group' do
+      let(:dynamic_field) { FactoryBot.build(:dynamic_field, dynamic_field_group: nil) }
+
+      it 'does not save' do
+        expect(dynamic_field.save).to be false
+      end
+
+      it 'returns correct error' do
+        dynamic_field.save
+        expect(dynamic_field.errors.full_messages).to include 'Dynamic field group must exist'
```
elohanlon:
```suggestion
        expect(dynamic_field.errors.full_messages).to include 'Dynamic field group is required'
```

spec/models/dynamic_field_spec.rb
```diff
@@ -0,0 +1,186 @@
...
+    context 'when missing created_by' do
+      let(:dynamic_field) { FactoryBot.build(:dynamic_field, created_by: nil) }
+
+      it 'does not save' do
+        expect(dynamic_field.save).to be false
+      end
+
+      it 'returns correct error' do
+        dynamic_field.save
+        expect(dynamic_field.errors.full_messages).to include 'Created by must exist'
```
elohanlon:
```suggestion
        expect(dynamic_field.errors.full_messages).to include 'Created by value is required'
```

spec/models/dynamic_field_spec.rb
```diff
@@ -0,0 +1,186 @@
...
+    context 'when missing updated_by' do
+      let(:dynamic_field) { FactoryBot.build(:dynamic_field, updated_by: nil) }
+
+      it 'does not save' do
+        expect(dynamic_field.save).to be false
+      end
+
+      it 'returns correct error' do
+        dynamic_field.save
+        expect(dynamic_field.errors.full_messages).to include 'Updated by must exist'
```
elohanlon:
```suggestion
        expect(dynamic_field.errors.full_messages).to include 'Updated by value is required'
```

spec/models/dynamic_field_spec.rb
```diff
@@ -0,0 +1,186 @@
...
+        expect(dynamic_field.errors.full_messages).to include 'Select options can\'t be blank'
+      end
+    end
+
+    context 'when creating controlled term dynamic field' do
+      let(:dynamic_field) { FactoryBot.build(:dynamic_field, controlled_vocabulary: nil) }
+
+      it 'requires controlled_vocabulary' do
+        expect(dynamic_field.save).to be false
+        expect(dynamic_field.errors.full_messages).to include 'Controlled vocabulary can\'t be blank'
```
elohanlon:
```suggestion
        expect(dynamic_field.errors.full_messages).to include 'Controlled vocabulary is required'
```

spec/models/dynamic_field_spec.rb
```diff
@@ -0,0 +1,186 @@
...
+          parent.reload
+        end
+
+        it 'does not save' do
+          expect(dynamic_field.save).to be false
+        end
+
+        it 'returns correct error' do
+          dynamic_field.save
+          expect(dynamic_field.errors.full_messages).to include 'String key has already been taken'
```
elohanlon:
```suggestion
          expect(dynamic_field.errors.full_messages).to include 'String key is already in use for an adjacent field or field group'
```

spec/models/dynamic_field_spec.rb
```diff
@@ -0,0 +1,186 @@
...
```
elohanlon:
```suggestion
          expect(dynamic_field.errors.full_messages).to include 'String key is already in use by a sibling field or field group'
```

spec/models/dynamic_field_group_spec.rb
```diff
@@ -0,0 +1,339 @@
...
+          parent.reload
+        end
+
+        it 'does not save' do
+          expect(dynamic_field_group.save).to be false
+        end
+
+        it 'returns correct error' do
+          dynamic_field_group.save
+          expect(dynamic_field_group.errors.full_messages).to include 'String key has already been taken'
```
elohanlon:
```suggestion
          expect(dynamic_field_group.errors.full_messages).to include 'String key is already in use by a sibling field or field group'
```

spec/models/dynamic_field_group_spec.rb
```diff
@@ -0,0 +1,339 @@
...
```
elohanlon:
```suggestion
          expect(dynamic_field_group.errors.full_messages).to include 'String key is already in use by a sibling field or field group'
```

app/models/concerns/dynamic_field_structure/string_key.rb
cgalarza: The reserved string keys validation is a carryover from HY2. I have no idea why we placed this restriction. I don't really understand the use case you describe above. If swapping out 'value' for 'pref_label' is the correct thing to do, I am happy to change it.

app/models/dynamic_field.rb
cgalarza: This is carry over code. We can probably remove the `DynamicField::`. Will do.

db/migrate/20190325175848_create_dynamic_field.rb
cgalarza: I wanted it leave it so that it would make migrating the data easier, and we can remove it later.

app/validators/valid_json_validator.rb
cgalarza: yup. will remove.

spec/models/dynamic_field_group_spec.rb
cgalarza: That's a cool idea. I'll add a TODO

spec/models/dynamic_field_group_spec.rb
cgalarza: I am not sure what the practice should be in cases like this. I just don't want to keep writing boiler plate models over and over again, so I depend on the string_key of the default factory. Might be worth aliasing the default factory to a trait called :name.

spec/models/dynamic_field_group_spec.rb
cgalarza: Good point.

spec/models/dynamic_field_group_spec.rb
cgalarza: I don't think they would because I just build the factories, not create them. I just wanted to test the addition of new fields. I am adding, saving, reloading and then test to make sure they are present.

spec/models/dynamic_field_spec.rb
cgalarza: Can you push the bixby changes to 3.x? There's going to be a lot more errors than just this one.

spec/models/dynamic_field_spec.rb
cgalarza: Since its just one quotation mark, i usually don't switch it to double quotation mark.

spec/models/dynamic_field_spec.rb
cgalarza: I think some of these new error messages are helpful, and I think I will implement some of them. I would like to keep the defaults in some cases tho. It think its just more code to maintain that is only beneficial in uncommon situations.

app/models/dynamic_field.rb
cgalarza: I just realized that we can probably remove `TYPES_TO_LABELS` all together, because that's used by the front-end code and we are no longer going to be able to use it. Also, it's easy enough to convert the strings to their label counterparts with code.

spec/models/dynamic_field_group_spec.rb
cgalarza: Updated error message!

spec/models/dynamic_field_spec.rb
cgalarza: This error message comes from the belongs_to association. Belongs_to associations are required in rails 5, unless specified otherwise. We could add in an additional validation to check for presence. It would be redundant, but would allow us to change the error messaging.

spec/models/dynamic_field_group_spec.rb
cgalarza: I'll change save to save!, per our discussion

app/models/concerns/dynamic_field_structure/string_key.rb
elohanlon: After we spoke about this, I'm now thinking that RESERVED_STRING_KEYS isn't necessary anymore.  Just to be safe though, let's keep it for a little while with a TODO note to eventually delete if not needed, and let's change 'value' to 'pref_label' so that it serves the same protective purposes as in Hyacinth 2.

elohanlon:
```suggestion
    RESERVED_STRING_KEYS = ['uri', 'pref_label']
```

db/migrate/20190325175848_create_dynamic_field.rb
elohanlon: 👍

spec/models/dynamic_field_spec.rb
elohanlon: Will do.

spec/models/dynamic_field_group_spec.rb
elohanlon: The reason I suggested this change is that it seems confusing for a user of the API or the GUI to see a message that refers to "the list".  What list should they assume the error is referring to?

spec/models/dynamic_field_spec.rb
elohanlon: I don't like the error redundancy that the user would see in the additional validation case, but I just feel like the phrasing "updated by must exist" is a little odd for a user seeing this in an error message.  My feelings about this aren't too strong though, so I can live with it as is.

spec/models/dynamic_field_spec.rb
elohanlon: This change doesn't bother me all that much, so I'm fine with keeping as is, and I get what you mean about the extra code.

app/models/concerns/dynamic_field_structure/string_key.rb
cgalarza: Done!

spec/models/dynamic_field_spec.rb
cgalarza: Thanks let me know!

spec/models/dynamic_field_group_spec.rb
cgalarza: I get that. Like maybe "not included in valid list" or something like that. But again, i feel like we should just change this across the app. I don't know how to do that, yet, or if its even possible

spec/models/dynamic_field_spec.rb
cgalarza: I get that, I just didn't see a way to easy change that. So I say we leave it as is in order to save time.

spec/models/dynamic_field_spec.rb
cgalarza: Cool, thanks!

app/models/concerns/dynamic_field_structure/string_key.rb
cgalarza: Will do.

spec/models/dynamic_field_group_spec.rb
cgalarza: We might be able to use localization for this, which might be awesome: https://github.com/rails/rails/blob/e69ff43060c1194d2a3bd9b8d9e23f3ae26b84b5/activemodel/lib/active_model/errors.rb#L401

cgalarza: Gotta love rails: https://guides.rubyonrails.org/i18n.html#error-message-scopes

cgalarza: @elohanlon Did you want to go through and change the error messages for list validation and belongs to validations across the app? And if so, what should we change the wording to?

cgalarza: As discussed we change the belongs_to validation to 'is_required' and the inclusion validation to 'is not among the list of allowed values'

PR #93, opened 2019-04-01
CRUD for dynamic_fields, dynamic_field_groups, dynamic_field_categories

cgalarza:


spec/requests/api/v1/dynamic_field_categories_spec.rb
```diff
@@ -0,0 +1,231 @@
...
+        it 'returns 204' do
+          expect(response.status).to be 204
+        end
+
+        it 'deletes record from database' do
+          expect(DynamicFieldCategory.find_by(id: id)).to be nil
+        end
+      end
+
+      context 'when deleting a dynamic_field_category that does not exists' do
```
elohanlon:
```suggestion
      context 'when deleting a dynamic_field_category that does not exist' do
```

spec/requests/api/v1/dynamic_field_groups_spec.rb
```diff
@@ -0,0 +1,224 @@
...
+        it 'returns 204' do
+          expect(response.status).to be 204
+        end
+
+        it 'deletes record from database' do
+          expect(DynamicFieldGroup.find_by(id: id)).to be nil
+        end
+      end
+
+      context 'when deleting a dynamic_field_group that does not exists' do
```
elohanlon:
```suggestion
      context 'when deleting a dynamic_field_group that does not exist' do
```

PR #96, opened 2019-04-03
enabled dynamic fields & field sets

cgalarza:
- Adding db migrations and models for enabled dynamic fields and field sets
- Adding tests for both

### DIFF
--- a/app/controllers/api/v1/projects/enabled_dynamic_fields_controller.rb
+++ b/app/controllers/api/v1/projects/enabled_dynamic_fields_controller.rb
@@ -33,7 +33,7 @@ module Api
           def update_params
             params.permit(
               enabled_dynamic_fields: [
-                :id, :dynamic_field_id, :required, :locked, :hidden, :owner_only, :default_value
+                :id, :dynamic_field_id, :required, :locked, :hidden, :owner_only, :default_value, field_set_ids: []
               ]
             )
           end
--- a/app/controllers/api/v1/projects/enabled_dynamic_fields_controller.rb
+++ b/app/controllers/api/v1/projects/enabled_dynamic_fields_controller.rb
@@ -1,0 +1,42 @@
+module Api
+  module V1
+    module Projects
+      class EnabledDynamicFieldsController < ApplicationApiController
+        load_resource :project, find_by: :string_key, id_param: :project_string_key
+
+        # GET /projects/:string_key/enabled_dynamic_fields/:digital_object_type
+        def show
+          render json: { enabled_dynamic_fields: enabled_dynamic_fields }, status: :ok
+        end
+
+        # PATCH /projects/:string_key/enabled_dynamic_fields/:digital_object_type
+        def update
+          update_params[:enabled_dynamic_fields].each { |h| h[:digital_object_type] = params[:digital_object_type] }
+          updated_enabled_dynamic_fields = {
+            enabled_dynamic_fields_attributes: update_params[:enabled_dynamic_fields]
+          }
+
+          if @project.update(updated_enabled_dynamic_fields)
+            render json: { enabled_dynamic_fields: enabled_dynamic_fields(reload: true) }, status: :ok
+          else
+            render json: errors(@project.errors.full_messages), status: :unprocessable_entity
+          end
+        end
+
+        private
+          def enabled_dynamic_fields(reload: false)
+            @project.reload if reload
+            @project.enabled_dynamic_fields.where(digital_object_type: params[:digital_object_type])
+          end
+
+          def update_params
+            params.permit(
+              enabled_dynamic_fields: [
+                :id, :dynamic_field_id, :required, :locked, :hidden, :owner_only, :default_value
+              ]
+            )
+          end
+      end
+    end
+  end
+end
--- a/app/controllers/api/v1/projects/enabled_dynamic_fields_controller.rb
+++ b/app/controllers/api/v1/projects/enabled_dynamic_fields_controller.rb
@@ -24,6 +24,7 @@ module Api
         end
 
         private
+
           def enabled_dynamic_fields(reload: false)
             @project.reload if reload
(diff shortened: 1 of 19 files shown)


app/models/project.rb
```diff
@@ -1,5 +1,9 @@
 class Project < ApplicationRecord
   has_many :publish_targets
```
cgalarza: @elohanlon Should this also be `dependent: :destroy`?

elohanlon: Oh yes, you're right

config/routes.rb
```diff
@@ -34,6 +34,10 @@
 
       resources :projects, param: :string_key, except: [:new, :edit] do
         resources :publish_targets, param: :string_key, except: [:new, :edit], module: 'projects'
+        resources :field_sets,                          except: [:new, :edit], module: 'projects'
+        resources :enabled_dynamic_fields,
+                  only: [:show, :update], module: 'projects',
```
elohanlon: Is `module: 'projects'` necessary here (and in the other two cases), since this `resources` definition is within `resources :projects`?

cgalarza: I couldn't get the route to use the nested controller without using the `scope` parameter. There might be a more succinct way to configure this though.

cgalarza: Maybe this: https://github.com/cul/ldpd-hyacinth/blob/master/config/routes.rb#L130-L133

app/controllers/api/v1/projects/field_sets_controller.rb
```diff
@@ -0,0 +1,55 @@
+module Api
+  module V1
+    module Projects
+      class FieldSetsController < ApplicationApiController
+        before_action :ensure_json_request
+
+        load_resource :project, find_by: :string_key, id_param: :project_string_key
+        load_resource :field_set, through: :project
+
+        # GET /projects/:string_key/fieldsets
```
elohanlon:
```suggestion
        # GET /projects/:string_key/field_sets
```

app/controllers/api/v1/projects/field_sets_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+        def index
+          render json: { field_sets: @project.field_sets }, status: :ok
+        end
+
+        # GET /projects/:string_key/fieldsets/:id
```
elohanlon:
```suggestion
        # GET /projects/:string_key/field_sets/:id
```

app/controllers/api/v1/projects/field_sets_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+        def show
+          render json: { field_set: @field_set }, status: :ok
+        end
+
+        # POST /projects/:string_key/fieldsets
```
elohanlon:
```suggestion
        # POST /projects/:string_key/field_sets
```

app/controllers/api/v1/projects/field_sets_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+        def create
+          if @field_set.save
+            render json: { field_set: @field_set }, status: :created
+          else
+            render json: errors(@field_set.errors.full_messages), status: :unprocessable_entity
+          end
+        end
+
+        # PATCH /projects/:string_key/fieldsets/:id
```
elohanlon:
```suggestion
        # PATCH /projects/:string_key/field_sets/:id
```

app/controllers/api/v1/projects/field_sets_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+        def update
+          if @field_set.update(field_set_params)
+            render json: { fieldset: @field_set }, status: :ok
+          else
+            render json: errors(@field_set.errors.full_messages), status: :unprocessable_entity
+          end
+        end
+
+        # DELETE /projects/:string_key/fieldsets/:id
```
elohanlon:
```suggestion
        # DELETE /projects/:string_key/field_sets/:id
```

config/routes.rb
elohanlon: We can always revisit it later.  I'm good with keeping as is.

app/controllers/api/v1/projects/enabled_dynamic_fields_controller.rb
```diff
@@ -0,0 +1,43 @@
+module Api
+  module V1
+    module Projects
```
elohanlon: At some point we'll need to add `include_examples 'requires user to have correct permissions' do` and check for project admin permissions, but that can come later.

spec/requests/api/v1/projects/enabled_dynamic_fields_spec.rb
```diff
@@ -0,0 +1,151 @@
+require 'rails_helper'
+
+RSpec.describe 'Enabled Dynamic Fields Requests', type: :request do
```
elohanlon: At some point we'll need to add include_examples 'requires user to have correct permissions' do and check for project admin permissions, but that can come later.  Probably want to add a TODO though.

spec/requests/api/v1/projects/field_sets_spec.rb
```diff
@@ -0,0 +1,177 @@
+require 'rails_helper'
+
+RSpec.describe 'Field Sets Requests', type: :request do
```
elohanlon: Another case where we'll need to add include_examples 'requires user to have correct permissions' do and check for project admin permissions, but that can come later too. Probably want to add a TODO here as well.

cgalarza: Project permissions have to no been implemented. I am working on that right now and I'll add tests to all these routes at that point.

spec/requests/api/v1/projects/enabled_dynamic_fields_spec.rb
cgalarza: see other comment

PR #98, opened 2019-04-08
field_export_profiles & export_rules

cgalarza:
- Adding models for export_rules and field_export_profiles, updated relationships between new models and current models
- Added CRUD functionality for field_export_profiles
- Allowing export_rules to be added to dynamic_field_groups

### DIFF
--- a/app/controllers/api/v1/dynamic_field_groups_controller.rb
+++ b/app/controllers/api/v1/dynamic_field_groups_controller.rb
@@ -15,6 +15,8 @@ module Api
         @dynamic_field_group.created_by = current_user
         @dynamic_field_group.updated_by = current_user
 
+        @dynamic_field_group.export_rules_attributes = export_rules_attributes
+
         if @dynamic_field_group.save
           render json: { dynamic_field_group: @dynamic_field_group }, status: :created
         else
@@ -25,6 +27,8 @@ module Api
       # PATCH /dynamic_field_groups/:id
       def update
         @dynamic_field_group.updated_by = current_user
+
+        @dynamic_field_group.export_rules_attributes = export_rules_attributes
 
         if @dynamic_field_group.update(update_params)
           render json: { dynamic_field_group: @dynamic_field_group }, status: :ok
@@ -46,14 +50,22 @@ module Api
 
         def create_params
           params.require(:dynamic_field_group).permit(
-            :string_key, :display_label, :sort_order, :is_repeatable, :xml_translation, :parent_type, :parent_id
+            :string_key, :display_label, :sort_order, :is_repeatable, :parent_type, :parent_id
           )
         end
 
         def update_params
           params.require(:dynamic_field_group).permit(
-            :display_label, :is_repeatable, :xml_translation, :sort_order, :parent_type, :parent_id
+            :display_label, :is_repeatable, :sort_order, :parent_type, :parent_id
           )
+        end
+
+        def export_rules_attributes
+          export_rules_params ? export_rules_params.fetch(:export_rules, []) : []
+        end
+
+        def export_rules_params
+          params.require(:dynamic_field_group).permit(export_rules: [:id, :translation_logic, :field_export_profile_id])
         end
     end
   end
--- a/app/controllers/api/v1/dynamic_field_groups_controller.rb
+++ b/app/controllers/api/v1/dynamic_field_groups_controller.rb
@@ -15,7 +15,7 @@ module Api
         @dynamic_field_group.created_by = current_user
         @dynamic_field_group.updated_by = current_user
 
-        @dynamic_field_group.export_rules_attributes = export_rules_attributes
+        @dynamic_field_group.export_rules_attributes = export_rules_params
 
         if @dynamic_field_group.save
           render json: { dynamic_field_group: @dynamic_field_group }, status: :created
@@ -28,7 +28,7 @@ module Api
       def update
         @dynamic_field_group.updated_by = current_user
 
-        @dynamic_field_group.export_rules_attributes = export_rules_attributes
+        @dynamic_field_group.export_rules_attributes = export_rules_params
 
         if @dynamic_field_group.update(update_params)
           render json: { dynamic_field_group: @dynamic_field_group }, status: :ok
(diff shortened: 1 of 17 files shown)


cgalarza:
@elohanlon, I strayed from the names we discussed a little bit, so let me know if you think another naming scheme makes sense. I am not sure where we should create new export_rules when a new field_export_profile is added. We can assume the front end will know what all the field_export_profiles are, or we can create any necessary new export_rules when an object is shown, created and updated.

elohanlon:
My current preference is to have the front end will know what all the field_export_profiles are, and populate the front end form with defaults for any new profile so that those are updated when a DynamicFieldGroup is saved, but we can see how we feel when we get to the UI step.

app/controllers/api/v1/dynamic_field_groups_controller.rb
```diff
@@ -46,15 +50,23 @@ def destroy
...
         end
 
         def update_params
           params.require(:dynamic_field_group).permit(
-            :display_label, :is_repeatable, :xml_translation, :sort_order, :parent_type, :parent_id
+            :display_label, :is_repeatable, :sort_order, :parent_type, :parent_id
           )
         end
+
+        def export_rules_attributes
```
elohanlon: I like your solution for making used of Rails nested attributes even when the public API doesn't expose the word "attributes", but I feel a little weird about having this method called `export_rules_attributes` since that ties the extraction of the param data to a Rails concept that doesn't have to do with param parsing.  Do you mind if we rename the method to `export_rules`, but still use that method in the way you're currently using it in the create and update methods?

db/migrate/20190404201812_create_field_export_rules.rb
```diff
@@ -0,0 +1,17 @@
+class CreateFieldExportRules < ActiveRecord::Migration[5.2]
```
elohanlon: How would you feel about changing this migration name to `CreateFieldExportProfilesAndExportRules`?

db/migrate/20190408195208_add_unique_index_to_export_rules.rb
```diff
@@ -0,0 +1,5 @@
+class AddUniqueIndexToExportRules < ActiveRecord::Migration[5.2]
```
elohanlon: Okay to combine this migration into the migration file that creates the `export_rules` table, just to have fewer migrations?

spec/models/export_rule_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+        )
+      end
+
+      it 'does not save' do
+        expect(export_rule_2.save).to be false
+      end
+
+      it 'returns correct error' do
+        export_rule_2.save
+        expect(export_rule_2.errors.full_messages).to include 'Dynamic field group has already been taken'
```
elohanlon: I think this is a case where an overridden error message make sense, since this doesn't fully describe the error to the user.  Maybe something like this? "Dynamic field group already has an export rule for field_export_profile #{field_export_profile.name}" ?

spec/requests/api/v1/dynamic_field_groups_spec.rb
```diff
@@ -1,6 +1,10 @@
 require 'rails_helper'
 
 RSpec.describe 'Dynamic Field Groups Requests', type: :request do
+  let(:field_export_profile) { FactoryBot.create(:field_export_profile) }
+
+  before { field_export_profile }
```
elohanlon: Can you use `let!` in the let statement above instead of having this before block?

spec/requests/api/v1/dynamic_field_groups_spec.rb
```diff
@@ -131,6 +136,17 @@
...
+    let(:export_rule) do
+      FactoryBot.create(
+        :export_rule,
+        dynamic_field_group: dynamic_field_group,
+        field_export_profile: field_export_profile
+      )
+    end
+
+    before do
+      dynamic_field_group.export_rules << export_rule
```
elohanlon: Is it necessary to add the export_rule to the dynamic_field_group's export_rules if you're running `create` on the export_rule above, and setting `dynamic_field_group: dynamic_field_group`?

spec/requests/api/v1/field_export_profiles_spec.rb
```diff
@@ -0,0 +1,213 @@
...
+      let(:request) { get '/api/v1/field_export_profiles' }
+    end
+
+    context 'when logged in user is an administrator' do
+      before { sign_in_user as: :administrator }
+
+      context 'when there are multiple results' do
+        before { get '/api/v1/field_export_profiles' }
+
+        it 'returns all projects' do
```
elohanlon:
```suggestion
        it 'returns all field export profiles' do
```

cgalarza:
@elohanlon I think that a good way to go. If we find it too clunky when we are working with the UI we can think about it more then.

spec/requests/api/v1/dynamic_field_groups_spec.rb
cgalarza: Whoa, I totally forgot about `let!`.

Will do.

spec/models/export_rule_spec.rb
cgalarza: Sure. I was thinking about it, but could not think of a good error message. I'll update!

db/migrate/20190408195208_add_unique_index_to_export_rules.rb
cgalarza: I would rather not do that because these are part of different commits. All the code regarding the addition of the unique index is isolated into one commit.

db/migrate/20190404201812_create_field_export_rules.rb
cgalarza: I don't feel like we need to change it. If you feel strongly about it, I can change it. I treat migration files as static once they are created.

app/controllers/api/v1/dynamic_field_groups_controller.rb
cgalarza: I like the current name because it makes it really clear were the parameters are being set on the instance of dynamic_fields. I wouldn't want to change it to `export_rules` because that makes it seem like its returning an array of export_rule objects. We can change it to `export_rule_params` if that's better.

spec/requests/api/v1/dynamic_field_groups_spec.rb
cgalarza: You're right. I should structure this test differently. Will fix.

db/migrate/20190404201812_create_field_export_rules.rb
elohanlon: I'm good with keeping as is then.

db/migrate/20190408195208_add_unique_index_to_export_rules.rb
elohanlon: 👍

app/controllers/api/v1/dynamic_field_groups_controller.rb
cgalarza: I merged both of this methods, after our discussion on chat

PR #99, opened 2019-04-11
Project Permissions

cgalarza:
- Changing to using CanCanCan DSL to assign permissions in ability
- Updating tests for project sub controllers
- Using symbols to check for authorization when controllers don't have a model backing it

Disclaimer: This could probably use more granular testing, but I can't look at it anymore. Thought I would pass it on to you guys to get your thoughts.

### DIFF
--- a/app/controllers/api/v1/groups_controller.rb
+++ b/app/controllers/api/v1/groups_controller.rb
@@ -49,11 +49,11 @@ module Api
         end
 
         new_permissions = update_params.fetch(:permissions, []).map do |perm|
-          if !Permission::SYSTEM_WIDE_PERMISSIONS.include?(perm)
+          if Permission.valid_system_wide_permission?(perm)
+            Permission.new(action: perm)
+          else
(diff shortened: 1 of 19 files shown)


app/controllers/api/v1/groups_controller.rb
```diff
@@ -49,7 +49,7 @@ def update
         end
 
         new_permissions = update_params.fetch(:permissions, []).map do |perm|
-          if !Permission::VALID_SYSTEM_WIDE_PERMISSIONS.include?(perm)
+          if !Permission::SYSTEM_WIDE_PERMISSIONS.include?(perm)
```
elohanlon: How would you feel about creating a static method on the Permission class like `Permission.valid_system_wide_permission?(perm)`?

cgalarza: I think that's a good idea. It will help clean up the code some. I'll add it in. Thanks!

PR #103, opened 2019-04-19
Hyc3 fedora

barmintor:
I'm working on the IT, but the mock-driven tests are in:
```bash
Coverage report generated for RSpec to /Users/benjamin/github/cul/hyacinth/coverage. 5008 / 5226 LOC (95.83%) covered.
```
so if you want to start reviewing you can. There's a lot.


.gitignore
```diff
@@ -44,3 +44,10 @@ yarn-debug.log*
 database.yml
 hyacinth.yml
 resque.yml
+solr.yml
```
elohanlon: Can we remove solr.yml from the gitignore if it's not being used?

barmintor: Can we wait to make that change until after Hyc3 is out? I actually have to switch back and forth between the 2.x and 3.x branches a lot right now.

elohanlon: Sure, but on that note I'd like to **highly** recommend checking out two copies of the repository, in different directories called hyacinth2 and hyacinth3.  I've had to deal with the same challenge and it's much easier that way.

Gemfile
```diff
@@ -41,6 +41,14 @@ gem 'rainbow', '~> 3.0'
 
 gem 'uri_service-client'
 
+gem 'best_type', '0.0.4'
+
+# Fedora 3 related gems
+gem 'rubydora'
+gem 'rest-client'
```
elohanlon: I'm not saying we need to change this now if it's complicated to do, but I'd like to recommend using Faraday rather than rest-client.  It's already being pulled in indirectly by the uri_service-client gem, and I believe the samvera community prefers it these days too.

app/models/concerns/digital_object_concerns/export_fields_behavior.rb
```diff
@@ -0,0 +1,32 @@
+module DigitalObjectConcerns
+  module ExportFieldsBehavior
+    extend ActiveSupport::Concern
+
+    # Exports dynamic afield data from the digital_object to string data per
```
elohanlon:
```suggestion
    # Exports dynamic field data from the digital_object to string data per
```

Gemfile
barmintor: rubydora uses rest-client

elohanlon: In that case, why are we pulling it in explicitly?

barmintor: TBh I don't remember. Does it run without it?

elohanlon: I haven't used rubydora before, so I don't know, but I'm guessing yes because rest-client is included in rubydora's gemspec: https://github.com/samvera/rubydora/blob/master/rubydora.gemspec#L20

barmintor: PR has removed this from the gemfile

config/templates/jetty.template.yml
```diff
@@ -0,0 +1,18 @@
+# Note: Just because we're setting startup_wait to 75,
```
elohanlon:
```suggestion
# Note: Just because we're setting startup_wait to a high value,
```

config/templates/jetty.template.yml
```diff
@@ -0,0 +1,18 @@
...
+# that doesn't mean that startup will always take 40 seconds.
```
elohanlon:
```suggestion
# that doesn't mean that startup will always take that long.
```

config/templates/jetty.template.yml
```diff
@@ -0,0 +1,18 @@
...
+# That just means that we'll allow up to 75 seconds for startup.
```
elohanlon:
```suggestion
# That just means that we'll allow up to that many seconds for startup.
```

lib/hyacinth/adapters/preservation_adapter/fedora3.rb
```diff
@@ -13,30 +25,68 @@ def initialize(adapter_config = {})
...
+          fedora_object.datastreams[HYACINTH_CORE_DATASTREAM_NAME].content =
+            JSON.generate(digital_object.to_serialized_form)
+          # serialize the other datastreams
+          FieldExportProfile.all.each do |profile|
+            assign(datastream_for(profile)).from(digital_object).to(fedora_object)
+          end
+
+          assign(Fedora3::ObjectProperties).from(digital_object).to(fedora_object)
+          assign(Fedora3::DCProperties).from(digital_object).to(fedora_object)
+          # TODO: add rel as appropriate
```
elohanlon: Can this TODO go away?

lib/hyacinth/adapters/preservation_adapter/fedora3.rb
```diff
@@ -13,30 +25,68 @@ def initialize(adapter_config = {})
...
+          assign(Fedora3::RelsExtProperties).from(digital_object).to(fedora_object)
+          assign(Fedora3::RelsIntProperties).from(digital_object).to(fedora_object)
+          # TODO: export structMetadata
```
elohanlon: Can this TODO go away?

lib/hyacinth/adapters/preservation_adapter/fedora3/dc_properties.rb
```diff
@@ -0,0 +1,79 @@
...
+
+        def proposed_dc_properties(hyacinth_obj = @hyacinth_obj)
+          properties = {}
+          # set the title
+          properties[:title] = [get_title(hyacinth_obj.dynamic_field_data)]
+          # set the identifiers
+          properties[:identifier] = hyacinth_obj.identifiers&.to_a || []
+          # add the pid as an identifier
+          properties[:identifier].concat Array.wrap(digital_object_pids(hyacinth_obj))
+          if hyacinth_obj.digital_object_type.casecmp('asset').zero?
```
elohanlon: Are you concerned here that when digital_object_type returns "asset" here, that "asset" may not always be all lower case? (thus the use of `casecmp('asset').zero?`)

barmintor: Me a month ago is a different person, really - if you think it's unclear or unnecessary, we can just delete it.

lib/hyacinth/adapters/preservation_adapter/fedora3/dc_properties.rb
```diff
@@ -0,0 +1,79 @@
...
+        def parse_dc_xml(fedora_obj)
+          properties = Hash.new { |hash, key| hash[key] = [] }
+          ng_xml = Nokogiri::XML(fedora_obj.datastreams['DC'].content)
+          ng_xml.remove_namespaces!
+          dc = ng_xml.at_css("dc")
+          dc.elements.each { |ele| properties[ele.name.to_sym] << ele.text.strip }
+          normal_properties(properties)
+        end
+
+        def normal_properties(properties = {})
```
elohanlon: Okay to rename this to `normalized_properties`?

lib/hyacinth/adapters/preservation_adapter/fedora3/pid_helpers.rb
```diff
@@ -0,0 +1,12 @@
+module Hyacinth
+  module Adapters
+    module PreservationAdapter
+      module Fedora3::PidHelpers
+        def digital_object_pids(hyacinth_obj)
```
elohanlon:
```suggestion
        def digital_object_fedora_pid_uris(hyacinth_obj)
```
Suggesting this change because this method returns fedora object URIs rather than PIDs.

lib/hyacinth/adapters/preservation_adapter/fedora3/rels_ext_properties.rb
```diff
@@ -0,0 +1,145 @@
...
+            digital_object_pids(parent)
+          end.flatten.compact
+        end
+
+        def project_string_keys_for(hyacinth_obj)
+          hyacinth_obj.projects.map(&:string_key)
+        end
+
+        def publisher_string_keys_for(hyacinth_obj)
+          hyacinth_obj.publish_entries.map { |string_key, _entry| string_key }
```
elohanlon: It will be so nice on the DLC management side to use string keys rather than pids for publisher.

lib/hyacinth/adapters/preservation_adapter/fedora3/rels_ext_properties.rb
```diff
@@ -0,0 +1,145 @@
...
+          type_infos_for(hyacinth_obj, :cmodel)
+        end
+
+        def rdf_types_for(hyacinth_obj)
+          type_infos_for(hyacinth_obj, :rdf_type)
+        end
+
+        def type_infos_for(hyacinth_obj, type_key)
+          infos = []
+          return infos unless (type_info = TYPE_INFORMATION[hyacinth_obj.digital_object_type.downcase])
```
elohanlon: I get why using downcase here would make sense if we're unsure of the format of digital_object_type, but it should be consistently all lower case, and I want to be able to assume that consistency across all occurrences across the app, otherwise we'd need to call `downcase` everywhere it's referenced, you know?.  So maybe okay to drop `downcase` here?

lib/hyacinth/adapters/preservation_adapter/fedora3/rels_ext_properties.rb
```diff
@@ -0,0 +1,145 @@
...
+          infos << type_info[type_key]
+          infos
+        end
+
+        def restrictions_for(hyacinth_obj)
+          restrictions = []
+          digital_object_data = hyacinth_obj.dynamic_field_data
+          return restrictions unless digital_object_data.key?('restrictions')
+          restrictions << ONSITE_RESTRICTION_LITERAL_VALUE if digital_object_data['restrictions'].key?('restricted_onsite')
```
elohanlon: I think you were basing this on Hyacinth 2 code, and if so this looks like a [Hyacinth 2] bug to me.  `restricted_onsite` and `restricted_size_image` boolean properties shouldn't behave differently.  We'd want to check for true/false in both, but we're only checking for the presence of the key for restricted_onsite.

lib/hyacinth/adapters/preservation_adapter/fedora3/rels_ext_properties.rb
```diff
@@ -0,0 +1,145 @@
...
+          restrictions << SIZE_RESTRICTION_LITERAL_VALUE if digital_object_data['restrictions']['restricted_size_image']&.casecmp('false')&.zero?
```
elohanlon: Are we expecting `digital_object_data['restrictions']['restricted_size_image']` to hold a string, or a boolean value?  I thought boolean.

lib/hyacinth/adapters/preservation_adapter/fedora3/rels_ext_properties.rb
```diff
@@ -0,0 +1,145 @@
...
+        def type_infos_for(hyacinth_obj, type_key)
+          infos = []
+          return infos unless (type_info = TYPE_INFORMATION[hyacinth_obj.digital_object_type.downcase])
+          infos << type_info[type_key]
+          infos
+        end
+
+        def restrictions_for(hyacinth_obj)
+          restrictions = []
+          digital_object_data = hyacinth_obj.dynamic_field_data
```
elohanlon: The term `digital_object_data` indicates top level data, while `dynamic_field_data` is a subfield of that top level data.  We should definitely change this line for that reason, since it's confusing.

Also, you won't find `restrictions` under `hyacinth_obj.dynamic_field_data`, since they're not user-defined dynamic fields.  We need to add a top level `restrictions` metadata attribute to the digital object base model.

lib/hyacinth/digital_object/asset_type.rb
```diff
@@ -0,0 +1,25 @@
+module Hyacinth
+  module DigitalObject
+    module AssetType
```
elohanlon: We may have to update the best_type gem to accommodate some of these, but if so, that's a good improvement.

lib/hyacinth/xml_generator/element.rb
```diff
@@ -0,0 +1,116 @@
+module Hyacinth
+  class XMLGenerator::Element
+    attr_reader :generator, :field_values
+    delegate :generate_field_val, :value_with_substitutions, :value_for_field_name,
+             :generate_field_val, :render_output_of_ternary, :render_output_of_join,
```
elohanlon:
```suggestion
             :render_output_of_ternary, :render_output_of_join,
```
`:generate_field_val` was in there twice.

lib/hyacinth/adapters/preservation_adapter/fedora3/dc_properties.rb
elohanlon: I believe you!  Deleting sounds good, thanks.

lib/tasks/hyacinth/ci.rake
```diff
@@ -1,7 +1,13 @@
+FEDORA_JETTY_ZIP_BASENAME = 'hyacinth-fedora-3.8.1-no-solr'.freeze
```
elohanlon: This isn't all that important for the PR, but I'm thinking that this constant should probably just be a full URL.  I don't think there's any real benefit to storing the basename separately here.

spec/hyacinth/adapters/preservation_adapter/fedora3/system_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+    context "field exports" do
+      let(:digital_object_title) { "Assigned Label" }
+      before do
+        FactoryBot.create(:export_rule) # creates descMetadata
+      end
+      it "persists templated field exports to datastreams" do
+        hyacinth_object.dynamic_field_data['name'] = [{ 'role' => "Farmer" }]
+        adapter.persist_impl("fedora3://#{object_pid}", hyacinth_object)
+        actual_xml = rubydora_object.datastreams["descMetadata"].content
+        actual_xml.sub!(/^<\?.+\?>/, '') # remove PI
```
elohanlon: This will probably seem really obvious after you tell me the answer, but what does PI stand for?

elohanlon: Is it https://en.wikipedia.org/wiki/Processing_Instruction ?  My XSLT experience is pretty limited, so I wasn't familiar with that term (and I wouldn't be surprised if other devs on our team might not know it either).  It looks like technically the XML Declaration isn't a processing instruction, so if that's what you meant do you mind if we change the comment to # remove XML declaration?

spec/hyacinth/adapters/preservation_adapter/fedora3/system_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+        actual_xml.gsub!('mods:', '') # remove ns
+        actual_xml.gsub!(/\s/, '') # remove whitespace
+        expect(actual_xml).to eql("<mods><name>Farmer</name></mods>")
+      end
+      it "makes a core object property assignments" do
```
elohanlon:
```suggestion
      it "makes a core object property assignment" do
```

elohanlon: Is this test meant to prove the successful assignment of one property or two (with state being implied as the second one, thus the `rubydora_object.state` equals `'A'` check)?  The test label threw me off, and I wasn't sure if you meant to include a `rubydora_object.state` check in this test.

spec/hyacinth/adapters/preservation_adapter/fedora3/system_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+        {
+          predicate: "http://www.loc.gov/premis/rdf/v1#hasRestriction",
+          object: "onsite restriction",
+          pid: object_pid,
+          isLiteral: true
+        }
+      end
+
+      it "persists model properties" do
+        hyacinth_object.dynamic_field_data['restrictions'] = { 'restricted_onsite' => true }
```
elohanlon: I mentioned this in an earlier comment too, but 'restrictions' shouldn't appear under dynamic_field data.  Even if the test is technically valid, it could be confusing to others.

spec/hyacinth/adapters/preservation_adapter/fedora3_spec.rb
```diff
@@ -0,0 +1,283 @@
...
+
+      before do
+        hyacinth_object.structured_children['structure'] << child_uid
+        expect(connection).to receive(:datastream_profile).with(object_pid, described_class::HYACINTH_CORE_DATASTREAM_NAME, nil, nil).and_return({})
+        allow(connection).to receive(:add_relationship) # tested elsewhere
+        expect(::DigitalObject::Base).to receive(:find).with(child_uid).and_return(child_hyacinth_object)
+      end
+
+      it "persists model properties" do
+        hyacinth_object.identifiers << 'keep'
```
elohanlon: I can't tell if this line is necessary for this test.  Is it?  Or copypasta?

spec/hyacinth/adapters/preservation_adapter/fedora3/system_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+        obj
+      end
+      let(:child_uid) { child_hyacinth_object.uid }
+
+      before do
+        hyacinth_object.structured_children['structure'] << child_uid
+      end
+
+      it "persists model properties" do
+        hyacinth_object.identifiers << 'keep'
```
elohanlon: I can't tell if this line is necessary for this test. Is it? Or copypasta?

spec/hyacinth/adapters/preservation_adapter/fedora3/system_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+        # because .save is stubbed, the datastream should still be dirty
+        expect(rubydora_object.datastreams['structMetadata'].content).to include("<mets:div LABEL=\"#{child_object_title}\" ORDER=\"1\" CONTENTIDS=\"#{child_uid}\"/>")
+      end
+    end
+    context "RelsInt properties for resources" do
+      let(:dsids) { ['structMetadata'] }
+      let(:hyacinth_object) { DigitalObject::Asset.new }
+      let(:resource_args) { { original_filename: '/old/path/to/file.doc', location: '/path/to/file.doc', checksum: 'asdf', file_size: 'asdf' } }
+      before do
+        hyacinth_object.resources['master'].send(:initialize, resource_args)
```
elohanlon: It looks like you're re-running the initialize method for an already-initialized object?  If this is a thing that people do, that's cool to know, but since I haven't seen it before it somehow feels wrong.  @cgalarza, out of curiosity, have you ever seen that done before?  If not, I thought you'd find this as interesting as I did.

spec/hyacinth/adapters/preservation_adapter/fedora3/system_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+      end
+      it "persists model properties" do
+        adapter.persist_impl("fedora3://#{object_pid}", hyacinth_object)
+        # obj = Rubydora::DigitalObject.new(object_pid, Rubydora.repository)
```
elohanlon: delete?

spec/hyacinth/adapters/preservation_adapter/fedora3_spec.rb
```diff
@@ -0,0 +1,283 @@
...
+      it "persists templated field exports to datastreams" do
+        hyacinth_object.dynamic_field_data['name'] = [{ 'role' => "Farmer" }]
+        adapter.persist_impl("fedora3://#{object_pid}", hyacinth_object)
+        actual_xml = rubydora_object.datastreams["descMetadata"].content
+        actual_xml.sub!(/^<\?.+\?>/, '') # remove PI
+        actual_xml.gsub!('mods:', '') # remove ns
+        actual_xml.gsub!(/\s/, '') # remove whitespace
+        expect(actual_xml).to eql("<mods><name>Farmer</name></mods>")
+      end
+      it "makes a core object property assignments" do
```
elohanlon: See earlier comment about same-name `it "makes a core object property assignments" do` test.

spec/hyacinth/adapters/preservation_adapter/fedora3_spec.rb
```diff
@@ -0,0 +1,283 @@
...
+        }
+      end
+      before do
+        expect(connection).to receive(:datastream_profile).with(object_pid, described_class::HYACINTH_CORE_DATASTREAM_NAME, nil, nil).and_return({})
+        expect(connection).to receive(:add_relationship).with(model_property)
+        expect(connection).to receive(:add_relationship).with(rdftype_property)
+        expect(connection).to receive(:add_relationship).with(restriction_property)
+      end
+      it "persists model properties" do
+        hyacinth_object.dynamic_field_data['restrictions'] = { 'restricted_onsite' => true }
```
elohanlon: See earlier comment about 'restrictions' not belonging inside dynamic_field_data.

spec/hyacinth/digital_object/resource_spec.rb
```diff
@@ -0,0 +1,46 @@
+require 'rails_helper'
+
+RSpec.describe Hyacinth::DigitalObject::Resource do
+  let(:argument_names) { [:import_location, :import_method, :import_checksum, :location, :checksum, :original_filename, :file_size] }
+  let(:arguments) { argument_names.map { |s| [s.to_s, s.to_s] }.to_h }
+  let(:attribute_names) { argument_names.select { |n| !n.to_s.starts_with?('import_') } }
```
elohanlon: Relying on the `import_` prefix for converting argument_names to attribute_names seems a little fragile.  This naming convention isn't enforced anywhere in the code.

elohanlon: Here's a suggestion that's unrelated to the above comment.  Since the `import_*` variables technically are attributes (because of their attr_accessor definitions in the Resource class), can we change `let(:attribute_names)` to `let(:serialized_attribute_names)` and `let(:attributes)` to `let(:serialized_attributes)`?

spec/noid/minter_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  it "correctly predicts the number of unique identifiers generated by a noid minter" do
+    minter = described_class.new(template: 'cul:.ree')
+    number_of_expected_mints = minter.template.max
+
+    number_of_expected_mints.times { minter.mint }
+    # The minter throws an exception when it runs out of mints
+    expect { minter.mint }.to raise_error(Exception)
+  end
+
+  it "doesn't mint duplicates" do
```
elohanlon: We might want to use a Set instead of a Hash for this test, but we can change that later.  (I realize this was just copied in from Hyacinth 2.)

lib/hyacinth/adapters/preservation_adapter/fedora3/rels_ext_properties.rb
barmintor: I think these use cases are going to diverge, so maybe leave this particular bit in for now?

spec/hyacinth/adapters/preservation_adapter/fedora3/system_spec.rb
barmintor: I mean, I don't recommend it in most cases - it's kind of a dirty trick

app/models/digital_object/base.rb
```diff
@@ -28,7 +28,8 @@ class Base
     metadata_attribute :serialization_version, Hyacinth::DigitalObject::TypeDef::String.new.default(-> { SERIALIZATION_VERSION })
     metadata_attribute :uid, Hyacinth::DigitalObject::TypeDef::String.new
     metadata_attribute :doi, Hyacinth::DigitalObject::TypeDef::String.new
-    metadata_attribute :digital_object_type, Hyacinth::DigitalObject::TypeDef::String.new
+    # constrain type to the keys for registered type classes
+    metadata_attribute :digital_object_type, Hyacinth::DigitalObject::TypeDef::String.new.constrained_to(Hyacinth.config.digital_object_types)
```
elohanlon: Can we make this a DigitalObject validation (in `DigitalObjectConcerns::Validations`) instead of something on the TypeDef object?  I originally had value constraints on the TypeDef and intentionally removed them because I didn't like having some validation in the TypeDef, but most other validations in `DigitalObjectConcerns::Validations`.  Note that validations are pretty much non-existent right now because adding them in `DigitalObjectConcerns::Validations` is a TODO.

app/models/digital_object/base.rb
```diff
@@ -27,7 +30,8 @@ class Base
...
```
elohanlon: Can we make this a DigitalObject validation (in DigitalObjectConcerns::Validations) instead of something on the TypeDef object? I originally had value constraints on the TypeDef and intentionally removed them because I didn't like having some validation in the TypeDef, but most other validations in DigitalObjectConcerns::Validations. Note that validations are pretty much non-existent right now because adding them in DigitalObjectConcerns::Validations is a TODO.

lib/hyacinth/digital_object/type_def/base.rb
```diff
@@ -37,6 +37,13 @@ def public_writer
           self # always return self to allow for chained calls
         end
 
+        # Constraint is checked by call to include?, so a values hash is verifying by key presence
+        def constrained_to(values = [])
```
elohanlon: Going along with my earlier comment, I think we'd want to remove this in favor of `DigitalObjectConcerns::Validations`.

elohanlon: It also doesn't protect against `instance_variable_set`, but a validation would.

lib/hyacinth/adapters/preservation_adapter/fedora3/dc_properties.rb
```diff
@@ -0,0 +1,85 @@
...
+              xml << "<dc:#{property}>#{value}</dc:#{property}>"
+            end
+          end
+          xml << XML_SUFFIX
+        end
+
+        private
+
+          def asset_type?(hyacinth_obj)
+            hyacinth_obj.respond_to? :asset_type
```
elohanlon: Maybe you'd be better off checking if `hyacinth_obj.is_a?(DigitalObject::Asset)` ?

lib/hyacinth/adapters/preservation_adapter/fedora3/rels_ext_properties.rb
```diff
@@ -0,0 +1,144 @@
...
+          infos = []
+          return infos unless (type_info = TYPE_INFORMATION[hyacinth_obj.digital_object_type])
+          infos << type_info[type_key]
+          infos
+        end
+
+        def restrictions_for(hyacinth_obj)
+          restrictions = []
+          restriction_data = hyacinth_obj.restrictions
+          restrictions << ONSITE_RESTRICTION_LITERAL_VALUE if restriction_data['restricted_onsite'].to_s.casecmp('true')&.zero?
```
elohanlon: What you have here will work, but ideally a DigitalObject validation (which doesn't currently exist) would ensure that the restriction values are booleans, so you wouldn't need to cast/string-compare.  If we end up persisting these values as strings instead of booleans, it'll cause problems in other parts of the code unless we're casting and string-comparing everywhere --- but at that point it's just safer to validate for boolean and then assume boolean everywhere else.

lib/hyacinth/adapters/preservation_adapter/fedora3/dc_properties.rb
barmintor: I started that way, but I thought this was more extensible and more testable?

lib/hyacinth/digital_object/type_def/base.rb
elohanlon: Now I'm good with keeping constraints and going along with the validation method we talked about. 👍

app/models/concerns/digital_object_concerns/validations/digital_object_type_validator.rb
```diff
@@ -0,0 +1,6 @@
+class DigitalObjectConcerns::Validations::DigitalObjectTypeValidator < ActiveModel::Validator
+  def validate(digital_object)
+    return if Hyacinth.config.digital_object_types.include?(digital_object.digital_object_type)
+    record.errors[:digital_object_type] << "Unregistered digital object type #{digital_object.digital_object_type}"
```
elohanlon: Did you mean `digital_object` instead of `record`?

app/models/concerns/digital_object_concerns/validations/restrictions_validator.rb
```diff
@@ -0,0 +1,8 @@
+class DigitalObjectConcerns::Validations::RestrictionsValidator < ActiveModel::Validator
+  def validate(digital_object)
+    digital_object.restriction_attributes.each do |restriction_name, type_def|
+      value = digital_object.restrictions[restriction_name]
+      record.errors[:restrictions] << "Invalid #{restriction_name} value #{value}" unless type_def.valid?(value)
```
elohanlon: Did you mean `digital_object` instead of `record` here too?

lib/hyacinth/adapters/preservation_adapter/fedora3/rels_ext_properties.rb
```diff
@@ -114,8 +114,8 @@ def type_infos_for(hyacinth_obj, type_key)
         def restrictions_for(hyacinth_obj)
           restrictions = []
           restriction_data = hyacinth_obj.restrictions
-          restrictions << ONSITE_RESTRICTION_LITERAL_VALUE if restriction_data['restricted_onsite'].to_s.casecmp('true')&.zero?
-          restrictions << SIZE_RESTRICTION_LITERAL_VALUE if restriction_data['restricted_size_image'].to_s.casecmp('true')&.zero?
+          restrictions << ONSITE_RESTRICTION_LITERAL_VALUE if restriction_data['restricted_onsite'] == true
+          restrictions << SIZE_RESTRICTION_LITERAL_VALUE if restriction_data['restricted_size_image'] == true
```
elohanlon: Do you need `== true` at all in the lines above?

PR #110, opened 2019-05-10
implement Ability logic for digital object permissions in projects

barmintor:
begin authorizing digital_objects_controller

### DIFF
--- a/app/controllers/api/v1/digital_objects_controller.rb
+++ b/app/controllers/api/v1/digital_objects_controller.rb
@@ -2,7 +2,8 @@ module Api
   module V1
     class DigitalObjectsController < ApplicationApiController
       before_action :ensure_json_request
-      before_action :set_digital_object, only: [:show, :edit, :update, :destroy]
+      before_action :load_resource, only: [:show, :edit, :update, :destroy, :preserve, :publish]
+      authorize_resource :digital_object, only: [:show, :edit, :update, :destroy, :preserve, :publish]
 
       # GET /digital_objects/search
       # GET /digital_objects/search.json
@@ -22,12 +23,16 @@ module Api
       # POST /digital_objects
       # POST /digital_objects.json
       def create
-        digital_object_data = JSON.parse(create_or_update_params['digital_object_data_json'])
-        @digital_object = DigitalObject::Base.class_for_type(digital_object_data['digital_object_type']).new
-        @digital_object.set_digital_object_data(digital_object_data)
-
-        if @digital_object.save
-          render :show, status: :created, location: @digital_object
+        digital_object_data = create_or_update_params
+        digital_object_record = DigitalObjectRecord.create
+        @digital_object = DigitalObject::Base.from_serialized_form(digital_object_record, digital_object_data)
+        @digital_object.projects.each do |project|
+          authorize! :create_objects, project
+        end
+        @digital_object.state ||= Hyacinth::DigitalObject::State::ACTIVE
+        # digital_object_record.save first because it will assign necessary digital_object_record attributes
+        if @digital_object.save && (digital_object_record.persisted? || digital_object_record.save)
+          show
         else
           render json: @digital_object.errors, status: :unprocessable_entity
         end
@@ -36,15 +41,16 @@ module Api
       # PATCH/PUT /digital_objects/1
       # PATCH/PUT /digital_objects/1.json
       def update
-        if @digital_object.update(digital_object_params)
-          render :show, status: :ok, location: @digital_object
+        digital_object_data = create_or_update_params
+        @digital_object.set_digital_object_data(digital_object_data, true)
+        if @digital_object.save
+          show
         else
           render json: @digital_object.errors, status: :unprocessable_entity
         end
       end
 
       def preserve
-        @digital_object = DigitalObject::Base.find(publish_params[:uid])
         if @digital_object.preserve
           render :show, status: :created, location: @digital_object
         else
@@ -54,8 +60,6 @@ module Api
 
       # Publish the The publish action also preserves.
       def publish
-        @digital_object = DigitalObject::Base.find(publish_params[:uid])
-
         # TODO: One day, if publish targets don't need to be saved in the
         # preservation system, we may want the publish method to accept
         # params like publish_to, unpublish_from, and republish. For now,
@@ -71,18 +75,20 @@ module Api
 
       # DELETE /digital_objects/1
       # DELETE /digital_objects/1.json
+      # Unimplemented
       def destroy
         @digital_object.destroy
         head :no_content
       end
 
-      private
-        def set_digital_object
-          @digital_object = DigitalObject::Base.find(params[:uid])
-        end
+      def load_resource
+        @digital_object ||= DigitalObject::Base.find(params[:id])
+      end
 
+      private
         def create_or_update_params
-          params.require(:digital_object).permit(:digital_object_data_json)
+          # TODO: decide how we want to validate dynamic field data parameters
+          params[:digital_object][:digital_object_data_json]&.permit!.to_h
         end
     end
   end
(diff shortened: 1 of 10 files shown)


barmintor:
This PR implements digital object authorizations in the digital_objects controller, and also provides rudimentary implementations oseveral digtal objects CRUD methods

spec/models/ability_spec.rb
```diff
@@ -105,4 +105,82 @@
...
+          :user, permissions: [
+            Permission.create(action: :delete_objects, subject: Project.to_s, subject_id: authorized_project.id)
+          ]
+        )
+      end
+
+      it { is_expected.to be_able_to(:destroy, authorized_object) }
+      it { is_expected.to be_able_to(:delete_objects, authorized_project) }
+      it { is_expected.not_to be_able_to(:delete_objects, unauthorized_project) }
+      it { is_expected.to be_able_to(:destroy, mixed_object) }
```
cgalarza: The way we would use this permissions is that they would only be deleted from within one of the projects that a digital object is in?

spec/requests/api/v1/digital_objects_spec.rb
```diff
@@ -1,9 +1,135 @@
 require 'rails_helper'
 
-RSpec.describe "Digital Objects API endpoint", type: :controller do
+RSpec.describe "Digital Objects API endpoint", type: :request do
+  let(:authorized_object) { FactoryBot.create(:digital_object_test_subclass, :with_sample_data, :with_lincoln_project) }
+  let(:authorized_project) { authorized_object.projects.to_a[0] }
```
cgalarza: Can this be `authorized_object.projects.first`?

spec/requests/api/v1/digital_objects_spec.rb
```diff
@@ -1,9 +1,135 @@
...
+      before do
+        sign_in_project_contributor to: :read_objects, project: authorized_project
+        get "/api/v1/digital_objects/#{authorized_object.uid}"
+      end
+
+      it 'returns 200' do
+        expect(response.status).to be 200
+      end
+      it "return a single Digital Object with the expected fields" do
+        expect(JSON.parse(response.body)).to include('uid' => authorized_object.uid)
```
cgalarza: I've been using `json_spec` to do the json matching in case that's easier.

https://github.com/collectiveidea/json_spec

spec/requests/api/v1/digital_objects_spec.rb
```diff
@@ -1,9 +1,135 @@
...
+      before do
+        sign_in_project_contributor to: :update_objects, project: authorized_project
+        patch "/api/v1/digital_objects/#{authorized_object.uid}", params: properties
+      end
+
+      it 'returns 200' do
+        expect(response.status).to be 200
+      end
+      it "return a single Digital Object with the expected fields" do
+        expect(JSON.parse(response.body)).to include(properties[:digital_object][:digital_object_data_json])
```
cgalarza: Should we be checking that the update changed the data?

spec/requests/api/v1/digital_objects_spec.rb
```diff
@@ -1,9 +1,135 @@
...
+    context "logged in" do
+      before do
+        sign_in_project_contributor to: :create_objects, project: authorized_project
+        post "/api/v1/digital_objects", params: properties
+      end
+
+      it 'returns 200' do
+        expect(response.status).to be 200
+      end
+      it "return a single Digital Object with the expected fields" do
```
cgalarza: json_spec might make it easier to check for these.

spec/requests/api/v1/digital_objects_spec.rb
```diff
@@ -1,9 +1,135 @@
...
+      end
+    end
+
+    context "logged in" do
+      before do
+        sign_in_project_contributor to: [:delete_objects, :read_objects], project: authorized_project
+        delete "/api/v1/digital_objects/#{authorized_object.uid}"
+      end
+
+      xit 'returns 200' do
```
cgalarza: Are you skipping these on purpose?

app/controllers/api/v1/digital_objects_controller.rb
```diff
@@ -22,12 +23,16 @@ def show
...
+        @digital_object = DigitalObject::Base.from_serialized_form(digital_object_record, digital_object_data)
+        @digital_object.projects.each do |project|
+          authorize! :create_objects, project
+        end
+        @digital_object.state ||= Hyacinth::DigitalObject::State::ACTIVE
+        # digital_object_record.save first because it will assign necessary digital_object_record attributes
+        if @digital_object.save && (digital_object_record.persisted? || digital_object_record.save)
+          show
         else
           render json: @digital_object.errors, status: :unprocessable_entity
```
cgalarza: I've been wrapping these in an `errors` function that's defined in `ApplicationApiController`. Though the digital object might have us re-think how we display the errors in the json. We can save this for later.

app/models/ability.rb
```diff
@@ -84,6 +78,60 @@ def to_list
     end
   end
 
+  def manage(project_id, project_string_key)
```
cgalarza: This looks great!

spec/requests/api/v1/digital_objects_spec.rb
barmintor: Yes - DigitalObject::Base doesn't have a delete method yet (I don't think) so...

spec/models/ability_spec.rb
barmintor: The spec is deceptive because it's only in one project - my intention here was about object deletion and not relationship breaking. But should that be checked when an update removes a project? Tricky. Here's another question: Should this be more like create - should you have to have permission to delete in all the projects?

spec/requests/api/v1/digital_objects_spec.rb
barmintor: Yes!

spec/requests/api/v1/digital_objects_spec.rb
barmintor: Will read up on it!

spec/requests/api/v1/digital_objects_spec.rb
barmintor: I think it does - the existing object has a different title, and the local properties changes it.

barmintor: Maybe I should comment that?

app/models/ability.rb
barmintor: That's a relief! Let me push a couple of minor changes in response to your feedback.

spec/models/ability_spec.rb
cgalarza: If an object is in two projects, I don't think we should allow a user with only permission to delete from one project to delete it. Maybe we should treat a delete for one project as the relationship breaking and then actually delete the object if they aren't pointing to any more projects. In a way, the manage_all_digital_objects permissions encompasses the delete_all permission. Should only people with those permissions be able to delete objects?

cgalarza: I agree, that requiring users to have permission to delete in all the projects an object is linked to should be sufficient. 

Yea, the relationship breaking could just be an update. Maybe just having a "Remove" button that sends off an update request. 

I still think we would want to delete an object if it's "Removed" from all the projects its a part of. Or say that a object can't be remove from the last project its a part of.

barmintor: The last-project question is a sticky one - part of me likes the idea of not letting the last one be removed, but... it also seems like a not-very-user-managed data at that point. I'm reluctant to delete when it isn't explicitly requested, but then should there be an admin-only orphans project?

cgalarza: I don't like the idea of having objects that arent managed by a project, it could be hard to find out where an orphaned object came from and what should be done with it.

I think we should not allow the remove from the last project available, because at that point it should be a deletion and only the users with the correct permissions would have that option.

PR #111, opened 2019-05-14
initial implementation of destroy and publish in digital objects API

barmintor:


lib/hyacinth/adapters/publication_adapter/hyacinth2.rb
```diff
@@ -0,0 +1,55 @@
...
+          update_doi(digital_object, location) if point_doi_to_this_publish_target && location
+          true
+        rescue StandardError => e
+          [false, e.message]
+        end
+
+        # @return [success, errors] [boolean, Array<String>] success will be true if
+        #         the unpublish was successful, or false otherwise. errors is an array
+        #         that will contain error messages if the unpublish failed.
+        def unpublish(publish_target, digital_object)
```
cgalarza: I think this method is missing updating the DOI entry if the digital object is unpublished from the "primary" publish target.

lib/hyacinth/adapters/publication_adapter/hyacinth2.rb
```diff
@@ -0,0 +1,55 @@
...
+          digital_object_pid = digital_object_pids(digital_object).first
+          return [false, "Never preserved to Fedora3"] unless digital_object_pid
+          connection = Faraday.new(publish_target.publish_url)
+          connection.token_auth(publish_target.api_key)
+          connection.delete(digital_object_pid)
+          true
```
cgalarza: Should this also be an array to follow the pattern?

I have to be honest I don't love sending the success and error message back in an array like this, but right now I can't think of a better solution. Maybe using a hash? I'll keep thinking about it.

spec/requests/api/v1/digital_objects_spec.rb
```diff
@@ -118,18 +120,71 @@
...
+  describe 'POST /api/v1/digital_objects/:id/publish' do
+    include_examples 'requires user to have correct permissions' do
+      let(:request) do
+        post "/api/v1/digital_objects/#{authorized_object.uid}/publish"
+      end
+    end
+
+    context "logged in" do
+      let(:publication_adapter) { instance_double(Hyacinth::Adapters::PublicationAdapter::Abstract) }
+      let(:configured_publication_adapter) { Hyacinth.config.publication_adapter }
```
cgalarza: Can use `let!` here to automatically run this.

spec/requests/api/v1/digital_objects_spec.rb
```diff
@@ -118,18 +120,71 @@
...
+      end
+      it 'unpublishes the object on delete' do
+        expect(publication_adapter).to receive(:publish).with(authorized_publish_target, digital_object_matcher, anything).and_return([true, []])
+        post "/api/v1/digital_objects/#{authorized_object.uid}/publish"
+      end
+      it "return json entity" do
+        expect(publication_adapter).to receive(:publish).with(authorized_publish_target, digital_object_matcher, anything).and_return([true, []])
+        post "/api/v1/digital_objects/#{authorized_object.uid}/publish"
+        new_object = JSON.parse(response.body)
+        expect(new_object['projects'].first).to include('string_key' => authorized_project.string_key)
```
cgalarza: Should this be checking that the publish target on the object was set correctly?

spec/requests/api/v1/digital_objects_spec.rb
barmintor: that's awesome & new to me

lib/hyacinth/adapters/publication_adapter/hyacinth2.rb
barmintor: Yes, to all of this.

app/models/concerns/digital_object_concerns/destroy_behavior.rb
```diff
@@ -0,0 +1,55 @@
...
+          self.digital_object_record.optimistic_lock_token = self.mint_optimistic_lock_token
+
+          # remove metadata storage
+          self.remove_from_metadata_storage
+          all_targets = self.projects.map do |project|
+            project.publish_targets.map(&:string_key)
+          end.flatten.uniq
+          self.set_pending_publish_entries({ 'publish_to' => [], 'unpublish_from' => all_targets })
+          # if everything worked, destroy digital object record
+          self.digital_object_record.destroy!
```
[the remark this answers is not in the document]
barmintor: there's not really an indexing service yet... I'll need to TODO it

app/models/concerns/digital_object_concerns/destroy_behavior.rb
```diff
@@ -0,0 +1,55 @@
+module DigitalObjectConcerns
+  module DestroyBehavior
+    extend ActiveSupport::Concern
+
+    # This method is like the other #save method, but it raises an error if the save fails.
```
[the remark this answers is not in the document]
barmintor: Yep.

app/models/concerns/digital_object_concerns/destroy_behavior.rb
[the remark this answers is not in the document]
barmintor: You know I'll just add the stub call to this and the save behavior.

app/models/concerns/digital_object_concerns/publish_behavior.rb
```diff
@@ -82,6 +65,31 @@ def publish(opts = {})
       self.errors.blank?
     end
 
+    def publish_to(publish_target, current_publish_entries, opts = {})
```
cgalarza: Giant 👍

PR #112, opened 2019-05-30
3.x doi update

barmintor:


app/models/concerns/digital_object_concerns/save_behavior.rb
```diff
@@ -52,6 +52,7 @@ def save_impl(opts = {})
         begin
           self.mint_uid_and_metadata_location_uri_if_new_record
           self.update_modification_info(current_datetime, opts[:user])
+          self.mint_reserved_doi_if_doi_blank
```
elohanlon: If something goes wrong during the save, the digital object's doi field will be reverted properly because of the existing code below, but I think the DOI will still exist indefinitely in a reserved state in the external DOI system.

app/models/publish_target.rb
```diff
@@ -10,12 +10,21 @@ def as_json(_options = {})
...
-      api_key: api_key
+      api_key: api_key,
+      is_allowed_doi_target: is_allowed_doi_target,
+      doi_priority: doi_priority
     }
   end
 
-  # TODO: Remove when DOI service is implemented
+  def from_json(json, include_root = include_root_in_json)
+    hash = ActiveSupport::JSON.decode(json)
```
elohanlon: I usually use ruby's built-in `JSON.parse`.  Any additional advantage to using `ActiveSupport::JSON`?

db/migrate/20190530154456_add_doi_location_to_publish_targets.rb
```diff
@@ -0,0 +1,6 @@
+class AddDoiLocationToPublishTargets < ActiveRecord::Migration[5.2]
```
elohanlon: Could we rename this migration?  Doesn't seem to match what it's doing.

lib/hyacinth/adapters/external_identifier_adapter/datacite/ezid_response.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def body
+    @response.body
+  end
+
+  def parse_body
+    parsed_body_hash = {}
+    # First, split by '\n' to get individual lines
+    parsed_body = @response.body.split("\n")
+    # Then split each line in 2 along ':'. NOTE: using #map!
```
elohanlon: Okay to delete this comment and the line below it?

lib/hyacinth/adapters/external_identifier_adapter/memory.rb
```diff
@@ -0,0 +1,67 @@
...
+        # @return [Boolean] true if this adapter can handle this type of identifier
+        def handles?(_id)
+          true
+        end
+
+        # Generates a new persistent id, ensuring that nothing currently uses that identifier.
+        # @return [String] a new id
+        def mint
+          new_id = Random.rand.to_s
+          @identifiers[new_id] = {}
```
elohanlon: Since this method is supposed to ensure that nothing currently uses the minted identifier, should you be checking for the presence of new_id in `@identifiers` and looping until you don't find it?

spec/fixtures/files/datacite/datacite.xml
```diff
@@ -0,0 +1,45 @@
...
+    <relatedIdentifier relatedIdentifierType="ISSN" relationType="IsPartOf">1932-6203</relatedIdentifier>
+    <relatedIdentifier relatedIdentifierType="ISBN" relationType="IsPartOf">0670734608</relatedIdentifier>
+    <relatedIdentifier relatedIdentifierType="DOI" relationType="IsVariantFormOf">10.1371/journal.pone.0119638</relatedIdentifier>
+  </relatedIdentifiers>
+  <creators>
+    <creator>
+      <creatorName>Salinger, J. D.</creatorName>
+    </creator>
+    <creator>
+      <creatorName>Lincoln, Abraham</creatorName>
```
elohanlon: Thank you for continuing to use Abraham Lincoln in our fixtures. 🙂

app/models/concerns/digital_object_concerns/save_behavior.rb
barmintor: I'm not sure what to do about this. Send the UID with the DOI mint metadata? Reluctant to rely on sending a withdrawal since we'd already be in an error state. We could have a pending DOI and do the mint after the save?

app/models/publish_target.rb
barmintor: Nope, I'll remove it.

lib/hyacinth/adapters/external_identifier_adapter/datacite/ezid_response.rb
barmintor: Sure (these are ported)

lib/hyacinth/adapters/external_identifier_adapter/memory.rb
barmintor: We can, sure - tho (as the in-memory implementation for testing) I'm inclined to just trust the random number generator?

spec/fixtures/files/datacite/datacite.xml
barmintor: I wish I could take credit for it, but: ported over.

db/migrate/20190530154456_add_doi_location_to_publish_targets.rb
barmintor: From before the field rename, I'll rename it. How's "AddDoiAttributesToPublishTargets"?

db/schema.rb
```diff
@@ -162,6 +162,8 @@
     t.string "api_key", null: false
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.boolean "is_allowed_doi_target", default: false, null: false
```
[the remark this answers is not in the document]
barmintor: I'm inclined to punt, since we'd have to come up with a suitable noun to replace it with...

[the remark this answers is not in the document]
cgalarza: that's cool

lib/hyacinth/adapters/external_identifier_adapter/datacite.rb
```diff
@@ -0,0 +1,113 @@
...
+          # setup EZID API info: credentials, url, etc.
+          api_session = Datacite::EzidSession.new(DATACITE[:user], DATACITE[:password])
+          # ApiSession#modify_identifier returns true if the response from the EZID server indicated
+          # success, else it returns false
+          api_session.modify_identifier(doi, _target: target_url, _status: IDENTIFIER_STATUS[:public])
+        end
+
+        def tombstone_impl(id)
+          return false if id.nil?
+          api_session = Datacite::EzidSession.new(DATACITE[:user], DATACITE[:password])
```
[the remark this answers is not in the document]
barmintor: Will do.

lib/hyacinth/adapters/external_identifier_adapter/datacite/metadata.rb
```diff
@@ -0,0 +1,181 @@
...
+  def genre_uri
+    @dynamic_field_data['genre'][0]['genre_term']['uri'] if @dynamic_field_data.key?('genre') && @dynamic_field_data['genre'][0].key?('genre_term')
+  end
+
+  # the abstract of an item
+  # @api public
+  # @return [String, nil]
+  # @note only returns the first abstract value
+  def abstract
+    @dynamic_field_data['abstract'][0]['abstract_value'] if @dynamic_field_data.key? 'abstract'
```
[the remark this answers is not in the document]
barmintor: Good call, will fix this up.

app/models/concerns/digital_object_concerns/save_behavior.rb
barmintor: I'm opening a separate issue for this question.

elohanlon: Sounds good.  As long as we have an issue, I'm fine with not solving with this PR.

db/migrate/20190530154456_add_doi_location_to_publish_targets.rb
elohanlon: Sounds good!

lib/hyacinth/adapters/external_identifier_adapter/memory.rb
elohanlon: I know the odds are really low, but since it's pretty easy to do, do you mind if we add this?  It'll make me feel slightly better.

```
def mint
  loop do 
    new_id = Random.rand.to_s
    break unless @identifiers.key?(new_id)
  end

  @identifiers[new_id] = {}
  new_id
end
```

PR #122, opened 2019-08-01
More updates to React Front End; Fixing travis build

cgalarza:


cgalarza:
These test pass on my local machine, but won't pass here until our jettywrapper fix gets merged in.

PR #126, opened 2019-09-03
Digital Object Api Fixes

cgalarza:
- Fixing DigitalObject#create action
- Adding development publication adapter
- Fixing setter for projects
- Adding /preserve route for digital objects

### DIFF
--- a/app/controllers/api/v1/digital_objects_controller.rb
+++ b/app/controllers/api/v1/digital_objects_controller.rb
@@ -24,14 +24,14 @@ module Api
       # POST /digital_objects.json
       def create
         digital_object_data = create_or_update_params
-        digital_object_record = DigitalObjectRecord.create
-        @digital_object = DigitalObject::Base.from_serialized_form(digital_object_record, digital_object_data)
+        @digital_object = Hyacinth.config.digital_object_types.key_to_class(digital_object_data['digital_object_type']).new
+        @digital_object.set_digital_object_data(digital_object_data, true)
+
         @digital_object.projects.each do |project|
           authorize! :create_objects, project
         end
-        @digital_object.state ||= Hyacinth::DigitalObject::State::ACTIVE
-        # digital_object_record.save first because it will assign necessary digital_object_record attributes
-        if @digital_object.save(update_index: true) && (digital_object_record.persisted? || digital_object_record.save)
+
+        if @digital_object.save(update_index: true)
           show
         else
           render json: @digital_object.errors, status: :unprocessable_entity
--- a/app/models/concerns/digital_object_concerns/digital_object_data/setters/projects.rb
+++ b/app/models/concerns/digital_object_concerns/digital_object_data/setters/projects.rb
@@ -5,7 +5,7 @@ module DigitalObjectConcerns
 
       def set_projects(digital_object_data)
         return unless digital_object_data.key?('projects')
-        self.projects = Set.new(digital_object_data['projects'].each { |digital_object_data_project|
+        self.projects = Set.new(digital_object_data['projects'].map { |digital_object_data_project|
           Project.find_by(string_key: digital_object_data_project['string_key'])
         })
       end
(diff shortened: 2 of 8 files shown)


app/controllers/api/v1/digital_objects_controller.rb
```diff
@@ -24,14 +24,14 @@ def show
       # POST /digital_objects.json
       def create
         digital_object_data = create_or_update_params
-        digital_object_record = DigitalObjectRecord.create
-        @digital_object = DigitalObject::Base.from_serialized_form(digital_object_record, digital_object_data)
+        @digital_object = Hyacinth.config.digital_object_types.key_to_class(digital_object_data['digital_object_type']).new
+        @digital_object.set_digital_object_data(digital_object_data, true)
```
barmintor: if we start thinking of a DigitalObject instance as, essentially, a decorated deeply nested hash - is the set_digital_object_data method essentially merge or replace? Could we name the methods like the corresponding Hash methods?

app/models/concerns/digital_object_concerns/digital_object_data/setters/projects.rb
```diff
@@ -5,7 +5,7 @@ module Projects
 
       def set_projects(digital_object_data)
         return unless digital_object_data.key?('projects')
-        self.projects = Set.new(digital_object_data['projects'].each { |digital_object_data_project|
+        self.projects = Set.new(digital_object_data['projects'].map { |digital_object_data_project|
```
barmintor: this seems like an error you fixed, but the result will be that projects is an array instead of the last resolved key from the digital object data. that sounds right to me, and the property has a plural name, but I just want to flag in case we use the projects attribute as if it were singular.

app/controllers/api/v1/digital_objects_controller.rb
elohanlon: @barmintor It's sort of a merge/replace, but there's also some data transformation / augmentation involved for some fields.  Sometimes a string key is provided for a field and it's resolved to an object like a Project.  Other times, for a URI dynamic field, a minimal hash is provided for a field value and that hash will have other key/value pairs added to it (pref_label, authority).  There are also some params expected by set_digital_object_data that don't map 1:1 with DigitalObject field names (e.g. publish_to, unpublish_from), but they do end up setting other fields (e.g. publish_target_entries).  Does that change the way you feel about it?

app/models/concerns/digital_object_concerns/digital_object_data/setters/projects.rb
cgalarza: I think we are always expecting a Set and we treat the field like an array (for the most part).

app/controllers/api/v1/digital_objects_controller.rb
cgalarza: `merge` does seem a lot clearer to me even though it is doing some data transformation. 

I know this is probably only a Carla problem, but when I read `set_digital_object_data` I always have to do a double take and make sure it isn't `set_dynamic_field_data`. `merge` implies that it accepts and merges all the data associated with a digital object.

cgalarza: I want to merge this pull request. What should we do about this? Should we make a ticket to change the name of this method, or just leave as is?

elohanlon: I'm open to other method names if you want something shorter, but the other case I'll make against `merge` is based on the method signature:

`def set_digital_object_data(new_digital_object_data, merge_dynamic_fields)`

If we called it `merge`, then we'd be optionally passing `merge` a second "but don't merge everything" arg.  Seems weird.

elohanlon: @cgalarza I'd say no harm in merging this PR, and we can break this discussion out into a separate conversation.  The change isn't related to the other changes in your PR.

cgalarza: Okay, let's do that!

PR #145, opened 2019-11-20
Using graphql to query for a singular digital object

cgalarza:
This is just the beginnings of the graphql implementation for digital object.

### DIFF
--- a/app/graphql/types/digital_object/asset_type.rb
+++ b/app/graphql/types/digital_object/asset_type.rb
@@ -1,0 +1,7 @@
+module Types
+  class DigitalObject::AssetType < Types::BaseObject
+    implements Types::DigitalObjectInterface
+
+    # ... additional fields
+  end
+end
--- a/app/graphql/types/digital_object/asset_type.rb
+++ b/app/graphql/types/digital_object/asset_type.rb
@@ -1,7 +1,9 @@
 module Types
-  class DigitalObject::AssetType < Types::BaseObject
-    implements Types::DigitalObjectInterface
+  module DigitalObject
+    class AssetType < Types::BaseObject
+      implements Types::DigitalObjectInterface
 
-    # ... additional fields
+      # ... additional fields
+    end
   end
 end
--- a/app/models/digital_object/base.rb
+++ b/app/models/digital_object/base.rb
@@ -77,5 +77,9 @@ module DigitalObject
     def project_ids
       projects.each.map(&:id)
     end
+
+    def parents
+      parent_uids.map { |i| DigitalObject::Base.find(i) }
+    end
   end
 end
--- a/app/models/digital_object/base.rb
+++ b/app/models/digital_object/base.rb
@@ -79,7 +79,7 @@ module DigitalObject
     end
 
     def parents
-      parent_uids.map { |i| DigitalObject::Base.find(i) }
+      @parents ||= parent_uids.map { |i| DigitalObject::Base.find(i) }
     end
   end
 end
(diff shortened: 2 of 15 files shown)


app/graphql/types/digital_object/asset_type.rb
```diff
@@ -0,0 +1,9 @@
+module Types
+  module DigitalObject
+    class AssetType < Types::BaseObject
+      implements Types::DigitalObjectInterface
```
elohanlon: @cgalarza Is it possible for the various digital object graphql types to inherit from a common digital object type that implements `Types::DigitalObjectInterface`, so you don't need to repeat `implements Types::DigitalObjectInterface` in each class?

app/models/digital_object/base.rb
```diff
@@ -77,5 +77,9 @@ def initialize
     def project_ids
       projects.each.map(&:id)
     end
+
+    def parents
+      parent_uids.map { |i| DigitalObject::Base.find(i) }
```
elohanlon: Maybe good to cache this?
```
@parents ||= parent_uids.map { |i| DigitalObject::Base.find(i) }
```

Or actually, to avoid stale data and also to use cache when possible, maybe something like this:
```
if @parents
  @parents = @parents.map do |p|
    if p.updated_at != DigitalObject.get_updated_at(p.uid)
        DigitalObject::Base.find(p.uid)
    else
        p
    end
  end
else
  @parents ||= parent_uids.map { |i| DigitalObject::Base.find(i) }
end
```
Note that `DigitalObject.get_updated_at` doesn't exist yet, but could query solr.

spec/factories/digital_objects/items.rb
```diff
@@ -1,11 +1,26 @@
...
-#
-#       instance
-#     end
-#   end
-# end
+FactoryBot.define do
+  factory :item, class: DigitalObject::Item do
+    initialize_with do
+      instance = new
+      instance.instance_variable_set('@uid', Random.rand.to_s)
```
elohanlon: I'd recommend changing `Random.rand.to_s` to `SecureRandom.uuid` to match our usual uid format.

app/models/digital_object/base.rb
cgalarza: I was on the fence of adding an instance variable because I wasn't sure about how it would affect the other instance variables that are created. I can definitely add your first suggestion tho.

spec/factories/digital_objects/items.rb
cgalarza: Sure. I'll have to fix it in the other fixture. In theory, we wouldn't need to set the uid regardless right?

app/graphql/types/digital_object/asset_type.rb
cgalarza: Potentially?

I think it balances out because either way you have to specify which class to inherit/implement. Creating an intermediate type doesn't really make it any clearer, it just adds another level of indirection.

elohanlon: I just suggested it because it's slightly DRYer since the class needs to inherit from something anyway.  If you'd prefer not to though, I'm okay with it as is.

app/models/digital_object/base.rb
elohanlon: Great, thanks!

spec/factories/digital_objects/items.rb
elohanlon: Oh yeah, you're right.  In that case, should we remove the whole `instance_variable_set ` line?

app/graphql/types/digital_object/asset_type.rb
cgalarza: 👍

spec/factories/digital_objects/items.rb
cgalarza: Yea, i think so. I removed it from this fixture and the other digital object fixture.

PR #146, opened 2019-11-20
adds seed data for descriptive metadata category, title group and subfields

JackBlackLight:
This is minimal seed data to allow for the creation of digital objects.


cgalarza:
@elohanlon, would you want this to be a rake task? I kind of like it as part of seed file. I actually think we should move the rake task that creates users to this file, that way its all in one place and one command loads the DB with everything we need.

db/seeds.rb
```diff
@@ -1,7 +1,13 @@
...
-#   Character.create(name: 'Luke', movie: movies.first)
+DynamicFieldCategory.destroy_all
+DynamicFieldGroup.destroy_all
+DynamicField.destroy_all
+
+user = User.find(1);
+
+DynamicFieldCategory.create!({id: 1, display_label: "Descriptive Metadata", sort_order: 1})
+
+df_group = DynamicFieldGroup.create!({id: 1, string_key: 'title', display_label: 'Title', sort_order: 0, is_repeatable: 0, parent_id: 1, created_by: user, updated_by: user, parent_type: "DynamicFieldCategory", xml_translation: '{\n    \"element\": \"mods:titleInfo\",\n    \"content\": [\n        {\n            \"element\": \"mods:nonSort\",\n            \"content\": \"{{title_non_sort_portion}}\"\n        },\n        {\n            \"element\": \"mods:title\",\n            \"content\": \"{{title_sort_portion}}\"\n        }\n    ]\n}'})
```
cgalarza: I would stray away from setting the identifiers as part of this process because if the developer has already created some dynamic fields, groups or categories with the same ids they would get errors. I think you should be able to set the category for the DynamicFieldGroup by using the `parent` field and passing in the DynamicFieldCategory created above, that way you don't have to reference the identifiers.

db/seeds.rb
```diff
@@ -1,7 +1,13 @@
...
-#
-#   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
-#   Character.create(name: 'Luke', movie: movies.first)
+DynamicFieldCategory.destroy_all
+DynamicFieldGroup.destroy_all
+DynamicField.destroy_all
+
+user = User.find(1);
+
+DynamicFieldCategory.create!({id: 1, display_label: "Descriptive Metadata", sort_order: 1})
```
cgalarza: I think the curly braces here are redundant. Though, I am curious as to why rubocop did not catch that.

db/seeds.rb
elohanlon: Agreed!  We can rely on database auto-increment id for this.

elohanlon:
> @elohanlon, would you want this to be a rake task? I kind of like it as part of seed file. I actually think we should move the rake task that creates users to this file, that way its all in one place and one command loads the DB with everything we need.

Either a rake task or we can have the seeds file call a bunch of rake tasks.  (I know we said we'd talk more about that.)  Either way, we don't want this seeds file to get really really long.

Most important, I'd say we want to write these seeds/tasks in a way that's idempotent.  It should be possible to add new seeds later on without blowing away existing database data.  General pattern:

```
if [this set of things hasn't been set up already]
  set the things up
end

if [this other set of things hasn't been set up already]
  set the other things up
end
```

JackBlackLight:
Hi, my latest commit moves the data creation code into a rake task and addresses the other issues we've talked about.

lib/tasks/hyacinth/setup.rake
```diff
@@ -56,5 +37,107 @@ namespace :hyacinth do
...
+      end
+      df_category = DynamicFieldCategory.find_by(display_label: "Descriptive Metadata")
+      if df_category
+        puts Rainbow("dynamic field category 'Descriptive Metadata' already exists (skipping).").blue.bright
+      else
+        df_category = DynamicFieldCategory.create!(display_label: "Descriptive Metadata")
+        puts Rainbow("Created dynamic field category 'Descriptive Metadata'").green
+      end
+
+      df_group = DynamicFieldGroup.find_by(display_label: "Title")
```
cgalarza: For DynamicFieldGroup and DynamicField lookups should be done by `:string_key`. In this case, for those two models `string_keys` have to be unique.

lib/tasks/hyacinth/setup.rake
```diff
@@ -56,5 +37,107 @@ namespace :hyacinth do
...
+          select_options: '{}',
+          additional_data_json: '{}'          
+        }
+      ]
+
+      default_dynamic_fields.each do |field|
+        if DynamicField.exists?(string_key: field[:string_key])
+          puts Rainbow("dynamic field #{field[:display_label]} already exists (skipping).").blue.bright
+        else
+          DynamicField.create!(
```
cgalarza: Here, you should be able to do `DynamicField.create!(**field)` unless there's some reason to be selective about what variables to use. You may not even need the two asterisk.

JackBlackLight:
@cgalarza, this commit should resolve those two points you mentioned.

cgalarza:
There's rubocop issues with the merge because we updated the rubocop configuration since you started this PR. @JackBlackLight, if you don't mind I can rebase and fix the issues.

JackBlackLight:
@cgalarza , sure! Sounds great.

PR #147, opened 2019-11-20
No longer using :index, :show, :edit and :new to define permissions

cgalarza:
We shouldn't use :index, :show, :edit and :new to define permissions. These actions are mapped to one of :read, :update, :create and are really only used for RESTful responses. Once we move everything to graphql we shouldn't need to use them anymore.


cgalarza:
read, update, create, delete and any other custom permissions (read_objects, publish, etc)

PR #151, opened 2019-12-05
Hyacinth 391

barmintor:


barmintor:
@cgalarza you're right about the validation, adding it now.

PR #152, opened 2019-12-05
Moving URI Service back into Hyacinth

cgalarza:
- Adding term and vocabulary models and tests
- Adding the ability to have more than one solr core
- Adding solr configuration for Terms solr core
- Adding adapter for Terms solr core

Note: This does not include the GraphQL endpoint, that will come in a later PR


app/models/term.rb
```diff
@@ -0,0 +1,150 @@
...
+  def self.local_uri_prefix
+    host = HYACINTH[:local_uri_prefix]
+
+    raise 'Missing local_uri_prefix in config/hyacinth.yml' unless host
+    host.ends_with?('/') ? host : "#{host}/"
+  end
+
+  private
+
+    def cast_custom_fields
```
elohanlon: In the future, I wonder if it might be worth using something like dry_schema to validate and cast custom field types:
https://dry-rb.org/gems/dry-schema/1.4/

app/models/term.rb
```diff
@@ -0,0 +1,150 @@
...
+
+    def set_uri_hash
+      self.uri_hash = Digest::SHA256.hexdigest(uri) if uri
+    end
+
+    # Check that uid, uri and term_type were not changed.
+    def uid_uri_and_term_type_unchanged
+      return unless persisted? # skip if object is new or is deleted
+
+      errors.add(:uid, 'Change of uid not allowed!') if uid_changed?
```
elohanlon: Compared to the old URI service gem, it's nice extending ApplicationRecord and getting the *_changed? methods automatically.

app/models/term.rb
cgalarza: Yea, I would love to not have all this custom code to do the validation and casting. I don't know if I would want to delay this PR with that change. Maybe it can be a Hyacinth 3.1 addition.

barmintor: Or roll it into typedefs

cgalarza: I don't feel like I have a very deep understand of TypeDefs. Could they be used outside of the digital object model?

cgalarza: I'll write up a ticket about this, and we can merge this PR

lib/solr/params.rb
```diff
@@ -0,0 +1,41 @@
...
+        q: nil,
+        qt: 'search',
+        fq: [],
+        # rows: URIService::DEFAULT_LIMIT,
+        start: 0
+      }
+    end
+
+    def fq(field, value)
+      @parameters[:fq] << "#{field}:\"#{Solr::Utils.escape(value)}\"" unless value.nil? ## should probably escape
```
elohanlon: @cgalarza Is this an old comment that can be removed because you're using `Solr::Utils.escape` already?  Or is is indicating something that I'm missing?

cgalarza: It can be deleted, i'll remove it in my next PR

app/models/term.rb
elohanlon: TypeDefs might work, but dry-schema could offer a simpler implementation since this isn't deeply nested data.  I'm open to considering both though.

elohanlon: If we did want to open up TypeDef use to non-DigitalObjects, we'd want to move it out of the DigitalObject namespace.

PR #153, opened 2019-12-05
Allow array value comparison for 'equal' and 'not_equal' xml translation rules

elohanlon:


barmintor:
Do I understand correctly that HUG is requesting that equal_to(val) act like a reversed version of `include?`

elohanlon:
@barmintor `equal` acts like a regular version of `include` (does array contain) and `not_equal` is like a reversed version (does array NOT contain).

does value equal [ANY OF THESE]
and
does value not_equal [ANY OF THESE]

barmintor:
I have some mixed feelings. Overloading `equal_to` this way (obviously) works, but creates a knowledge/documentation need. My instinct is that new comparison functions would be preferable (any_of/not_any_of), but that would be a lot more work (not just the addition, but type checking the value_to_compare_to.

barmintor:
The work looks otherwise perfectly good, so my only question is whether you worry that hiding two behaviors in one function is going to generate bugs down the line. I don't think value_for_field_name will ever return an array, but if it did this would become a complicated thing to explain.

elohanlon:
@barmintor You make good points.  I just pushed an updated commit with `equals_any_of` and `equals_none_of` instead.  What do you think?

I've also applied these changes to Hyacinth 3, but haven't pushed yet.

elohanlon:
Thanks for the helpful feedback!

PR #154, opened 2019-12-06
split projects into primary_project and other_projects (HYACINTH-390)

barmintor:
also validates appropriate is_primary state of the various projects


app/models/concerns/digital_object_concerns/digital_object_data/setters/projects.rb
```diff
@@ -6,8 +6,18 @@ module Projects
...
+      end
+
+      def set_primary_project(digital_object_data)
+        return unless digital_object_data.key?('primary_project')
+        self.primary_project = Project.find_by(string_key: digital_object_data['primary_project']['string_key'])
+      end
+
+      def set_other_projects(digital_object_data)
+        return unless digital_object_data.key?('other_projects')
+        self.other_projects = Set.new(digital_object_data['other_projects'].map { |digital_object_data_project|
```
elohanlon: In the case of a project string key typo, this would result in nil elements being saved in the other_projects array (unless there's another protection elsewhere in the code that I'm not thinking of / haven't seen yet).  We should raise an exception or have a validation for this so we can somehow inform the user of their error.  I think that an exception might be easier to handle, since a validation could only tell the user that one of their `other_projects` is nil, and not which original invalid string value resulted in a conversion to nil.

elohanlon: In the case of a project string key typo, this would result in nil elements being saved in the other_projects array (unless there's another protection elsewhere in the code that I'm not thinking of / haven't seen yet). We should raise an exception or have a validation for this so we can somehow inform the user of their error. I think that an exception might be easier to handle, since a validation could only tell the user that one of their other_projects is nil, and not which original invalid string value resulted in a conversion to nil.

barmintor: yes, the validation would be too late unless we actually stored a jsonified "bad project" entry. I'll look into an exception.

app/models/concerns/digital_object_concerns/digital_object_data/setters/projects.rb
```diff
@@ -6,8 +6,18 @@ module Projects
...
       def set_projects(digital_object_data)
-        return unless digital_object_data.key?('projects')
-        self.projects = Set.new(digital_object_data['projects'].map { |digital_object_data_project|
+        set_primary_project(digital_object_data)
+        set_other_projects(digital_object_data)
+      end
+
+      def set_primary_project(digital_object_data)
+        return unless digital_object_data.key?('primary_project')
+        self.primary_project = Project.find_by(string_key: digital_object_data['primary_project']['string_key'])
```
elohanlon: Since we agreed to raise an exception for set_other_projects when an invalid project string key is supplied, I think we should do the same here.

app/validators/projects_validator.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class ProjectsValidator < ActiveModel::Validator
+  def validate(digital_object)
+    digital_object.errors[:projects] << "Only primary projects can be assigned to primary_project" if digital_object.primary_project && !digital_object.primary_project.is_primary
```
elohanlon: Raising an exception in the primary_project/other_projects setter code will prevent unintentional `nil.is_primary` calls here, so that's good.

app/validators/projects_validator.rb
```diff
@@ -0,0 +1,9 @@
...
+    return unless digital_object.other_projects.detect(&:is_primary)
```
elohanlon: Raising an exception in the other_projects setter code will prevent unintentional `nil.is_primary` calls here, so that's good.

lib/hyacinth/digital_object/type_def/json_serializable_array.rb
```diff
@@ -3,7 +3,19 @@
 module Hyacinth
   module DigitalObject
     module TypeDef
-      class JsonSerializableArray < Hyacinth::DigitalObject::TypeDef::JsonSerializableBase
+      class JsonSerializableArray < Hyacinth::DigitalObject::TypeDef::JsonSerializableCollection
+        def from_serialized_form_impl(json_var)
+          return nil if json_var.nil?
```
elohanlon:
```suggestion
          return nil if json_var.nil?
          raise ArgumentError, "Expected array, but got: #{json_var.class}" unless json_var.is_a?(Array)
```

lib/hyacinth/digital_object/type_def/json_serializable_collection.rb
```diff
@@ -0,0 +1,15 @@
...
+
+module Hyacinth
+  module DigitalObject
+    module TypeDef
+      class JsonSerializableCollection < Hyacinth::DigitalObject::TypeDef::Base
+        def initialize(type = nil)
+          super()
+          @translator = type.new if type
+          raise NotImplementedError, "Cannot instantiate #{self.class}. Instantiate a subclass instead." if self.class == JsonSerializableCollection
+        end
```
elohanlon: If you add these two methods...
```
def to_serialized_form_impl(value)
          value # no conversion necessary for JsonSerializableBase objects.  JSON#generate will handle them properly.
        end

        def from_serialized_form_impl(value)
          value # no conversion necessary for JsonSerializableBase objects.  JSON#parse will generate correct form.
        end
```

Can this class just be named to `JsonSerializableBase`, and replace the existing `JsonSerializableBase` class?

lib/hyacinth/digital_object/type_def/json_serializable_collection.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module Hyacinth
+  module DigitalObject
+    module TypeDef
+      class JsonSerializableCollection < Hyacinth::DigitalObject::TypeDef::Base
+        def initialize(type = nil)
+          super()
```
elohanlon:
```suggestion
          super
```

elohanlon:
@barmintor I really like what you've done with the JsonSerializableCollection class, making translation recursive!

lib/hyacinth/digital_object/type_def/json_serializable_collection.rb
barmintor: This would cause an error because the super method has no arguments defined, and the bar call to super passes all the arguments the method received.

lib/hyacinth/digital_object/type_def/json_serializable_collection.rb
barmintor: I'd like to think about that in the same scope as removing the Projects and PublishEntries typedefs - I'm unsure about it right now.

lib/hyacinth/digital_object/type_def/json_serializable_array.rb
barmintor: I think this is a good change but there are similar oversights in one of the other classes so I'm going to push a separate commit.

lib/hyacinth/digital_object/type_def/json_serializable_collection.rb
elohanlon: Ohhhh, right.

lib/hyacinth/digital_object/type_def/json_serializable_array.rb
elohanlon: 👍

PR #157, opened 2019-12-09
HYACINTH-408: Refactored adapters to simplify usage and improve compatibility with development environment constant reloading

elohanlon:


config/templates/hyacinth.template.yml
```diff
@@ -2,60 +2,60 @@ development:
   local_uri_prefix: localhost:3000 # prefix to use when minting local URIs
   metadata_storage:
     adapters:
-      - type: disk
+      - type: Hyacinth::Adapters::StorageAdapter::Disk
```
cgalarza: Having the class name directly in this configuration file seems odd to me. It seems like the config file knows a little too much about the structure of the application. 

Could it be possible to derive the class name from the type? All the adapters start with `Hyacinth::Adapters`. That's an easy thing to add to every class name that comes through. For all the top level keys that end in `storage`, we know we have to look in `Hyacinth::Adapters::StorageAdapters`. So for 'disk', we know it has to start with `Hyacinth::Adapters` and then because its metadata storage it has to be in the `StorageAdapter` namespace. Then we camelize `disk` and append it to the class name.

Alternatively, when the configuration for Storage Adapters is passed to the Storage Adapters Abstract class, it could pass the namespace it expects all of its adapters to be within to `create_from_config`.

elohanlon: Good suggestion.  I like that it would result in a much more concise config file too.  I went with your second suggestion about `create_from_config`.  Will push the commit as soon as I finish testing.

If you don't mind, I'll just use the already-camelized names (i.e. DatabaseEntryLock) in the config because that allows us to look at the class name for reference and not think about the (admittedly easy) translation.  Find and replace will also be easier if we change names later.  Going snake case to camelized case is 1:1, so it doesn't buy much, and it avoids having us translate it for the config just to internally translate it back.

Thanks!

PR #158, opened 2019-12-12
Hyacinth 387

barmintor:
REST query to fetch only projects appropriate to object creation


app/controllers/api/v1/projects_controller.rb
```diff
@@ -9,6 +9,7 @@ class ProjectsController < ApplicationApiController
 
       # GET /projects
       def index
+        scope_for_digital_object_creation if params[:can_create]
```
cgalarza: Should we call this `is_primary` since we are just filtering for primary projects and not the projects that a person can add objects to?

barmintor: I thought we wanted something more abstract than that, but if we're ok with going with `is_primary` it will make the graphql refactor easier later on.

cgalarza: I thought `can_create` would only return the projects for which a person can create objects, but if that's not the case I am okay with going with `is_primary` or `primary: true` or something.

elohanlon: I'm in favor of going with `is_primary` too, just so the client-side caller is clear about what we're asking for here -- so as to avoid suggesting that we intend to get only create-access projects.  Plus, when all of this goes away in favor of graphql, we can do it the real way and consider both primary and other projects that the user has create permission for.

PR #166, opened 2019-12-16
Rights data serialization

elohanlon:
@cgalarza It would be great to get your thoughts on this when you have a chance.  I wanted to make sure that I implemented things in a way that aligns with your other digitalObjectInterface-related views, and that my fixes for react warnings didn't work against any of your preferences.  Thanks!

### DIFF
--- a/app/javascript/hyacinth_ui_v1/components/digital_objects/digitalObjectInterface.js
+++ b/app/javascript/hyacinth_ui_v1/components/digital_objects/digitalObjectInterface.js
@@ -53,15 +53,17 @@ const digitalObjectInterface = WrappedComponent => class extends React.Component
                 <Tab to={`/digital_objects/${id}/rights`} name="Rights" />
 
                 {
-                  (data.digitalObjectType === 'item') && (
+                  (data.digitalObjectType === 'item') ? (
                     <Tab to={`/digital_objects/${id}/children`} name="Manage Child Assets" />
                   )
+                    : <></>
                 }
 
                 {
-                  (data.digitalObjectType === 'asset') && (
+                  (data.digitalObjectType === 'asset') ? (
                     <Tab to={`/digital_objects/${id}/parents`} name="Parents" />
                   )
+                    : <></>
                 }
 
                 <Tab to={`/digital_objects/${id}/assignment`} name="Assign This" />
--- a/app/javascript/hyacinth_ui_v1/components/digital_objects/metadata/MetadataShow.js
+++ b/app/javascript/hyacinth_ui_v1/components/digital_objects/metadata/MetadataShow.js
@@ -25,11 +25,11 @@ class MetadataShow extends React.PureComponent {
       });
   }
 
-  renderField(dynamicField, data, elementKey) {
+  renderField(dynamicField, data) {
     const { displayLabel, fieldType } = dynamicField;
 
     return (
-      <InputGroup key={elementKey}>
+      <InputGroup key={dynamicField.stringKey}>
         <Label align="right">{displayLabel}</Label>
         <PlainText value={fieldType === 'controlled_term' ? data.prefLabel : data } />
       </InputGroup>
@@ -55,7 +55,7 @@ class MetadataShow extends React.PureComponent {
                   if (c.type === 'DynamicFieldGroup') {
                     return this.renderGroup(c, d[camelCase(c.stringKey)]);
                   } if (c.type === 'DynamicField') {
-                    return this.renderField(c, d[camelCase(c.stringKey)], `${c.stringKey}_${i + 1}`);
+                    return this.renderField(c, d[camelCase(c.stringKey)]);
                   }
                 }
                 return '';
--- a/app/javascript/hyacinth_ui_v1/components/digital_objects/metadata/MetadataShow.js
+++ b/app/javascript/hyacinth_ui_v1/components/digital_objects/metadata/MetadataShow.js
@@ -25,11 +25,11 @@ class MetadataShow extends React.PureComponent {
       });
   }
 
-  renderField(dynamicField, data) {
+  renderField(dynamicField, data, elementKey) {
     const { displayLabel, fieldType } = dynamicField;
 
     return (
-      <InputGroup>
+      <InputGroup key={elementKey}>
         <Label align="right">{displayLabel}</Label>
         <PlainText value={fieldType === 'controlled_term' ? data.prefLabel : data } />
       </InputGroup>
@@ -50,12 +50,12 @@ class MetadataShow extends React.PureComponent {
           </Card.Header>
           <Card.Body>
             {
-              children.map((c) => {
+              children.map((c, i) => {
                 if (d[camelCase(c.stringKey)]) {
                   if (c.type === 'DynamicFieldGroup') {
                     return this.renderGroup(c, d[camelCase(c.stringKey)]);
                   } if (c.type === 'DynamicField') {
-                    return this.renderField(c, d[camelCase(c.stringKey)]);
+                    return this.renderField(c, d[camelCase(c.stringKey)], `${c.stringKey}_${i + 1}`);
                   }
                 }
                 return '';
--- a/app/javascript/hyacinth_ui_v1/components/digital_objects/rights/ItemRightsForm.js
+++ b/app/javascript/hyacinth_ui_v1/components/digital_objects/rights/ItemRightsForm.js
@@ -21,111 +21,26 @@ import RightsForWorksOfArtSculptureAndPhotographs from './subsections/RightsForW
 import UnderlyingRights from './subsections/UnderlyingRights';
 
 class ItemRightsForm extends React.Component {
-  state = {
-    rights: {
-      descriptiveMetadata: {
-        typeOfContent: '',
-        countryOfOrigin: {},
-        filmDistributedToPublic: '',
-        filmDistributedCommercially: '',
-      },
-      copyrightStatus: {
-        copyrightStatement: '',
-        copyrightNote: [''],
-        copyrightRegistered: '',
-        copyrightRenewed: '',
-        copyrightDateOfRenewal: '',
-        copyrightExpirationDate: '',
-        culCopyrightAssessmentDate: '',
-      },
-      additionalRightsToRecord: {
-        enabled: false,
-      },
-      copyrightOwnership: {
-        enabled: false,
-        copyrightOwners: [
-          {
-            name: {},
-            heirs: '',
-            contactInformation: '',
-          },
-        ],
-      },
-      columbiaUniversityIsCopyrightHolder: {
-        enabled: false,
-        dateOfTransfer: '',
-        dateOfExpiration: '',
-        transferDocumentionEnabled: false,
-        transferDocumentation: '',
-        otherTransferEvidenceEnabled: false,
-        otherTransferEvidence: '',
-        transferDocumentationNote: '',
-      },
-      licensedToColumbiaUniversity: {
-        enabled: false,
-        dateOfLicense: '',
-        terminationDateOfLicense: '',
-        credits: '',
-        acknowledgements: '',
-        licenseDocumentationLocation: '',
-      },
-      contractualLimitationsRestrictionsAndPermissions: {
-        a: false,
-        b: false,
-        c: false,
-        d: false,
-        e: false,
-        avA: false,
-        avB: false,
-        avC: false,
-        avD: false,
-        avE: false,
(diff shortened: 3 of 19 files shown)


app/javascript/hyacinth_ui_v1/components/digital_objects/digitalObjectInterface.js
```diff
@@ -53,15 +53,17 @@ const digitalObjectInterface = WrappedComponent => class extends React.Component
                 <Tab to={`/digital_objects/${id}/rights`} name="Rights" />
 
                 {
-                  (data.digitalObjectType === 'item') && (
+                  (data.digitalObjectType === 'item') ? (
                     <Tab to={`/digital_objects/${id}/children`} name="Manage Child Assets" />
                   )
+                    : <></>
```
barmintor: Are the empty elements important? I'm assuming the `to` attribute means we don't care about indexes of the tabs?

app/models/concerns/digital_object_concerns/digital_object_data/setters/rights.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module DigitalObjectConcerns
+  module DigitalObjectData::Setters
+    module Rights
+      extend ActiveSupport::Concern
+
+      def set_rights(digital_object_data)
+        return unless digital_object_data.key?('rights')
+        # TODO: We need to add some kind of validation here, or potentially
```
barmintor: can we tag HYACINTH-429 in this TODO?

elohanlon: Done!  Good suggestion.

app/javascript/hyacinth_ui_v1/components/digital_objects/digitalObjectInterface.js
elohanlon: The empty elements were the only way I found to get rid of a JS warning I was seeing.  When they weren't here before, the <Tabs> parent kept complaining about receiving non-element values for one or more of its children.

And yes, I'm pretty sure the tab index doesn't matter here, but I'd feel slightly better if @cgalarza confirms because she's used it more than I have.

cgalarza: I think we should go back to how I originally implemented this bit. 

The warnings are going to appear because of my propTypes validation. The root issue of that warning is how I am defining the types and not how this is specifically implemented. The propTypes defined in the `Tabs` component need to be able to handle a use case like this. I use these tab components in other places so it might come up again.

app/javascript/hyacinth_ui_v1/components/digital_objects/metadata/MetadataShow.js
```diff
@@ -25,11 +25,11 @@ class MetadataShow extends React.PureComponent {
       });
   }
 
-  renderField(dynamicField, data) {
+  renderField(dynamicField, data, elementKey) {
```
cgalarza: Is this to remove warning around not having a unique key for each dynamic field?

app/javascript/hyacinth_ui_v1/components/digital_objects/rights/ItemRightsForm.js
```diff
@@ -136,10 +51,111 @@ class ItemRightsForm extends React.Component {
     }));
   }
 
-  onSubmitHandler() {
-    // To be implemented.
+  onSubmitHandler = () => {
+    const { data: { uid } } = this.props;
```
cgalarza: Lines 55 & 56 can be combined into one, if you'd like.

app/javascript/hyacinth_ui_v1/components/digital_objects/rights/ItemRightsForm.js
```diff
@@ -155,7 +171,7 @@ class ItemRightsForm extends React.Component {
       },
     } = this.state;
 
-    const { id } = this.state;
+    const { data: { uid } } = this.props;
```
cgalarza: I know I go back and forth between using `uid` and `id` in the frontend. But I think we should really move towards using `id.` It's a pattern i've also been trying to follow with the graphql endpoints. No need to change it here since this will have to be converted to a functional component in the near future.

app/javascript/hyacinth_ui_v1/components/ui/forms/buttons/SubmitButton.js
```diff
@@ -21,7 +21,7 @@ class SubmitButton extends React.Component {
   }
 
   render() {
-    const { formType, saveData, ...rest } = this.props;
+    const { formType, saveData, staticContext, ...rest } = this.props;
```
cgalarza: Why did you add staticContext here?

spec/models/digital_object/base_spec.rb
```diff
@@ -97,6 +98,7 @@
             pending_unpublish_from: [],
             structured_children: { 'type' => 'sequence', 'structure' => [] },
             dynamic_field_data: {},
+            rights: {},
```
cgalarza: Should we add a test to make sure a populated json get stored? or do we want to wait to implement rights validation before we do that?

app/javascript/hyacinth_ui_v1/components/digital_objects/digitalObjectInterface.js
elohanlon: I'm finding these warnings in the js console to be distracting while I'm doing js development, and they're thrown on the **error** level so they look like errors while I'm developing.  Would it be okay if we ticket the changes you described and when they're in place, we can remove this temporary fix?  Or if you want to push a commit onto this branch, feel free, and we can remove this now.

app/javascript/hyacinth_ui_v1/components/digital_objects/metadata/MetadataShow.js
elohanlon: Right!

app/javascript/hyacinth_ui_v1/components/digital_objects/rights/ItemRightsForm.js
elohanlon: Good suggestion.  Thanks!

app/javascript/hyacinth_ui_v1/components/digital_objects/rights/ItemRightsForm.js
elohanlon: Isn't it easier to go with `uid` because it's the actual name of the property on the digital object, so it'll play better with the common destructuring assignment use case?

app/javascript/hyacinth_ui_v1/components/ui/forms/buttons/SubmitButton.js
elohanlon: If I don't, it gets pulled into `...rest`, and then gets passed along to another component and creates a warning about unexpected property `staticContext`.

app/javascript/hyacinth_ui_v1/components/digital_objects/digitalObjectInterface.js
cgalarza: I would rather ticket or remove the propType check entirely.

app/javascript/hyacinth_ui_v1/components/digital_objects/rights/ItemRightsForm.js
cgalarza: The way I imagine it, the graphql endpoint would not return a `uid`, it would just return an `id`. While we are hitting the RESTful endpoint `uid` makes sense, but once we move to the graphql endpoint we should start using `id`.

app/javascript/hyacinth_ui_v1/components/digital_objects/digitalObjectInterface.js
elohanlon: Sounds good!  Are you okay with creating the ticket because you best understand the changes you want us to make?

app/javascript/hyacinth_ui_v1/components/digital_objects/rights/ItemRightsForm.js
elohanlon: Maybe we should change:
`metadata_attribute :uid, Hyacinth::DigitalObject::TypeDef::String.new`
to:
`metadata_attribute :id, Hyacinth::DigitalObject::TypeDef::String.new` ?

app/javascript/hyacinth_ui_v1/components/digital_objects/metadata/MetadataShow.js
cgalarza: I think I have forgotten how I implemented this, but is it possible to use something from the `dynamicField` hash?

spec/models/digital_object/base_spec.rb
elohanlon: I want to rethink digital object validation overall, so I'm good with waiting until we add rights validation.  Right now it's sort of covered by the general tests for any generically defined `metadata_attribute`.

app/javascript/hyacinth_ui_v1/components/ui/forms/buttons/SubmitButton.js
cgalarza: I have seen this error and I haven't had time to figure out a better solution. This works for now, but it will continue to come up, maybe we should ticket this? or at least put a comment right above?

app/javascript/hyacinth_ui_v1/components/digital_objects/rights/ItemRightsForm.js
cgalarza: I would 100% approve of that. Is it possible to do without colliding with DigitalObjectRecord?

spec/models/digital_object/base_spec.rb
cgalarza: 👍

app/javascript/hyacinth_ui_v1/components/digital_objects/rights/ItemRightsForm.js
elohanlon: I think it should be fine as long as we don't delegate DigitalObject::Base's :id to DigitalObjectRecord.  Will be good to do in the near future.

app/javascript/hyacinth_ui_v1/components/ui/forms/buttons/SubmitButton.js
elohanlon: I don't know if it's necessary to ticket or comment because I think this is just a general coding pattern that we'll need to watch out for when using `...rest`, as it grabs a whole bunch of things that we don't always want to pass along to various sub-components.  We should just be careful about only passing down exactly what we need to various child components.  It'll prevent warnings and will also make PropType validation more useful, so we can trust that the documented prop types are the only ones actually being received.

Mentioning @barmintor @fcd1 @JackBlackLight here so they're all aware of this too.

cgalarza: I though this problem only exists when we use the `withRouter` HOC component. When we move to functional components, we will use hooks to pull `history` and `params` from react-router and will no longer have a need for `withRouter` and I am hoping that removes this problem. That's why I think we should comment, or else we might continue to deconstruct a prop we don't actually have to deconstruct.

elohanlon: Oh okay.  So you're concerned specifically about the staticContext prop?  I can add a comment about that.  I just meant that use of `...rest` in a similar way anywhere in the app is something to watch out for, if passing unnecessary props.

app/javascript/hyacinth_ui_v1/components/digital_objects/digitalObjectInterface.js
cgalarza: I can create it.

app/javascript/hyacinth_ui_v1/components/digital_objects/metadata/MetadataShow.js
elohanlon: Good idea.  I'll use `dynamicField. stringKey` instead.

app/javascript/hyacinth_ui_v1/components/ui/forms/buttons/SubmitButton.js
elohanlon: Just added the comment we talked about.

PR #168, opened 2019-12-17
Only display createdAt and updatedBy uid if available

cgalarza:


app/javascript/hyacinth_ui_v1/components/digital_objects/system_data/SystemData.js
```diff
@@ -58,13 +58,13 @@ class SystemData extends React.Component {
       <>
         <Row as="dl">
           <Col as="dt" lg={2} sm={4}>Created By</Col>
-          <Col as="dd" lg={10} sm={8}>{createdBy.uid || '-- Assigned After Save --'}</Col>
+          <Col as="dd" lg={10} sm={8}>{(createdBy && createdBy.uid) || '-- Not Assigned --'}</Col>
```
elohanlon: @cgalarza This looks so much like a ternary operator situation that I feel like it might be worth swapping to that.  I don't feel too strongly about it though, so if you'd prefer to keep as is I'm fine with that.

cgalarza: whoops, will do that

elohanlon: Thanks!

PR #169, opened 2019-12-18
Hyacinth 351

barmintor:
Enforce distinct project permission for rights updates via REST API

### DIFF
--- a/app/controllers/api/v1/digital_objects/rights_controller.rb
+++ b/app/controllers/api/v1/digital_objects/rights_controller.rb
@@ -1,0 +1,51 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    module DigitalObjects
+      class RightsController < ApplicationApiController
+        before_action :ensure_json_request
+        before_action :load_resource, only: [:show, :edit, :update]
+
+        # GET /digital_objects/1/rights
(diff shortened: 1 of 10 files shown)


app/controllers/api/v1/digital_objects/rights_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+          authorize! :update_rights, @digital_object
+        end
+
+        # PATCH/PUT /digital_objects/1/rights
+        # PATCH/PUT /digital_objects/1/rights.json
+        def update
+          authorize! :update_rights, @digital_object
+          digital_object_rights_data = update_params
+          if digital_object_rights_data[:rights]
+            @digital_object.rights = digital_object_rights_data[:rights]
```
elohanlon: We have a set_digital_object_data method that allows us to set various properties on the digital object (including rights), but there have been a couple of times now where we instead enabled a public writer to a DigitalObject::Base metadata_attribute.  We should probably talk offline about when we want to use that high-level method vs. enable direct value assignment.

elohanlon: Update: What @barmintor is doing here is great, and we'll switch over to explicit private_writer for fields that shouldn't be called from outside the DigitalObject class.

PR #174, opened 2019-12-20
Hyacinth 402

barmintor:


app/javascript/hyacinth_ui_v1/components/users/UserEdit.js
```diff
@@ -14,35 +13,25 @@ import InputGroup from '../ui/forms/InputGroup';
...
-const UPDATE_USER = gql`
-  mutation UpdateUser($input: UpdateUserInput!){
-    updateUser(input: $input){
-      user {
-        id
-      }
-    }
-  }
-`;
+  let gqlResponse = useQuery(
```
cgalarza: The authenticated user's abilities are actually available to all components. They are set in the `<App />` component via the AbilityContext. Retrieving the abilities for the logged in user here shouldn't be necessary. 

Abilities can be checked with the `Can` component. I believe it can also be retrieved via a hook. 
```
import { Can } from '../../../util/ability_context';
{ ........... }
<Can I="update" of={{ subjectType: 'Project', stringKey: data.project.stringKey }}>
{ ...... }
</Can>
```

More documentation here: https://github.com/stalniy/casl/tree/master/packages/casl-react

app/javascript/hyacinth_ui_v1/components/users/UserEdit.js
```diff
@@ -97,29 +93,33 @@ function UserEdit() {
...
+        email,
+        currentPassword,
+        password,
+        passwordConfirmation,
       },
+    };
+    if (ability.can('manage', 'User')) {
+      if (!isAdmin) userData.input.isActive = isActive;
+    }
+    if (authenticatedUser.isAdmin) {
```
cgalarza: This can be converted to `ability.can('manage', 'all')`.

app/javascript/hyacinth_ui_v1/components/users/UserEdit.js
```diff
@@ -236,6 +237,7 @@ function UserEdit() {
                     sm={12}
                     value={isAdmin}
                     label="Administrator"
+                    disabled={!authenticatedUser.isAdmin}
```
cgalarza: This can also be converted to `ability.can('manage', 'all')`.

app/javascript/hyacinth_ui_v1/components/users/UserEdit.js
```diff
@@ -59,34 +48,41 @@ function UserEdit() {
   const [manageUsers, setManageUsers] = useState(false);
   const [readAllDigitalObjects, setReadAllDigitalObjects] = useState(false);
   const [manageAllDigitalObjects, setManageAllDigitalObjects] = useState(false);
+  const [canActivate, setCanActivate] = useState(false);
 
-  const { loading, error } = useQuery(
-    GET_USER,
+  const canActivateUser = (subject, subjectAbility, object) => {
```
cgalarza: Instead of passing the AuthenticatedUser abilities can be checked using `abilities.can?(action, subject)`.

app/javascript/hyacinth_ui_v1/components/users/UserEdit.js
barmintor: hmm, I didn't see that in the permissions I inspected - it was `('manage', 'Users')`, which is definitely a mistake. Is all what we want?

PR #179, opened 2019-12-23
Cleaning up ControlledVocabularySelect

cgalarza:


cgalarza:
@barmintor created a specific PrimaryVocabularySelect for the digital object interface and I am wondering if we should keep all digital object specific inputs under `/digital_objects/inputs` or something like that.

elohanlon:
@cgalarza Moving all digital object specific inputs under `components/digital_objects/inputs` sounds good to me.  👍 

If we're doing some restructuring anyway, maybe we should also rename the `components/ui` directory to `components/shared`, since everything in `components` is related to the UI.

cgalarza:
@elohanlon I'll move the inputs around as a separate PR. I strayed away from naming the folder `shared` because I didn't want to get it confused with other shared classes that aren't particularly UI components. For example, if we have a class that holds a set of methods we find useful, etc.

PR #183, opened 2020-01-02
HYACINTH-389 and HYACINTH-441

barmintor:
* Organize validator classes into modules where appropriate
* Add a switch to limit dynamic field enabled to primary project (HYACINTH-389)
* Add cancel button to bottom of Project Enabled Dynamic Fields form (HYACINTH-441)

### DIFF
--- a/app/javascript/hyacinth_ui_v1/components/projects/enabled_dynamic_fields/EnabledDynamicFieldForm.js
+++ b/app/javascript/hyacinth_ui_v1/components/projects/enabled_dynamic_fields/EnabledDynamicFieldForm.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import {
   Row, Col, Form, Badge, Card,
 } from 'react-bootstrap';
@@ -6,8 +7,7 @@ import produce from 'immer';
 import { withRouter } from 'react-router-dom';
 import axios from 'axios';
 import Select from 'react-select';
-
-import SubmitButton from '../../layout/forms/SubmitButton';
+import FormButtons from '../../ui/forms/FormButtons';
 import hyacinthApi from '../../../util/hyacinth_api';
 import withErrorHandler from '../../../hoc/withErrorHandler/withErrorHandler';
 
@@ -20,7 +20,7 @@ class EnabledDynamicFieldEdit extends React.Component {
   }
 
   componentDidMount() {
-    axios.all([this.getDynamicFieldGraph(), this.getEnabledDynamicFields()])
+    axios.all([EnabledDynamicFieldEdit.getDynamicFieldGraph(), this.getEnabledDynamicFields()])
       .then(axios.spread((graph, enabledDynamicFields) => {
         const allFieldsHash = {};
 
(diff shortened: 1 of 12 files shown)


app/javascript/hyacinth_ui_v1/components/projects/enabled_dynamic_fields/EnabledDynamicFieldForm.js
```diff
@@ -20,7 +20,7 @@ class EnabledDynamicFieldEdit extends React.Component {
   }
 
   componentDidMount() {
-    axios.all([this.getDynamicFieldGraph(), this.getEnabledDynamicFields()])
+    axios.all([EnabledDynamicFieldEdit.getDynamicFieldGraph(), this.getEnabledDynamicFields()])
```
cgalarza: Why not use `this`?

cgalarza: Oh, is it because it's a static method?

barmintor: Yes - eslint rules pushing me in different directions; I picked static.

app/validators/enabled_dynamic_field/open_to_all_projects_validator.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class EnabledDynamicField::OpenToAllProjectsValidator < ActiveModel::Validator
+  def validate(enabled_dynamic_field)
+    return if enabled_dynamic_field.open_to_all_projects || enabled_dynamic_field.project&.is_primary
+    enabled_dynamic_field.errors[:open_to_all_projects] << "Only primary projects have exclusively enabled dynamic fields"
```
elohanlon: @barmintor Maybe worth changing this error message to align more with the field name?  Something like "Option 'Open to all projects' can only be enabled for fields in primary projects."  Might help with troubleshooting if we encounter this validation error in the future.

app/javascript/hyacinth_ui_v1/components/projects/enabled_dynamic_fields/EnabledDynamicFieldForm.js
```diff
@@ -273,10 +289,15 @@ class EnabledDynamicFieldEdit extends React.Component {
   }
 
   render() {
-    const { dynamicFieldGraphs, disabled } = this.state;
+    const {
+      dynamicFieldGraphs, disabled,
```
elohanlon: The `disabled` state value seems to be derived from the `formType` props value, which never changes as far as I can tell.  I know you didn't add this, so this is more of a suggestion for the future.

Could we either just use formType, or pass in disabled as a prop?  In the PermissionsEditor component, I pass in a `readOnly` prop that has the same effect as `disabled`, and I'd propose that as a good naming alternative.  

Generally though, it would be simpler to check a boolean rather than a string value (show vs. edit), assuming we only ever plan to have two modes.

PR #186, opened 2020-01-08
Vocabulary and Terms GraphQL Endpoint

cgalarza:
- Adding Create, Update and Delete Mutations for Vocabularies and Terms
- Added queries for vocabularies, vocabulary, term and terms
- Created a results wrapper type that can be used to group results and provide a place to add additional data about the result set
- Added test for all queries and mutations


app/graphql/types/scalar/any_primative_type.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
```
barmintor: can we rename this AnyPrimitiveType after the merge?

spec/requests/graphql/mutations/term/update_term_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+
+      it 'returns error in json' do
+        expect(response.body).to be_json_eql(%(
+          "Vocabulary is locked."
+        )).at_path('errors/0/message')
+      end
+    end
+  end
+
+  def query
```
barmintor: Starting to wonder if we should move all these gql literals into a spec helper/support file in a future ticket.

cgalarza: I actually like the queries rights in the testing spec for them. We usually never use these queries in two different specs.

app/graphql/types/scalar/any_primative_type.rb
cgalarza: I will do that right now!

spec/requests/graphql/mutations/term/update_term_spec.rb
barmintor: Definitely useful as long as that's true! I thought there were some repeated in this PR, but I might have misread.

app/graphql/types/scalar/any_primative_type.rb
barmintor: Thanks!

app/graphql/mutations/term/base_mutation.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class Mutations::Term::BaseMutation < Mutations::BaseMutation
+  def find_vocabulary!(string_key)
```
elohanlon: Can we rename this to something like `find_unlocked_vocabulary!` ?  Given that the current name looks like a plain object finder method, I feel like it's a little unintuitive that it would raise an error if the vocabulary is locked.  Name change would help.

app/graphql/mutations/term/create_term.rb
```diff
@@ -0,0 +1,34 @@
...
+class Mutations::Term::CreateTerm < Mutations::Term::BaseMutation
+  argument :vocabulary_string_key, ID, required: true
+  argument :pref_label, String, required: true
+  argument :alt_labels, [String], required: false
+  argument :authority, String, required: false
+  argument :uri, String, required: false
+  argument :term_type, Types::TermCategory, required: false # enum local, temporary, external
+  argument :custom_fields, [Types::CustomFieldAttributes], required: false
+
+  field :term, Types::TermType, null: true
```
elohanlon: When would this field return null?

app/graphql/mutations/term/delete_term.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class Mutations::Term::DeleteTerm < Mutations::Term::BaseMutation
+  argument :vocabulary_string_key, ID, required: true
+  argument :uri, String, required: true
+
+  field :term, Types::TermType, null: true
```
elohanlon: Same question here about when `term` would return null.

app/graphql/mutations/term/update_term.rb
```diff
@@ -0,0 +1,30 @@
...
+
+class Mutations::Term::UpdateTerm < Mutations::Term::BaseMutation
+  argument :vocabulary_string_key, ID, required: true
+  argument :uri, String, required: true
+  argument :pref_label, String, required: false
+  argument :alt_labels, [String], required: false
+  argument :authority, String, required: false
+  argument :custom_fields, [Types::CustomFieldAttributes], required: false
+
+  field :term, Types::TermType, null: true
```
elohanlon: Same question here about when `term` would return null.

app/graphql/mutations/vocabulary/create_vocabulary.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class Mutations::Vocabulary::CreateVocabulary < Mutations::BaseMutation
+  argument :string_key, ID, required: true
+  argument :label, String, required: true
+  argument :locked, Boolean, required: false
+
+  field :vocabulary, Types::VocabularyType, null: true
```
elohanlon: Same question here about when `vocabulary` would return null.  Since I've already posted a few comments about this, there may be something obvious that I'm not seeing, haha.  I won't post anymore comments about this in your other graphql classes.

app/graphql/mutations/term/create_term.rb
cgalarza: When there is an error with the request or if the object isn't found.

app/graphql/types/base_object.rb
```diff
@@ -2,5 +2,30 @@
 
 module Types
   class BaseObject < GraphQL::Schema::Object
+    def self.results_type(&block)
+      wrapped_type = self
+
+      GraphQL::ObjectType.define do
+        type_name = wrapped_type.is_a?(GraphQL::BaseType) ? wrapped_type.name : wrapped_type.graphql_name
+
+        name("#{type_name}Results")
```
elohanlon: Cool addition!  I was looking forward to seeing this implementation and I like it.

app/graphql/mutations/term/create_term.rb
elohanlon: I just tested things out in a mutation in my local copy to confirm, and it looks like any error raised will short circuit the rest of the resolve logic and you won't actually end up returning that field at all (null or otherwise).

Here's an example with the updateUser mutation, where I'm sending a user id that isn't valid:

```
mutation {
  updateUser(input: {id:"does-not-exist"}){
    user {
      id
    }
  }
}
```

The response I get is:

```
{
  "data": {
    "updateUser": null
  },
  "errors": [
    {
      "message": "Couldn't find User",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "updateUser"
      ]
    }
  ]
}
```

I can do `field :user, Types::UserType, null: true` or `null: false` in that mutation and it doesn't make a difference, so that's why I thought that setting `null: true` for a field was a way of saying "there are cases where I *won't* raise an error that graphql will catch for the entire request (and put in the top level errors array), but I still want this particular field to return null".  Do you see what I mean?  Maybe there's something here that I'm not getting, but I wanted to make sure we're on the same page.  I didn't think you'd ever see `{ createTerm: { user: null } }` when an error is raised.  I thought you'd get `{createTerm: null}` instead.

app/graphql/types/custom_field_definition_type.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Types
+  class CustomFieldDefinitionType < Types::BaseObject
+    description 'A custom field definition'
+
+    field :field_key, ID, null: false
+    field :label, String, null: false
+    field :data_type, String, null: false # Enum string, integer, boolean
```
elohanlon: Doesn't have to be for this PR, but is a GraphQL enum type useful for data_type?

app/graphql/types/extensions/paginate.rb
```diff
@@ -0,0 +1,36 @@
...
+
+      def resolve(object:, arguments:, context:)
+        next_args = arguments.dup
+        next_args.delete(:limit)
+        next_args.delete(:offset)
+        yield(object, arguments)
+      end
+
+      def after_resolve(object:, arguments:, value:, context:, memo:)
+        raise GraphQL::ExecutionError, 'Paginate extension can only be used on ActiveRecord:Relation objects' unless value.is_a?(ActiveRecord::Relation)
```
elohanlon: If we want to use this later with Digital Object search result sets, they won't be ActiveRecord::Relation objects.  We can cross that bridge when we get to it though.  One alternative might be to just make sure that the `value` responds to the expected methods.  Or maybe we'll end up handling pagination separately for that.

elohanlon: Ah, I see Types::Extensions:: SolrSearch below, and how it's easy enough to introduce another FieldExtension later on, so I'm not worried about Digital Object results -- which could be Rsolr search result sets, but may actually be an aggregation of actual DigitalObject::Base subclass objects instantiated from the solr docs.

app/graphql/types/query_type.rb
```diff
@@ -50,6 +58,17 @@ def digital_object(id:)
       digital_object
     end
 
+    def vocabulary(string_key:)
+      vocabulary = Vocabulary.find_by!(string_key: string_key)
+      ability.authorize!(:read, vocabulary)
+      vocabulary
+    end
+
+    def vocabularies(_arguments)
```
elohanlon: Why can't this unused argument be removed?  Something to do with the pagination extension?

app/graphql/types/scalar/any_primitive_type.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+class Types::Scalar::AnyPrimitiveType < Types::BaseScalar
```
elohanlon: Creative solution for custom fields.  👍

spec/requests/graphql/mutations/term/create_term_spec.rb
```diff
@@ -0,0 +1,253 @@
...
+
+      let(:expected_json) do
+        %({
+            "term": {
+              "prefLabel": "Hippogriff",
+              "altLabels": [],
+              "authority": null,
+              "termType": "temporary",
+              "customFields": [
+                { "field": "classification", "value": null },
```
elohanlon: It might be a good idea to ensure/test that customFields always come back the same order, for consistency.  Alphabetical order, based on field name.  If you agree, we can ticket that as future work.

spec/requests/graphql/mutations/term/create_term_spec.rb
```diff
@@ -0,0 +1,253 @@
...
+            ]
+          }
+        }
+      end
+
+      before { graphql query, variables }
+
+      it 'returns error in json' do
+        expect(response.body).to be_json_eql(%(
+          "Uri can't be blank; Uri hash can't be blank"
```
elohanlon: "Uri hash can't be blank" might be confusing for users, since they can't edit that value and aren't even aware that it exists.  Maybe we can update the model to remove the redundant validation error.  If you agree, I can ticket.

spec/requests/graphql/mutations/term/update_term_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+    end
+
+    context 'when updating term_type' do
+      let(:term) { FactoryBot.create(:external_term, vocabulary: vocabulary) }
+      let(:variables) do
+        {
+          input: {
+            vocabularyStringKey: vocabulary.string_key,
+            uri: term.uri,
+            termType: 'local'
```
elohanlon: Would the mutation be rejected altogether because termType is present here, and it isn't a registered argument in the UpdateTerm mutation?

spec/requests/graphql/mutations/vocabulary/create_vocabulary_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+        expect(Vocabulary.count).to be 1
+        expect(Vocabulary.first.string_key).to eql 'collections'
+      end
+
+      it 'returns newly created vocabulary in json' do
+        expect(response.body).to be_json_eql(expected_response).at_path('data/createVocabulary')
+      end
+    end
+
+    context 'when string_key is missing' do
```
elohanlon:
```suggestion
    context 'when stringKey is missing' do
```

spec/requests/graphql/query/term_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+            }
+          }
+        )
+      end
+      before do
+        vocabulary.update(locked: true)
+        graphql query(vocabulary.string_key, term.uri)
+      end
+
+      it 'returns expected json response' do
```
elohanlon:
```suggestion
      it 'still returns the term, despite the vocabulary being locked' do
```
I thought this might be more descriptive for anyone viewing this test at a later time.  It took me a minute to realize what was being tested here.

spec/requests/graphql/mutations/term/create_term_spec.rb
cgalarza: They are already returned in alphabetical order based on field name. TermType sorts them.

spec/requests/graphql/mutations/term/create_term_spec.rb
cgalarza: I understand that its confusing but I don't think its a good idea to remove the validation from the model. That's the last check the model does before saving the record and if that check doesn't catch it then we will just be returning a database error.

spec/requests/graphql/mutations/term/update_term_spec.rb
cgalarza: Yea, that sounds right. But what I am trying to make sure is that termType isn't overridden, what prevents that from happening isn't as important. Though this is a test that I converted from the REST endpoint so maybe its not necessary anymore?

app/graphql/mutations/term/base_mutation.rb
cgalarza: I'll make that change.

app/graphql/mutations/term/create_term.rb
cgalarza: I see what you're saying. I based my implementation on the mutation examples on the graphql-ruby site and they handle errors differently then we do for mutations. So you're right, the null parameter should be removed it. I almost think this should be a ticket because this isn't the only place where I did that.

app/graphql/types/custom_field_definition_type.rb
cgalarza: @barmintor Did you create a ticket for creating enums?

barmintor: I did HYACINTH-456

app/graphql/mutations/term/create_term.rb
barmintor: I am 👍 to making the null param removals another ticket.

app/graphql/mutations/term/base_mutation.rb
elohanlon: Thanks!

app/graphql/mutations/term/create_term.rb
elohanlon: Ticket sounds good!  We can definitely do post-merge.

spec/requests/graphql/mutations/term/create_term_spec.rb
elohanlon: Oh good!  Never mind then!  Is it worth writing a test to ensure they're sorted, or no?  I'm fine with skipping if you don't think so.

app/graphql/types/custom_field_definition_type.rb
cgalarza: great, resolving this convo

spec/requests/graphql/mutations/term/create_term_spec.rb
elohanlon: I didn't mean that we should remove the validation, just that it might be worth removing the redundant message when those two messages appear together, since if uri is blank, uri hash definitely won't be valid.  It's almost like saying "stop at the first error because that needs to be resolved before the second error can be properly evaluated".

spec/requests/graphql/mutations/term/create_term_spec.rb
cgalarza: I think it's okay to skip, the tests are going to start failing if we don't sort anyways.

spec/requests/graphql/mutations/term/create_term_spec.rb
elohanlon: And even if we do skip the validation in the specific case when uri is blank, the validation is still essential as a protection if the code ever changes, paired with the db unique index you mentioned.

app/graphql/mutations/term/create_term.rb
barmintor: HYACINTH-457

spec/requests/graphql/mutations/term/create_term_spec.rb
cgalarza: ahh I see. We usually want to return multiple messages to the front end, for example, if you missed adding your first and last name that returns two errors. We generate these messages (validation errors) the same way for all models. I guess I am not sure how you envision implementing the change.

app/graphql/types/query_type.rb
cgalarza: Yea, in the #resolve function for the extension i have to pass  the arguments to the yield method. The vocabularies method, in this case, must take those results. I don't think there's a way around it.

elohanlon: Oh okay.  Makes sense!

spec/requests/graphql/mutations/term/create_term_spec.rb
elohanlon: In term.rb, can we add `if: Proc.new { |t| t.uri.present? }` to the end of the `validates :uri_hash` validation?

spec/requests/graphql/mutations/term/update_term_spec.rb
elohanlon: Yeah, I think it might not be necessary now.  In a way, this test is now just testing that any invalid argument doesn't end up modifying the term type.  I double checked and we have a `'cannot change term_type'` test in term_spec.rb, so I'm comfortable with removing this test.

spec/requests/graphql/mutations/term/create_term_spec.rb
cgalarza: Okay, feel free to ticket that work.

spec/requests/graphql/mutations/term/update_term_spec.rb
cgalarza: Okay, I will remove it.

spec/requests/graphql/mutations/term/create_term_spec.rb
elohanlon: Ticket created.

PR #187, opened 2020-01-08
Prohibiting Deactivated Users from Logging in

cgalarza:


barmintor:
whoops forgot: 👻

app/controllers/concerns/hyacinth/users/cas_authentication_behavior.rb
```diff
@@ -53,16 +53,20 @@ def identify_uni_user(user_uni, cas_logout_uri)
           if user_uni.present?
             possible_user = User.where(email: (user_uni + '@columbia.edu')).first
 
-            if possible_user.present?
-              sign_in possible_user, bypass: true
+            if possible_user.present? && possible_user.is_active
```
elohanlon: @cgalarza I think we might be able to reduce perceived complexity of this method with a little refactoring.  What do you think of this as a complete replacement for the method?

```
def identify_uni_user(user_uni, cas_logout_uri)
  if user_uni.blank?
    render inline: 'CAS Authentication failed, Please try again later.'
    return
  end

  possible_user = User.find_by(email: user_uni + '@columbia.edu')

  if possible_user.present?
    if possible_user.is_active
      bypass_sign_in possible_user
      cookies[:signed_in_using_uni] = true # Use this cookie to know when to do a CAS logout upon Devise logout
      flash[:notice] = 'You are now logged in.'
      redirect_to root_path, status: 302
      return
    end

    flash[:alert] = I18n.t 'devise.failure.inactive'
  else
    flash[:alert] = 'The UNI ' + user_uni + ' is not authorized to log into Hyacinth (no account exists with email ' + user_uni + '@columbia.edu).  If you believe that you should have access, please contact an application administrator.'
  end
  
  # Log out user
  redirect_to(cas_logout_uri + '?service=' + URI.encode_www_form_component(root_url))
end
```

cgalarza: I honestly think, we need to ticket either a) refactoring this entire concern or b) moving over to using a omniauth-cas or another gem.

elohanlon: Just a suggestion for readability, but do whatever you think is right.  I'll approve the PR.

elohanlon: btw, I'm also in favor of moving to the omniauth-cas gem.

cgalarza: I created a ticket: HYACINTH-472

elohanlon: 👍

PR #191, opened 2020-01-13
Allowing Tabs PropType to be more permissive for children

cgalarza:


app/javascript/hyacinth_ui_v1/components/ui/tabs/Tabs.js
```diff
@@ -2,20 +2,18 @@ import React from 'react';
...
+  return (
+    <Nav fill variant="tabs" className="mb-2">
+      {children}
+    </Nav>
+  );
 }
 
 Tabs.propTypes = {
-  children: PropTypes.arrayOf(PropTypes.element).isRequired,
+  children: PropTypes.arrayOf(PropTypes.node).isRequired,
```
elohanlon: @cgalarza I think you can just do `children: PropTypes.node.isRequired` here, since that worked for me when I created the new `DigitalObjectInterface` component.

cgalarza: Yea, I think that's right. Unless we want to enforce that there's always more than one element, but I don't think we have to do that. I'll push a change soon.

barmintor: Am I right in understanding this PR is: straightforward conversion to functional components of the 4 implicated components, and the swap of the `Tabs.propTypes.children` to `node` instead of `element`?

cgalarza: Yup, pretty much

barmintor: Ok, then given that you are already looking at the array change, I'm going to approve.

PR #192, opened 2020-01-14
HYACINTH-443 digital object frontend refactor

elohanlon:
Although HYACINTH-443 is the overall parent ticket, this also addresses HYACINTH-442 and HYACINTH-385.  We should squash this commit when we merge it, since the branch contains a lot of incomplete committed code that changed as we talked more about digital object view restructuring.


barmintor:
58 files! This is going to take a while.

app/javascript/hyacinth_ui_v1/components/digital_objects/children/Children.js
```diff
@@ -1,39 +1,37 @@
...
-      <>
-        <TabHeading>Manage Child Assets</TabHeading>
-        <p>Child Digital Objects</p>
-        <p># Total</p>
-
-
-      </>
-    );
-  }
+import { getChildrenDigitalObjectQuery } from '../../../graphql/digitalObjects';
```
barmintor: Ugh - I see why you had to do this component, but - I also did this in my branch. I'll have to do some rebasing if we merge this first.

elohanlon:
Yeah, this one is intense.  Thanks for taking a look!

app/javascript/hyacinth_ui_v1/components/digital_objects/rights/rights_form/ItemRightsForm.js
```diff
@@ -0,0 +1,222 @@
...
+import ContractualLimitationsRestrictionsAndPermissions from './subsections/ContractualLimitationsRestrictionsAndPermissions';
+import RightsForWorksOfArtSculptureAndPhotographs from './subsections/RightsForWorksOfArtSculptureAndPhotographs';
+import UnderlyingRights from './subsections/UnderlyingRights';
+
+function ItemRightsForm(props) {
+  const { digitalObject } = props;
+  const { id, rights: initialRights, dynamicFieldData } = digitalObject;
+  const history = useHistory();
+
+  const defaultRights = () => ({
```
barmintor: Can we move this template object function into its own file?

app/javascript/hyacinth_ui_v1/graphql/projects.js
```diff
@@ -42,13 +42,26 @@ export const deleteProjectMutation = gql`
   }
 `;
 
+const getProjectsFields = `
```
barmintor: Nice - I need to start doing this.

spec/requests/graphql/query/digital_object_spec.rb
```diff
@@ -2,8 +2,8 @@
 
 require 'rails_helper'
 
-RSpec.describe 'Retrieving Field Set', type: :request do
-  let(:authorized_object) { FactoryBot.create(:item, :with_project) }
+RSpec.describe 'Retrieving Digital Object', type: :request do
```
barmintor: Ha, I resisted the urge to fix this in my branch - thanks for fixing it here.

app/javascript/hyacinth_ui_v1/components/digital_objects/rights/rights_form/ItemRightsForm.js
elohanlon: Sure!  It was already there before my refactor, so I was going for minimal change where possible, but I like the idea of having it in a different file. Will push a commit for this soon.

spec/requests/graphql/query/digital_object_spec.rb
elohanlon: 👍

app/javascript/hyacinth_ui_v1/components/digital_objects/rights/rights_form/ItemRightsForm.js
elohanlon: Pushed!

spec/support/graphql_helper.rb
```diff
@@ -8,17 +8,30 @@ def graphql(query, variables = {})
...
-  def projects_query
-    <<~GQL
-      query {
-        projects {
-          stringKey
-          displayLabel
-          projectUrl
-          isPrimary
+  def projects_query(is_primary: nil)
+    if is_primary.nil?
```
cgalarza:
```suggestion
    is_primary_filter = is_primary ? "isPrimary: \"#{is_primary}\"" : ""
    <<~GQL
      query {
         projects(#{is_primary_filter}) {
         .............
```

cgalarza: This way you don't have to have the query written out twice. I think you can also use graphql variables.

app/javascript/hyacinth_ui_v1/components/digital_objects/new/DigitalObjectNewForm.js
```diff
@@ -0,0 +1,60 @@
...
+            if (can) {
+              return (
+                <>
+                <DigitalObjectSummary digitalObject={initialDigitalObject} />
+                <MetadataForm formType="new" digitalObject={initialDigitalObject} />
+                </>
+              );
+            }
+
+            return 'Invalid permissions for this view.';
```
cgalarza: Could we make this entire route protected? or is it important to display this message in this context?

app/javascript/hyacinth_ui_v1/components/digital_objects/DigitalObjectSummary.js
```diff
@@ -1,48 +1,60 @@
...
-    const {
-      data: {
-        projects,
-        pid,
-        uid,
-        doi,
+function DigitalObjectSummary(props) {
+
+  const { digitalObject } = props;
+  const { id, doi, primaryProject, otherProjects, numberOfChildren } = digitalObject;
```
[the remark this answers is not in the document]
elohanlon: Good suggestion! Thanks!

app/javascript/hyacinth_ui_v1/components/digital_objects/assignments/Assignments.js
```diff
@@ -0,0 +1,37 @@
...
+    <DigitalObjectInterface digitalObject={digitalObject}>
+      <TabHeading>Assignments</TabHeading>
+      <p>This feature is currently unavailable.</p>
+    </DigitalObjectInterface>
+  );
+}
+
+export default Assignments;
+
+Assignments.propTypes = {
```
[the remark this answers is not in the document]
elohanlon: I think it works either way, but it would look better to have before.  Will change, thanks!

app/javascript/hyacinth_ui_v1/components/digital_objects/metadata/Metadata.js
```diff
@@ -1,29 +1,58 @@
...
+  } = useQuery(getMetadataDigitalObjectQuery, {
+    variables: { id },
+  });
+
+  if (digitalObjectLoading) return (<></>);
+  if (digitalObjectError) return (<GraphQLErrors errors={digitalObjectError} />);
+  const { digitalObject } = digitalObjectData;
+
+  return (
+    <Switch>
```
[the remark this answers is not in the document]
elohanlon: Ah! I forgot to delete this file.  Good catch, thanks!

app/javascript/hyacinth_ui_v1/components/digital_objects/new/DigitalObjectNewForm.js
elohanlon: The reason I didn't do it at a higher level (in DigitalObjects.js) is because it made it possible to avoid yet another Digital Object retrieval request at that level.

You won't actually get to this particular route unless you manually navigate to it though.  Normally the only method is by submitting the project/digital_object_type select form in the DigitalObjectNew.  But if someone did somehow type in the URL manually, I added this <Can> check to hide the object building form if they don't have the correct project rights (since a request for the project is made in this component).

I think the alternative to this <Can> would be to add a <Switch>/<ProtectedRoute> filter to this component's render content, but I think it would do effectively the same thing?

cgalarza: I might be missing something, but I don't think you need to retrieve a digital object to check permissions for this route. You only need the project string key and that's passed through as a parameter.

app/javascript/hyacinth_ui_v1/components/digital_objects/rights/RightsTab.js
```diff
@@ -0,0 +1,40 @@
...
+function RightsTab(props) {
+  const { digitalObject, editButton, children } = props;
+
+  return (
+    <DigitalObjectInterface digitalObject={digitalObject}>
+      <TabHeading>
+        Rights
+        { editButton
+          && (
+            <EditButton
```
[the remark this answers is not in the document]
elohanlon: That check is done at a higher level.  See: `{ routePath: `${path}/rights/edit`, Component: RightsEdit, action: 'assess_rights' },` in DigitalObject.js.

The RightsEdit uses the RightsTab component internally, passing a flag for showing the edit button, so the checking doesn't occur in RightsTab.  This allows the RightsTab component to be more ignorant of context, as it's just told whether or not to show the edit button by the outer component.  Hopefully will make component testing easier later on too.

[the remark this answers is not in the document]
cgalarza: But does it show the button even if a user doesn't have the ability to edit the rights?

spec/support/graphql_helper.rb
elohanlon: Going with `projects(isPrimary: #{is_primary})` in the end (as we just talked about), since apparently passing null is fine here.  Thanks for suggesting the alternative!  Much better than what I had originally.

app/javascript/hyacinth_ui_v1/components/digital_objects/new/DigitalObjectNewForm.js
elohanlon: Ah!  You're right.   I was thinking about an earlier implementation of <DigitalObjectProtectedRoute>, but even in the latest version it only needs project string keys.  In that case, I could do that check on a higher level in DigitalObject.js.  Will do, thanks!

app/javascript/hyacinth_ui_v1/components/digital_objects/rights/RightsTab.js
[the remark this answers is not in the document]
elohanlon: Ohhhh, I see what you mean.  I should be able to fix that by passing a permission-based value into the editButton prop when using the component in the RightsShow component.  I also need to do a similar thing in the MetadataShow component.  I'm glad you brought this up!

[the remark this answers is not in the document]
elohanlon: Fixed in my local copy and I'll push soon.

elohanlon:
@cgalarza gave me verbal approval for this PR, so I'll merge it!

PR #193, opened 2020-01-15
Hyacinth 418

barmintor:
In this PR:
- install ActiveStorage
- use ActiveStorage uploads in concert with a new REST resource to create child Assets
- graphql query for child assets
- simple React UI for creating child assets
- simple React component for displaying a list of digital objects by id, title and type

Not in PR:
- reconciling multiple places to link/display child objects
- download of the master resource on the created Asset objects

### DIFF
--- a/app/controllers/api/v1/digital_objects/uploads_controller.rb
+++ b/app/controllers/api/v1/digital_objects/uploads_controller.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    module DigitalObjects
+      class UploadsController < ApplicationApiController
+        before_action :ensure_json_request
+        before_action :load_resource, only: [:create]
+
+        # POST /digital_objects/1/uploads
+        def create
+          authorize! :update, @digital_object
+          blob = ActiveStorage::Blob.create_before_direct_upload!(**create_params)
+
+          # this is a development hack to alow us to get the path despite ActiveStorage's protests
+          ActiveStorage::Current.host = 'stubhost'
+
+          render json: blob.as_json(root: false, methods: :signed_id).merge(direct_upload: {
+            url: blob.service_url_for_direct_upload.sub("http://#{ActiveStorage::Current.host}", ''),
+            headers: blob.service_headers_for_direct_upload
+          })
+        end
+
+        def load_resource
+          @digital_object ||= DigitalObject::Base.find(params[:id])
+        end
+
+        private
+
+          def create_params
+            # TODO: decide how we want to validate rights parameters
+            params.require(:blob).permit(:filename, :byte_size, :checksum, :content_type, :metadata).to_h.symbolize_keys
+          end
+      end
+    end
+  end
+end
--- a/app/controllers/api/v1/digital_objects/uploads_controller.rb
+++ b/app/controllers/api/v1/digital_objects/uploads_controller.rb
@@ -3,6 +3,10 @@
 module Api
   module V1
     module DigitalObjects
+      # Because Rails 5 ActiveStorage does not export the BlobRecord class,
+      # we must create the direct upload via REST. In Rails 6, the create logic
+      # below could be moved into a GraphQL mutation, and the returned metadata used
+      # to initiate the ActiveStorage upload.
       class UploadsController < ApplicationApiController
         before_action :ensure_json_request
         before_action :load_resource, only: [:create]
--- a/app/graphql/mutations/create_asset.rb
+++ b/app/graphql/mutations/create_asset.rb
@@ -1,0 +1,51 @@
+# frozen_string_literal: true
+
+class Mutations::CreateAsset < Mutations::BaseMutation
+  argument :signed_blob_id, String, "Signed blob ID generated via `createDirectUpload` mutation", required: true
+  argument :parent_id, ID, required: true
+
+  field :asset, Types::DigitalObject::AssetType, null: true
+
+  def resolve(signed_blob_id:, parent_id:)
+    parent = DigitalObject::Item.find(parent_id)
+    ability.authorize! :create_objects, parent.primary_project
+
+    blob = ActiveStorage::Blob.find_signed(signed_blob_id)
+
+    asset = initialize_child_asset(parent, blob.filename.to_s)
+    begin
+      resource = initialize_asset_resource_for_blob(asset, blob, Hyacinth::Config.resource_storage)
+      asset.save!
+      { asset: asset }
+    rescue StandardError => e
+      cleanup_on_failure(asset, resource)
+      raise e
+    ensure
+      blob&.destroy
+    end
+  end
+
+  def initialize_child_asset(parent, file_name)
+    asset = DigitalObject::Asset.new
+    asset.primary_project = parent.primary_project
+    asset.add_parent_uid(parent.uid)
+    asset.asset_type = BestType.dc_type.for_file_name(file_name)
+    asset
+  end
+
+  def initialize_asset_resource_for_blob(asset, blob, storage)
+    asset.with_primary_resource do |_resource_name, resource|
+      resource.original_filename = blob.filename.to_s
+      resource.media_type = BestType.mime_type.for_file_name(blob.filename.to_s)
+      resource.location = storage.generate_new_location_uri(SecureRandom.uuid, 'upload')
+      storage.with_writeable(resource.location) do |output_file|
+        blob.download { |chunk| output_file << chunk }
+      end
+    end
+  end
+
+  def cleanup_on_failure(asset, resource)
+    Hyacinth::Config.resource_storage.delete(resource.location) if resource&.location
+    asset.destroy unless asset.new_record?
+  end
+end
--- a/app/javascript/hyacinth_ui_v1/components/digital_objects/DigitalObjectList.js
+++ b/app/javascript/hyacinth_ui_v1/components/digital_objects/DigitalObjectList.js
@@ -1,0 +1,52 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Card } from 'react-bootstrap';
+import { LinkContainer } from 'react-router-bootstrap';
+import { has } from 'lodash';
+
+const titleForDigitalObject = (digObj) => {
+  let title = '[No Title]';
+  if (has(digObj, 'dynamicFieldData.title[0]')) {
+    const titleData = digObj.dynamicFieldData.title[0];
+    title = titleData.titleSortPortion;
+    if (titleData.titleNonSortPortion) {
+      title = `${titleData.titleNonSortPortion} ${title}`;
+    }
+  }
+  return title;
+};
+
+
+const DigitalObjectList = (props) => {
+  const { digitalObjects } = props;
+
+  return (
+    <>
+      {
+        digitalObjects.map(digitalObject => (
+          <Card key={digitalObject.id} className="mb-3">
+            <Card.Header>{titleForDigitalObject(digitalObject)}</Card.Header>
+            <Card.Body>
+              <LinkContainer to={`/digital_objects/${digitalObject.id}`}>
+                <Card.Link>{digitalObject.id}</Card.Link>
+              </LinkContainer>
+              <Card.Text>{digitalObject.digitalObjectType}</Card.Text>
+            </Card.Body>
+          </Card>
+        ))
+      }
+    </>
+  );
+};
+
+DigitalObjectList.propTypes = {
+  digitalObjects: PropTypes.arrayOf(
+    PropTypes.shape({
+      id: PropTypes.string.isRequired,
+      digitalObjectType: PropTypes.string.isRequired,
+      dynamicFieldData: PropTypes.object.isRequired,
+    }),
+  ).isRequired,
+};
+
+export default DigitalObjectList;
--- a/app/javascript/hyacinth_ui_v1/components/digital_objects/DigitalObjectList.js
+++ b/app/javascript/hyacinth_ui_v1/components/digital_objects/DigitalObjectList.js
@@ -2,20 +2,6 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Card } from 'react-bootstrap';
 import { LinkContainer } from 'react-router-bootstrap';
-import { has } from 'lodash';
-
-const titleForDigitalObject = (digObj) => {
-  let title = '[No Title]';
-  if (has(digObj, 'dynamicFieldData.title[0]')) {
-    const titleData = digObj.dynamicFieldData.title[0];
-    title = titleData.titleSortPortion;
-    if (titleData.titleNonSortPortion) {
-      title = `${titleData.titleNonSortPortion} ${title}`;
-    }
-  }
-  return title;
-};
-
 
 const DigitalObjectList = (props) => {
   const { digitalObjects } = props;
@@ -25,7 +11,7 @@ const DigitalObjectList = (props) => {
       {
         digitalObjects.map(digitalObject => (
           <Card key={digitalObject.id} className="mb-3">
-            <Card.Header>{titleForDigitalObject(digitalObject)}</Card.Header>
+            <Card.Header>{digitalObject.title}</Card.Header>
             <Card.Body>
               <LinkContainer to={`/digital_objects/${digitalObject.id}`}>
                 <Card.Link>{digitalObject.id}</Card.Link>
@@ -44,7 +30,7 @@ DigitalObjectList.propTypes = {
     PropTypes.shape({
       id: PropTypes.string.isRequired,
       digitalObjectType: PropTypes.string.isRequired,
-      dynamicFieldData: PropTypes.object.isRequired,
+      title: PropTypes.string.isRequired,
     }),
   ).isRequired,
 };
--- a/app/javascript/hyacinth_ui_v1/components/digital_objects/children/Children.js
+++ b/app/javascript/hyacinth_ui_v1/components/digital_objects/children/Children.js
@@ -1,34 +1,53 @@
-import React from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
+import { Card } from 'react-bootstrap';
 import { useQuery } from '@apollo/react-hooks';
 
 import DigitalObjectInterface from '../DigitalObjectInterface';
+import DigitalObjectList from '../DigitalObjectList';
+import AssetNew from '../new/AssetNew';
 import TabHeading from '../../ui/tabs/TabHeading';
-import { getChildrenDigitalObjectQuery } from '../../../graphql/digitalObjects';
+import { getChildStructureQuery } from '../../../graphql/digitalObjects';
 import GraphQLErrors from '../../ui/GraphQLErrors';
 
-function Children(props) {
+const Children = (props) => {
   const { id } = props;
-
+  const [inputId] = useState(AssetNew.randomInputId());
   const {
-    loading: digitalObjectLoading,
-    error: digitalObjectError,
-    data: digitalObjectData,
-  } = useQuery(getChildrenDigitalObjectQuery, {
+    loading: childStructureLoading,
+    error: childStructureError,
+    data: childStructureData,
+    refetch: refreshChildStructure,
+  } = useQuery(getChildStructureQuery, {
     variables: { id },
   });
 
-  if (digitalObjectLoading) return (<></>);
-  if (digitalObjectError) return (<GraphQLErrors errors={digitalObjectError} />);
-  const { digitalObject } = digitalObjectData;
-
+  if (childStructureLoading) return (<></>);
+  if (childStructureError) return (<GraphQLErrors errors={childStructureError} />);
+  const { childStructure: { parent, structure } } = childStructureData;
   return (
-    <DigitalObjectInterface digitalObject={digitalObject}>
+    <DigitalObjectInterface digitalObject={parent}>
       <TabHeading>Manage Child Assets</TabHeading>
-      <p>This feature is currently unavailable.</p>
+      <Card className="mb-3">
+        <Card.Body>
+          <Card.Title>
+            Add New Asset
+          </Card.Title>
+          <AssetNew parentId={id} refetch={refreshChildStructure} inputId={inputId} />
+        </Card.Body>
+      </Card>
+      <Card className="mb-3">
+        <Card.Body>
+          <Card.Title>
+            Child Digital Objects
+          </Card.Title>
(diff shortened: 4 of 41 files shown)


barmintor:
There are some conflicts between this PR and #192 that will require reconciliation.

app/graphql/mutations/create_asset.rb
```diff
@@ -0,0 +1,47 @@
...
+  def resolve(signed_blob_id:, parent_id:)
+    parent = DigitalObject::Item.find(parent_id)
+    ability.authorize! :create_objects, parent.primary_project
+
+    blob = ActiveStorage::Blob.find_signed(signed_blob_id)
+
+    asset = initialize_child_asset(parent, blob.filename.to_s)
+
+    begin
+      asset.resources['master'] = initialize_asset_resource_for_blob(asset, blob, Hyacinth::Config.resource_storage)
```
elohanlon: It might be good to use a constant here, and below, instead of literal string `'master'`.

barmintor: if we were to use a constant, does it make more sense for it to be at this layer, or on the model? If the latter, I would advocate taking up separately, maybe in a `.default_resource &block` type implementation that completely removes the reference from the mutation layer.

barmintor: "taking up separately" subsequent to this PR, that is

app/controllers/api/v1/digital_objects/uploads_controller.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    module DigitalObjects
+      # Because Rails 5 ActiveStorage does not export the BlobRecord class,
+      # we must create the direct upload via REST. In Rails 6, the create logic
+      # below could be moved into a GraphQL mutation, and the returned metadata used
+      # to initiate the ActiveStorage upload.
+      class UploadsController < ApplicationApiController
```
cgalarza: Uploading files right now is scoped to digital objects, when we convert this to be a graphql mutation will the upload action be applicable to the upload of all files? One important piece that this current request takes care of is checking permissions, which is checked a second time when the asset is created. Are the two checks necessary? If so, will there be multiple input mutations for each type of upload?

barmintor: Yeah, so this is a big question: We could have an abstracted file upload mutation that just checks valid user, and then clean up the direct upload if the action fails; or we could have two actions. Since I can't implement the independent file upload mutation until Rails 6, I felt like answering that question was out of scope. In the deleted stub code I actually had that very question in the comments.

app/javascript/hyacinth_ui_v1/components/digital_objects/new/AssetNew.js
```diff
@@ -0,0 +1,54 @@
...
+
+function debugObject(obj) {
+  console.log(obj);
+  return new Promise(resolve => resolve(obj));
+}
+
+function AssetNew(props) {
+  const { parentId, refetch } = props;
+  const [createAsset] = useMutation(createAssetMutation);
+  const elementId = `input-${String.toString(Math.random()).replace('.', '')}`;
```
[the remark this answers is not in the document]
barmintor: Sort of? But we don't have immediate plans for multiple file uploads. Do you recommend calculating it as a property of the component function itself? That seems like a pretty compartmentalized change.

app/javascript/hyacinth_ui_v1/components/digital_objects/new/AssetNew.js
```diff
@@ -0,0 +1,54 @@
...
+    const {
+      target: {
+        validity,
+        files: [file],
+      },
+    } = event;
+    if (!validity.valid) return false;
+    function resetFileInput() { document.getElementById(elementId).value = ''; }
+    const url = `/api/v1/digital_objects/${parentId}/uploads`;
+    return uploadFile(file, url)
```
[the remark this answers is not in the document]
barmintor: I got bound up here on the REST errors vs GraphQL errors thing - I think it would be much easier to implement after the Rails 6 change, but if that looks like it's a long way off I'll work something out.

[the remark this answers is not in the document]
cgalarza: That makes sense, we can definitely address this later.

app/controllers/api/v1/digital_objects/uploads_controller.rb
cgalarza: ha, okay, a discussion that can happen later

app/graphql/mutations/create_asset.rb
elohanlon: Definitely in the model, and subsequent PR sounds good!

app/graphql/mutations/create_asset.rb
```diff
@@ -0,0 +1,47 @@
...
+      asset.save!
+      parent.append_child_uid(asset.uid)
+      parent.save
+      { asset: asset }
+    ensure
```
elohanlon: Ignore this if I'm wrong, but it appears that if initialize_asset_resource_for_blob successfully copies the upload to our final preservation storage location, and then `asset.save!` fails, the newly copied file won't be cleaned up?

Note that deleting an asset will not destroy it's master resource file on disk, since we definitely wouldn't want to do that for files that "reference" their preservation storage location on import (e.g. files in a bag).

barmintor: That's true - but I'm also not sure how to fix it. Change `Asset.destroy` to clean up resources if the resource storage service has the location uri?

barmintor: Add a third thing to the `ensure` block?

elohanlon: Yeah, since this is JUST for asset creation via file upload, adding a third thing to the `ensure` block sounds good.

app/javascript/hyacinth_ui_v1/components/digital_objects/new/AssetNew.js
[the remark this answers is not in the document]
elohanlon: Reminder: We do currently support multiple file uploads in Hyacinth 2, so we'll want to support that when we get to our feature parity milestone.

app/javascript/hyacinth_ui_v1/components/digital_objects/children/Children.js
```diff
@@ -1,39 +1,61 @@
...
-    return (
-      <>
-        <TabHeading>Manage Child Assets</TabHeading>
-        <p>Child Digital Objects</p>
-        <p># Total</p>
-
-
-      </>
-    );
+const getChildAssetsQuery = gql`
```
elohanlon: After rebasing, could be good to put this in digitalObject.js with the other digital object queries.

app/javascript/hyacinth_ui_v1/components/digital_objects/children/Children.js
```diff
@@ -1,39 +1,61 @@
...
+  query ChildAssets($id: ID!){
+    childAssets(id: $id) {
+      id
+    }
   }
-}
+`;
+
+const Children = (props) => {
+  const { match: { params: { id } } } = props;
```
elohanlon: Since this is a functional component, I think you could use the useParams hook here instead.  If you search through the project code for `useParams`, you'll see some examples.

elohanlon: Since this is a functional component, I think you could use the useParams hook here instead.  If you search through the project code for `useParams`, you'll see some examples.

app/javascript/hyacinth_ui_v1/components/digital_objects/children/Children.js
```diff
@@ -1,39 +1,61 @@
...
+      </Card>
+    </>
+  );
+};
+
+Children.propTypes = {
+  match: PropTypes.object.isRequired,
+};
 
 export default digitalObjectInterface(Children);
```
elohanlon: This component will need some restructuring after your rebase, since lower case d digitalObjectInterface no longer exists.  If you search through the project code for `<DigitalObjectInterface` you'll see examples of how to use the new component.

app/javascript/hyacinth_ui_v1/components/digital_objects/new/AssetNew.js
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Form } from 'react-bootstrap';
+import { useMutation } from '@apollo/react-hooks';
+import uploadFile from '../../../graphql/file_uploads';
+import { createAssetMutation } from '../../../graphql/assets';
+
+function debugObject(obj) {
```
elohanlon: It looks like this isn't being used anywhere, but I could be wrong.

spec/factories/digital_objects/assets.rb
```diff
@@ -0,0 +1,25 @@
...
+    parent { nil } # parent should be passed in when factory build or create or called, otherwise this object won't validate
+    asset_type { BestType.dc_type.for_file_name("foo.xyz") }
+    initialize_with do
+      instance = new
+      instance.instance_variable_set(
+        '@dynamic_field_data',
+        'title' => [
+          {
+            'non_sort_portion' => 'The',
+            'sort_portion' => 'Best Asset Ever'
```
elohanlon: I do believe that this is the best asset ever.

elohanlon: I do believe that this is the best asset ever.

spec/factories/digital_objects/items.rb
```diff
@@ -23,5 +23,11 @@
         digital_object.primary_project = create(:project)
       end
     end
+
+    trait :with_asset do
+      after(:create) do |digital_object|
+        digital_object.append_child_uid(create(:asset, parent: digital_object).uid)
```
elohanlon: One concern about this method of adding an asset to the item factory object is that you've set up a single-directional relationship between these two objects when normally there would be a bidirectional relationship.  Could lead to inconsistencies during testing.

elohanlon: One concern about this method of adding an asset to the item factory object is that you've set up a single-directional relationship between these two objects when normally there would be a bidirectional relationship. Could lead to inconsistencies during testing.

app/javascript/hyacinth_ui_v1/components/digital_objects/new/AssetNew.js
barmintor: it's really just there for debugging!

spec/factories/digital_objects/items.rb
barmintor: Does the asset factory use the parent delta to save the parent again? It might just be redundant.

spec/requests/graphql/mutations/create_asset_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+          }
+        }
+      end
+
+      before { graphql query, variables }
+
+      it 'returns a new asset' do
+        expect(response.body).to have_json_type(String).at_path('data/createAsset/asset/id')
+      end
+      it 'deletes the upload' do
```
elohanlon: Can you also add a test that confirms that the copied file doesn't exist if the asset fails to save (confirming that `cleanup_on_failure`) is being called and doing its job?

app/javascript/hyacinth_ui_v1/components/digital_objects/new/AssetNew.js
elohanlon: Makes sense!  I just mean probably good to remove before we merge this PR.

app/javascript/hyacinth_ui_v1/graphql/file_uploads.js
```diff
@@ -0,0 +1,20 @@
+import * as ActiveStorage from 'activestorage';
```
elohanlon: If a file has a single default export of a variable, I tend to give the file the same name as the single default export, so I'd recommend renaming to `uploadFile.js`.

app/javascript/hyacinth_ui_v1/components/digital_objects/new/AssetNew.js
```diff
@@ -41,11 +40,14 @@ function AssetNew(props) {
   };
   return (
     <Form>
-      <input id={elementId} type="file" required onChange={onChange} parentid={parentId} direct_upload="true" />
+      <input id={AssetNew.elementId} type="file" required onChange={onChange} parentid={parentId} direct_upload="true" />
     </Form>
   );
 }
 
+AssetNew.elementId = `input-${String.toString(Math.random()).replace('.', '')}`;
```
elohanlon: I think this goes against good reusable component design, since you're effectively creating a static variable that will be the same for every occurrence of this component for the lifetime of the react app.  Things like unique element id would be better off set in the component's state.

app/javascript/hyacinth_ui_v1/components/digital_objects/new/AssetNew.js
```diff
@@ -0,0 +1,56 @@
...
+      .then(resetFileInput);
+  };
+  return (
+    <Form>
+      <input id={AssetNew.elementId} type="file" required onChange={onChange} parentid={parentId} direct_upload="true" />
+    </Form>
+  );
+}
+
+AssetNew.elementId = `input-${String.toString(Math.random()).replace('.', '')}`;
```
elohanlon: I think this goes against good reusable component design, since you're effectively creating a static variable that will be the same for every occurrence of this component for the lifetime of the react app. Things like unique element id would be better off set in the component's state.

app/models/digital_object/asset.rb
```diff
@@ -16,5 +18,12 @@ class Asset < DigitalObject::Base
     def initialize
       super
     end
+
+    # yields to a block with the primary resource name and object
+    # returns the resource
+    def with_primary_resource(&block)
```
elohanlon: 👍

spec/factories/digital_objects/items.rb
elohanlon: Yeah, I think the `append_child_uid` part is redundant.

You can try replacing:
`digital_object.append_child_uid(create(:asset, parent: digital_object).uid)`
with
`create(:asset, parent: digital_object)`

elohanlon: Actually, then your actual factory-returned instance of the parent object wouldn't be reloaded to recognize the addition, so I guess it does make sense to keep it as is.

elohanlon: We just need to ensure, long-term, that whatever parent adding logic happens inside of the digital object save method is mirrored here in the `with_asset` trait.  Hyacinth generally assumes that children add their parents, rather than the other way around.

app/javascript/hyacinth_ui_v1/graphql/file_uploads.js
barmintor: I think there will be other functions here when I decompose after Rails 6

app/javascript/hyacinth_ui_v1/components/digital_objects/new/AssetNew.js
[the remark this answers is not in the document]
barmintor: Added to Children component state to be refactored when we expand to multiple uploads

app/javascript/hyacinth_ui_v1/components/digital_objects/children/Children.js
barmintor: it is done, with a new query response type that dereferences the child objects for display and the parent for the interface.

app/javascript/hyacinth_ui_v1/components/digital_objects/children/Children.js
barmintor: params matching is totally unnecessary, so I removed it in favor of a simple destructuring assignment.

app/javascript/hyacinth_ui_v1/components/digital_objects/children/Children.js
barmintor: done per SystemData, etc.

app/javascript/hyacinth_ui_v1/components/digital_objects/new/AssetNew.js
barmintor: Removed.

spec/requests/graphql/mutations/create_asset_spec.rb
barmintor: This is harder to do with random resource storage locations.

app/javascript/hyacinth_ui_v1/components/digital_objects/new/AssetNew.js
barmintor: Done.

app/javascript/hyacinth_ui_v1/components/digital_objects/DigitalObjectList.js
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Card } from 'react-bootstrap';
+import { LinkContainer } from 'react-router-bootstrap';
+import { has } from 'lodash';
+
+const titleForDigitalObject = (digObj) => {
```
elohanlon: This function isn't necessary now that `title` is a query-able field in the DigitalObject graphql query.  This might also eliminate the need for retrieving dynamicFieldData in getChildStructureQuery, if that's the only reason you were including it.

app/javascript/hyacinth_ui_v1/components/digital_objects/children/Children.js
elohanlon: Great, thanks!

spec/requests/graphql/mutations/create_asset_spec.rb
elohanlon: Maybe you can mock the uri generation response?  I'm just concerned that we're not testing it.

app/javascript/hyacinth_ui_v1/graphql/file_uploads.js
elohanlon: 👍

app/javascript/hyacinth_ui_v1/components/digital_objects/new/AssetNew.js
elohanlon: 👍

app/javascript/hyacinth_ui_v1/components/digital_objects/children/Children.js
elohanlon: Great, thanks!

app/javascript/hyacinth_ui_v1/components/digital_objects/DigitalObjectList.js
barmintor: I removed these two functions and patched the relevant REST controller to mimic the gql query. I didn't tag the commit with the issue number, since it seems like a standalone thing.

spec/requests/graphql/mutations/create_asset_spec.rb
barmintor: Ok, an awkward test but it is in place.

elohanlon: Thanks for adding that.  Checking it out now.

app/javascript/hyacinth_ui_v1/components/digital_objects/DigitalObjectList.js
elohanlon: Good move, thanks!

spec/requests/graphql/mutations/create_asset_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+      let(:expected_error_message) { "Testing upload deletes" }
+
+      before do
+        authorized_object
+        expect(Hyacinth::Config.resource_storage).to receive(:generate_new_location_uri).and_return(predictable_location_uri)
+        call_count = 0
+        allow(Hyacinth::Config.metadata_storage).to receive(:write) do |location, content|
+          call_count += 1
+          raise expected_error_message if call_count > 1
+          original_write.call(location, content)
```
elohanlon: @barmintor Where is `original_write` defined?

barmintor: you got it.

barmintor: oh, wait - is it? But if it's not, why does this test run is the question.

barmintor: I don't think this can be in a request spec, not without doing very questionable things. I think we should remove this commit from the PR and open a ticket for unit testing the graphql resolvers.

barmintor: "not without doing very questionable things" or digging into the design of digital object concerns and storage adapters, which I think is out of scope of the ticket.

barmintor: I'm pushing a non-request spec that uses mocks and isolates the mutation.

elohanlon: Yeah, I agree.  Ticketing and handling outside of PR sounds good.  Also, maybe we should consider moving the cleanup stuff into the model if possible, rather than the graphql mutation.

spec/graphql/mutations/create_asset_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      before do
+        authorized_object
+        expect(mutation).to receive(:initialize_child_asset).and_return(asset)
+        expect(context).to receive(:[]).with(:ability).and_return(ability)
+        allow(ability).to receive(:authorize!).and_return(true) # not testing auth
+        # this failed save should trigger a delete of the newly created asset object
+        expect(asset).to receive(:save!).and_raise(expected_error_message)
+        expect(Hyacinth::Config.resource_storage).to receive(:generate_new_location_uri).and_return(predictable_location_uri)
+        expect(Hyacinth::Config.resource_storage).to receive(:delete).with(predictable_location_uri).and_call_original
+      end
```
elohanlon:
```suggestion
        expect(Hyacinth::Config.resource_storage).to receive(:with_writeable).and_wrap_original do |m, *args, &b|
          result = m.call(*args, &b)
          expect(args[0]).to eq(predictable_location_uri)
          expect(Hyacinth::Config.resource_storage.exists?(args[0])).to eq(true)
          result
        end
      end
```

elohanlon:
@barmintor Thanks a lot for all of the updates.  This is great.  I added one minor suggestion to your test, just to verify that the file exists before we verify that it was deleted.  Okay if we include that?

Just approved the PR for merging. 👍

PR #196, opened 2020-01-17
Hyacinth 385 hiding permission restricted ui elements

elohanlon:


app/javascript/hyacinth_ui_v1/components/digital_objects/preserve_publish/PreservePublish.js
```diff
@@ -23,19 +24,22 @@ function PreservePublish(props) {
   if (digitalObjectError) return (<GraphQLErrors errors={digitalObjectError} />);
   const { digitalObject } = digitalObjectData;
 
-  return (
-    <DigitalObjectInterface digitalObject={digitalObject}>
-      <TabHeading>
-        Preserve / Publish
-      </TabHeading>
+  const canUpdateObject = digitalObjectAbility.can('update_objects', { primaryProject: digitalObject.primaryProject, otherProjects: digitalObject.otherProjects });
```
barmintor: Why doesn't it use the Can component here?

elohanlon: The `<Can>` component can only handle single subject checks, but we're checking against both the primary project and array of projects.  It's the same reason why we couldn't use `<ProtectedRoute>` for digital object project-related checks.  We could create a <DigitalObjectCan> component if you and @cgalarza like the sound of that.  If good, I can create a ticket for that for future refactoring.

cgalarza: It would be nice, but i don't think it's necessary. Up to you guys.

PR #200, opened 2020-01-22
Rails 6

elohanlon:
Ready for review and local browser testing.


config/environments/deployed.rb
```diff
@@ -26,26 +24,23 @@
   # Apache or NGINX already handles this.
   config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?
 
-  # Compress JavaScripts and CSS.
-  config.assets.js_compressor = :uglifier
+  # Compress CSS using a preprocessor.
```
cgalarza: Do we not need to define the js_compressor? Or does it default to something we like?

config/environments/development.rb
```diff
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
 Rails.application.configure do
-  # Verifies that versions and hashed value of the package contents in the project's package.json
-  config.webpacker.check_yarn_integrity = true
```
cgalarza: We don't need this anymore? or is this flag set somewhere else?

db/migrate/20200109190635_create_active_storage_tables.active_storage.rb
```diff
@@ -1,5 +1,5 @@
 # This migration comes from active_storage (originally 20170806125915)
-class CreateActiveStorageTables < ActiveRecord::Migration[5.2]
+class CreateActiveStorageTables < ActiveRecord::Migration[6.0]
```
cgalarza: Is it okay to just change the version like that? Or should we squash all the migrations?

config/environments/deployed.rb
elohanlon: Good question.  The absence of a js_compressor setting is the default for new Rails 6 apps and seems to be unnecessary, so that's why I removed it.  I think it's okay because we're handling js compilation via yarn.  I should test deployment though, to ensure that this won't be a problem.  I'll follow up with another comment after that test!

elohanlon: So interestingly enough, the latest rails docs say:
```
Possible options for JavaScript compression are :closure, :uglifier and :yui. These require the use of the closure-compiler, uglifier or yui-compressor gems, respectively.
```
I'm not sure what the default is if you don't specify it, but it might actually be a nil default, which would make sense because out Gemfile.lock doesn't include the closure-compiler, uglifier or yui-compressor gems anywhere.

config/environments/development.rb
elohanlon: That doesn't appear in the rails 6 environment files anymore, and I believe it's because the preferred approach is to add the integrity check to webpacker.yml instead, and we have it there for the development environment.

db/migrate/20200109190635_create_active_storage_tables.active_storage.rb
elohanlon: Yes, as long as we're confident that the 6.0 migration API defaults don't cause any issues for our existing migration code, and I didn't see any issues.

I'm doing this so that we don't one day find ourselves upgrading to Rails 7 and still having migrations from 5.2 that we need to re-evaluate (if they drop support for the 5.2 migration API at that time).  We'll be a little closer.  Since I did the evaluation already, it'll be reduced future tech debt.

I'm fine with us squashing the migrations, but I think that should be part of a separate PR.

config/environments/development.rb
cgalarza: 👍

db/migrate/20200109190635_create_active_storage_tables.active_storage.rb
cgalarza: 👍

config/environments/deployed.rb
cgalarza: 👍

config/environments/development.rb
barmintor: oh that's interesting.

db/migrate/20200109190635_create_active_storage_tables.active_storage.rb
barmintor: I had the same question! I guess the real test will be when we add the next migration.

elohanlon:
Test deploy successful!  Will merge with a merge commit.

PR #202, opened 2020-01-24
Graphql dynamic fields hyacinth 399 wip

barmintor:


barmintor:
Opening a WIP PR for review, debugging, and work distribution.

app/graphql/mutations/update_dynamic_field_category.rb
```diff
@@ -0,0 +1,19 @@
...
+
+class Mutations::UpdateDynamicFieldCategory < Mutations::BaseMutation
+  argument :id, Integer, required: true
+  argument :display_label, String, required: false
+  argument :sort_order, Integer, required: false
+
+  field :dynamic_field_category, Types::DynamicFieldCategoryType, null: true
+
+  def resolve(id:, **attributes)
+    dynamic_field_category = DynamicFieldCategory.find_by!(id: id)
```
cgalarza: We might be able to use `find!` here! That's kinda exciting.

app/graphql/types/dynamic_field_group_type.rb
```diff
@@ -0,0 +1,20 @@
...
+    field :id, ID, null: false
+    field :string_key, ID, null: false
+    field :display_label, String, null: false
+    field :sort_order, Integer, null: false
+    field :is_repeatable, Boolean, null: false
+    field :export_rules, [GraphQL::Types::JSON], null: true
+    field :created_by, UserType, null: true
+    field :updated_by, UserType, null: true
+    field :children, [DynamicFieldGroupChildType], null: true
+    field :parent, DynamicFieldCategoryType, null: false
```
cgalarza: The parent here can be both a DynamicFieldCategory and a DynamicFieldGroup, so we need a union type here, like for children.

app/graphql/mutations/create_dynamic_field.rb
```diff
@@ -0,0 +1,29 @@
...
+  argument :dynamic_field_group_id, ID, required: true
+
+  field :dynamic_field, Types::DynamicFieldType, null: true
+
+  def resolve(**attributes)
+    ability.authorize! :create, DynamicField
+
+    attributes[:created_by] = context[:current_user]
+    attributes[:updated_by] = context[:current_user]
+    dynamic_field = DynamicField.new(**attributes)
```
cgalarza: I think we can use `create!` here, instead of saving as a separate action.

cgalarza:
@JackBlackLight I just went through some of the PR and left some comments. It looks great. Should I review everything now? or should I hold off a bit?

app/graphql/mutations/create_dynamic_field.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+class Mutations::CreateDynamicField < Mutations::BaseMutation
+  argument :string_key, ID, required: true
+  argument :display_label, String, required: true
+  argument :sort_order, Integer, required: true
+  argument :is_facetable, Boolean, required: true
```
cgalarza: I don't think the `is_*` fields need to be required. The database will assign default values if they aren't provided.

app/graphql/mutations/create_dynamic_field_category.rb
```diff
@@ -0,0 +1,18 @@
...
+class Mutations::CreateDynamicFieldCategory < Mutations::BaseMutation
+  argument :display_label, String, required: true
+  argument :sort_order, Integer, required: true
+
+  field :dynamic_field_category, Types::DynamicFieldCategoryType, null: true
+
+  def resolve(**attributes)
+    ability.authorize! :create, DynamicFieldCategory
+
+    dynamic_field_category = DynamicFieldCategory.new(**attributes)
```
cgalarza: Same here, I think we can use `create!`.

app/graphql/mutations/update_dynamic_field.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+class Mutations::UpdateDynamicField < Mutations::BaseMutation
+  argument :string_key, ID, required: true
+  argument :display_label, String, required: true
+  argument :sort_order, Integer, required: true
+  argument :is_facetable, Boolean, required: true
```
cgalarza: Same here, I don't think the `is_*` fields should be required.

app/graphql/mutations/create_dynamic_field_category.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class Mutations::CreateDynamicFieldCategory < Mutations::BaseMutation
+  argument :display_label, String, required: true
+  argument :sort_order, Integer, required: true
```
cgalarza: `sort_order` actually doesn't have to be required by this endpoint even though it is required in the database schema. There is logic in `DynamicFieldStructure::Sortable` that sets `sort_order` if one isn't present. It's kind of a hidden feature.

cgalarza: The same applies to dynamic field group and dynamic field

app/graphql/mutations/create_dynamic_field_group.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class Mutations::CreateDynamicFieldGroup < Mutations::BaseMutation
+  argument :string_key, ID, required: true
+  argument :display_label, String, required: true
+  argument :sort_order, Integer, required: true
```
cgalarza: `sort_order` and `is_repeatable` can be optional here.

app/graphql/types/dynamic_field_group_child_type.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module Types
+  class DynamicFieldGroupChildType < Types::BaseUnion
+    description "objects that can be children of a dynamic field group"
+    possible_types Types::DynamicFieldGroupType, Types::DynamicFieldType
+    def self.resolve_type(object, _context)
+      if object.is_a?(DynamicFieldGroup)
+        Types::DynamicFieldGroupType
+      else
```
cgalarza: I would recommend explicitly checking that its a "DynamicField" and then throwing an error if its not.

app/graphql/types/dynamic_field_group_type.rb
```diff
@@ -0,0 +1,20 @@
...
+    field :parent_id, ID, null: false
```
cgalarza: Even though we require `parent_id` and `parent_type` for the dynamic field group mutations, i don't think we need to expose those fields here because they should be available via the `parent` field.

app/graphql/mutations/create_dynamic_field_group.rb
elohanlon: If we are allowing sort_order as an argument, then the resolve method will probably need to handle updating the sort_order of adjacent elements.  As in:

field_a (order 1)
field_b (order 2)
field_c (order 3)

running this: create field_z with order 2
should result in updates to the order values of field_b and field_c:

field_a (order 1)
field_z (order 2)
field_b (order 3)
field_c (order 4)

app/graphql/mutations/create_dynamic_field_category.rb
elohanlon: Related: Please see my comment about sort_order in `Mutations::CreateDynamicFieldGroup`.

app/graphql/mutations/create_dynamic_field.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+class Mutations::CreateDynamicField < Mutations::BaseMutation
+  argument :string_key, ID, required: true
+  argument :display_label, String, required: true
+  argument :sort_order, Integer, required: true
```
elohanlon: Related: Please see my comment about sort_order in `Mutations::CreateDynamicFieldGroup`.

app/graphql/mutations/create_dynamic_field_group.rb
```diff
@@ -0,0 +1,25 @@
...
+  argument :parent_type, String, required: true
+
+  field :dynamic_field_group, Types::DynamicFieldGroupType, null: true
+
+  def resolve(**attributes)
+    ability.authorize! :create, DynamicFieldGroup
+
+    attributes[:created_by] = context[:current_user]
+    attributes[:updated_by] = context[:current_user]
+    dynamic_field_group = DynamicFieldGroup.new(**attributes)
```
elohanlon: We can also use `DynamicFieldGroup.create!` here.

app/graphql/mutations/update_dynamic_field.rb
elohanlon: I think we also don't need `required: true` for sort_order here, since there will be many cases where we're either ONLY updating dynamic field info without regard to sort_order.

And also, now that I think about it, there will be times when we're only updating sort order.  That may be a separate mutation though, one that factors in the relative order of fields and field groups.

app/graphql/mutations/create_dynamic_field_group.rb
cgalarza: Last time we talked about this, we talked about just doing separate requests if we wanted to update the order of a whole list of elements.

app/graphql/mutations/update_dynamic_field_category.rb
elohanlon: It's so rare in Hyacinth, haha.

app/javascript/hyacinth_ui_v1/components/dynamic_fields/DynamicFieldIndex.js
```diff
@@ -5,73 +5,92 @@ import produce from 'immer';
...
-  }
 
-  componentDidMount() {
-    this.getDynamicFieldCategories();
-  }
+import { Table } from 'react-bootstrap';
+
+
+
+  const DYNAMICFIELDS = gql`
```
elohanlon: All caps for a constant is common in other languages, but in JavaScript this is actually an uncommon thing to do for a scoped (in this case, scoped to the file its defined in) variable that just happens to be an unchangeable `const`.  I'd recommend going for regular camelCase here.

spec/requests/graphql/mutations/create_dynamic_field_category_spec.rb
```diff
@@ -0,0 +1,83 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Mutations::CreateDynamicFieldCategory, type: :request do
+  describe '.resolve' do
```
elohanlon: When doing a request test, we want to think of this as a test of the external API rather than a test of a particular method (like `resolve`), so you should be able to remove the `describe '.resolve' do` block and just run the stuff inside.

spec/requests/graphql/mutations/create_dynamic_field_group_spec.rb
```diff
@@ -0,0 +1,78 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Mutations::CreateDynamicFieldGroup, type: :request do
+  let(:parent) { FactoryBot.create(:dynamic_field_category) }
+  # let!(:field_export_profile) { FactoryBot.create(:field_export_profile) }
+
+  describe '.resolve' do
```
elohanlon: Same note here about not needing this describe wrapper.

spec/requests/graphql/mutations/create_dynamic_field_spec.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Mutations::CreateDynamicField, type: :request do
+  let(:dynamic_field) { FactoryBot.create(:dynamic_field) }
+  let(:parent) { FactoryBot.create(:dynamic_field_group) }
+
+
+  describe '.resolve' do
```
elohanlon: Same note here about not needing this describe wrapper.

spec/requests/graphql/mutations/update_dynamic_field_group_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+          isRepeatable: true,
+          parentId: 1,
+          parentType: 'DynamicFieldCategory' 
+          } 
+        }
+      end
+
+      before { 
+        graphql query, variables 
+        puts response.body
```
barmintor: stray `puts` should be cleaned up here, you noticed while we were talking.

JackBlackLight:
Hi, I've addressed the comments in this PR and added additional tests. 

One test is failing -  'adds child to parent' in create_dynamic_field_group_spec - so I'm still looking at that.

spec/requests/graphql/mutations/update_dynamic_field_category_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    let(:variables) { { input: { id: dynamic_field_category.id, displayLabel: 'Best Dynamic Field Category' } } }
+    let(:request) { graphql query, variables }
+  end
+
+  context 'when logged in user has appropriate permissions' do
+    before { sign_in_user as: :administrator }
+
+    context 'when updating record' do
+      let(:variables) do
+        { input: { id: dynamic_field_category.id, displayLabel: 'Best Dynamic Field Category' } }
```
cgalarza: I would test updating a few more of the DynamicFieldCategories fields here.

spec/requests/graphql/mutations/update_dynamic_field_group_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+        }
+      end
+
+      before { 
+        graphql query, variables
+      }
+
+      it 'correctly updates record' do
+        dynamic_field_group.reload
+        expect(dynamic_field_group.display_label).to eql 'Best Dynamic Field Group'
```
cgalarza: I would add tests here that check that `sort_order` and `isRepeatable` are correctly updated.

spec/requests/graphql/mutations/update_dynamic_field_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+     let(:variables) { { input: { stringKey: dynamic_field.string_key, displayLabel: 'New Location' } } }
+    let(:request) { graphql query, variables }
+  end
+
+
+  context 'when logged in user is an administrator' do
+    before { sign_in_user as: :administrator }
+
+    context 'when updating display_label' do
+      let(:variables) do
```
cgalarza: Again, I would check that a few other fields can be successfully updated.

app/graphql/types/dynamic_field_collation_type.rb
```diff
@@ -0,0 +1,16 @@
...
+
+module Types
+  class DynamicFieldCollationType < Types::BaseUnion
+    description 'Properties of a Collation of Dynamic Fields'
+    possible_types DynamicFieldCategoryType, DynamicFieldGroupType
+
+    def self.resolve_type(object, _context)
+      if object.is_a?(DynamicFieldCategory)
+        DynamicFieldCategoryType
+      else object.is_a?(DynamicFieldGroup)
```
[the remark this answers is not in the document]
JackBlackLight: corrects else/elsif, but allowing graphql to raise error

app/graphql/types/dynamic_field_group_child_type.rb
```diff
@@ -0,0 +1,16 @@
...
+
+module Types
+  class DynamicFieldGroupChildType < Types::BaseUnion
+    description "objects that can be children of a dynamic field group"
+    possible_types DynamicFieldGroupType, DynamicFieldType
+
+    def self.resolve_type(object, _context)
+      if object.is_a?(DynamicFieldGroup)
+        Types::DynamicFieldGroupType
+      else object.is_a?(DynamicField)
```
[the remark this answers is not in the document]
JackBlackLight: corrects else/elsif, but allowing graphql to raise error

spec/requests/graphql/query/dynamic_field_category_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+RSpec.describe 'Retrieving Dynamic Field Category', type: :request do
+  let(:dynamic_field_category) { FactoryBot.create(:dynamic_field_category) }
+
+  before { sign_in_user }
+
+  context 'when id is valid' do
+    before { graphql query(dynamic_field_category.id) }
+
+    it 'returns 200' do
```
cgalarza: I just noticed that this test can be deleted.

spec/requests/graphql/query/dynamic_field_group_spec.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'Retrieving Dynamic Field Group', type: :request do
+  let(:dynamic_field_group) { FactoryBot.create(:dynamic_field_group) }
+  let(:field_export_profile) { FactoryBot.create(:field_export_profile) }
```
cgalarza: I think this can also be deleted?

spec/requests/graphql/mutations/update_dynamic_field_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+        expect(dynamic_field.controlled_vocabulary).to eql 'vocabulary2'
+      end
+    end
+
+    context 'when updating select options' do
+      let(:variables) do
+        {
+          input: {
+            stringKey: dynamic_field.string_key,
+            selectOptions: '{}'
```
elohanlon: @JackBlackLight would you mind setting selectOptions to a non-empty hash for this test?  I think it'll make it a more resilient test in case somehow we ever default to an empty hash for a select field.

PR #204, opened 2020-01-27
Hyacinth 482

barmintor:
add the asset rights flag to project graphql type to drive conditional display of asset rights form
part of HYACINTH-336

### DIFF
--- a/app/graphql/types/project_type.rb
+++ b/app/graphql/types/project_type.rb
@@ -24,6 +24,8 @@ module Types
 
     field :enabled_digital_object_types, [String], null: false
 
+    field :has_asset_rights, Boolean, null: false
+
     def field_set(id:)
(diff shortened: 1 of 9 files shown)


app/graphql/types/project_type.rb
```diff
@@ -24,6 +24,8 @@ class ProjectType < Types::BaseObject
 
     field :enabled_digital_object_types, [String], null: false
 
+    field :has_asset_rights, Boolean, null: true
```
fcd1: Given my lack of experience with GraphQL, this may be completely off-base, but in my code I had 'null: false' because:
1) Mimics column definition in db.
2) Set to false for similar fields, such as is_primary.

barmintor: I didn't see the default, let me fix this and push the change

PR #208, opened 2020-01-29
DigitalObject::Asset validations confirm acceptability of rights assessment (HYACINTH-483)

barmintor:
PR to umbrella ticket branch

### DIFF
--- a/app/models/concerns/digital_object_concerns/assets/validations.rb
+++ b/app/models/concerns/digital_object_concerns/assets/validations.rb
@@ -5,5 +5,11 @@ module DigitalObjectConcerns::Assets::Validations
 
   included do
     validates :asset_type, inclusion: { in: Hyacinth::DigitalObject::AssetType::VALID_TYPES, message: "Invalid asset type: %{value}" }
+    validate :validate_rights_updates
+  end
+
+  def validate_rights_updates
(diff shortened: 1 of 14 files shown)


app/models/concerns/digital_object_concerns/assets/validations.rb
```diff
@@ -5,5 +5,11 @@ module DigitalObjectConcerns::Assets::Validations
 
   included do
     validates :asset_type, inclusion: { in: Hyacinth::DigitalObject::AssetType::VALID_TYPES, message: "Invalid asset type: %{value}" }
+    validate :validate_rights_updates
+  end
+
+  def validate_rights_updates
+    return unless rights.keys.detect { |key| rights[key].present? && !'access_condition'.eql?(key.to_s) }
```
cgalarza: I think i am calling the 'access_condition section `restriction_on_access` to follow what the field group was called in HY2. Which should we follow?

barmintor: I am spitballing based on what we had in master. Can we update the subsection key in your branch after merge?

cgalarza: hmm, I wondering if this should wait until after I merge my code because i have new rights mutations in place.

barmintor:
Closing in favor of #213

PR #210, opened 2020-01-29
Hyacinth 336 fcd1

fcd1:
UI for updating/displaying the asset-rights flag (HYACINTH-339)


barmintor:
This looks good - the only things I think are missing are: updating the graphql request spec for project create/update, and amending the project validation for `has_asset_rights`

PR #212, opened 2020-01-30
Validation: Asset Rights flag can only be true if project is a Primar…

fcd1:
…y project


barmintor:
merged via cherry-pick into hyacinth-336

PR #213, opened 2020-01-30
DigitalObject::Asset validations confirm acceptability of rights assessment (HYACINTH-483)

barmintor:


spec/requests/api/v1/digital_objects/rights_spec.rb
```diff
@@ -74,6 +74,60 @@
           JSON.parse(response.body)['digital_object']['rights'].to_json
         ).to be_json_eql(properties[:digital_object][:rights].to_json)
       end
+      context "to assess an Asset object" do
```
cgalarza: This test will have to be migrated to work with mutations, but we can work on that separately. I can add it in as part of my work.

PR #217, opened 2020-02-04
2.x captions

barmintor:


app/controllers/concerns/hyacinth/digital_objects/captions.rb
```diff
@@ -38,7 +69,11 @@ def clear_captions_and_reimport_transcript
 
     # Never trust parameters from the scary internet, only allow the white list through.
     def captions_params
-      params.permit(:captions_text)
+      params.permit(:captions_vtt, :file)
+    end
+
+    def synchronized_transcript_params
+      params.permit(:synchronized_transcript_text)
```
elohanlon: Do we need to permit `:file` here?

barmintor: We don't have the corresponding form object/ability to edit as text for the synchronized transcript - only what comes from the synchronizer.

elohanlon: Ah right!  In that case, never mind!  Approving the PR now. 👍

PR #218, opened 2020-02-05
Cleaning rights and dynamic field data

cgalarza:


spec/hyacinth/utils/clean_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+          {
+            'collection_authorized_term_uri' => 'http://example.com/a/b/c'
+          }
+        ],
+        'an_array_field' => ['some_value']
+      }
+    end
+
+    it 'cleans data as expected' do
+      expect(described_class.remove_blank_fields!(hash_with_blank_fields)).to eq(expected_hash_without_blank_fields)
```
elohanlon:
```suggestion
      expect(described_class.remove_blank_fields!(hash_with_blank_fields)).to eq(expected_hash_without_blank_fields)
      expect(hash_with_blank_fields).to eq(expected_hash_without_blank_fields)
```
Just to prove that the original has been modified.

spec/hyacinth/utils/clean_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+            'collection_authorized_term_uri' => 'http://example.com/a/b/c'
+          }
+        ],
+        'note' => [
+          {
+            'note_value' => '                         ', # A bunch of spaces
+            'note_type' => ''
+          }
+        ],
+        'an_array_field' => ['', nil, 'some_value', ['', nil]]
```
elohanlon: How would you feel about adding `true` and `false` example values in the above array, to test out boolean handling?

spec/models/concerns/digital_object_concerns/digital_object_data/setters/rights_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+            'countryOfOrigin' => nil,
+            'filmDistributedToPublic' => nil,
+            "filmDistributedCommercially" => nil
+          }
+        ],
+        'copyrightStatus' => [
+          {
+            'copyrightNote' => '   ',
+            'copyrightRegistered' => true,
+            'copyrightRenewed' => true,
```
elohanlon: Do you mind if we change this `true` to `false` just to prove that `false` values aren't removed by the cleaning?

cgalarza:
@elohanlon I was thinking about the true/false cleanup and realized that if we don't remove false values for every dynamic field group that has some kinda checkbox we are going to end up with bits of data that look like `[{ is_something: false }]`. I think its best if we remove false values, like we were and I will adjust the fields in the rights module to not use booleans when "false" values need to be saved. Let me know what you think!

I'll go through and add your suggested changes.

elohanlon:
@cgalarza That sounds good to me!

PR #221, opened 2020-02-10
Hyacinth 494

barmintor:


app/graphql/types/pageable.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module Types
+  module Pageable
```
elohanlon: Is there any benefit to making this extend ActiveSupport::Concern? (if that would work)

spec/requests/graphql/query/digital_objects_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    let(:expected_response) do
+      %(
+        [
+          { "id": "#{authorized_object.uid}", "title": "The Best Item Ever", "digitalObjectType": "item" }
+        ]
+      )
+    end
+
+    it "return digital objects with expected fields" do
+      puts response.body
```
cgalarza: Can be removed?

app/graphql/types/pageable.rb
barmintor: I couldn't think of one but if you can?

spec/requests/graphql/query/digital_objects_spec.rb
barmintor: dangit

app/graphql/types/pageable.rb
elohanlon: I guess it's the same as asking "why ever use ActiveSupport::Concern in a rails app at all?"

The reason the question crossed my mind is because we've been using it for included modules in other places.  We don't have to, but I just thought I'd put it out there.  Up to you.

elohanlon: Thinking more about it, using `ActiveSupport::Concern` is a little DRYer, right?

spec/requests/graphql/query/digital_objects_spec.rb
barmintor: done!

PR #222, opened 2020-02-11
HYACINTH-476-adapter-refactor

elohanlon:


.rubocop_todo.yml
```diff
@@ -1,6 +1,6 @@
 # This configuration was generated by
```
cgalarza: Did you generate this rubocop todo with `rubocop --auto-gen-config  --auto-gen-only-exclude --exclude-limit 10000`? or something like it?

elohanlon: I used `rubocop --auto-gen-config --exclude-limit 10000` because that's what the comment in the previous TODO said was used, but I can re-run with `--auto-gen-only-exclude` if that's better.

elohanlon: Ah, I see how `--auto-gen-only-exclude` would help here.  It would avoid having the todo add things like this:

```
Metrics/AbcSize:
  Exclude:	  Max: 51
```

I'll do that now!

elohanlon: Updated!

app/graphql/mutations/create_asset.rb
```diff
@@ -36,12 +36,12 @@ def initialize_child_asset(parent, file_name)
     asset
   end
 
-  def initialize_asset_resource_for_blob(asset, blob, storage)
```
barmintor: I don't understand the reason for this change (it looks like the parameterized reference has been exchanged for a static reference in the method) and it seems like it will make it somewhat more difficult to test. Is there a context for it?

lib/hyacinth/adapters/storage_adapter/abstract.rb
```diff
@@ -5,26 +5,30 @@ module Adapters
...
+        def uri_prefix
+          "#{@uri_protocol}://"
         end
 
-        # @return [string] the expected prefix for a location_uri associated with this adapter ('disk://', 'memory://', etc.)
-        def uri_prefix
+        # @param location_uri [String]
+        # @return [Boolean] true if this adapter can handle this type of location_uri
+        def handles?(location_uri)
+          location_uri.start_with?(uri_prefix)
```
barmintor: This is an existing bug, but: Should `handles?(nil)` return false or throw an error?

app/graphql/mutations/create_asset.rb
elohanlon: It seemed like an opportunity to remove an unneeded param and match with the way that`Hyacinth::Config.resource_storage` was called in other places in the app (and the associated test modification wasn't complicated), but if you'd prefer the previous way I'm glad to change it back.

lib/hyacinth/adapters/storage_adapter/abstract.rb
elohanlon: Good catch, thanks!  I just found the same thing in the `PreservationAdapter` class and updated it there too.

PR #228, opened 2020-02-19
Updating dynamic field components

cgalarza:
Updating dynamic field, dynamic field group and dynamic field categories to use our graphql endpoint. In order to support all the requirements of this components I had to add/change parts of the schema.

Disclaimer: In some cases, there are more efficient ways to query but for expedience sake I decided to implement what was faster. We can revisit when we have more time.


cgalarza:
@elohanlon Will do!

cgalarza:
@elohanlon Looking back, do you think `collect_path` should be a class method instead of an instance method?

elohanlon:
@cgalarza Yeah, I agree.  Class method is a good way to go.

PR #230, opened 2020-02-19
Hyacinth 467 csv export

elohanlon:


elohanlon:
FYI: More to come in a future PR, related to this high-level Batch Export ticket.  This is just a stable set of code that I'd like to get into master.

app/graphql/mutations/batch_export/create_batch_export.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class Mutations::BatchExport::CreateBatchExport < Mutations::BaseMutation
+  argument :search_params, String, required: true
+
+  field :batch_export, Types::BatchExportType, null: true
+
+  def resolve(**attributes)
+    # Note: No ability-based authorization for creating a CSV export.
```
cgalarza: I would honestly still do the ability check here. I know it will pass for all users, but if it changes in the future it offers us protection. We currently do the check for mutations like create new term even though everybody can create a new term.

app/models/ability.rb
```diff
@@ -22,6 +22,8 @@ def initialize(user)
       can [:read, :create], Term
       can :read, Vocabulary           # Need to allow this because Terms are nested under vocabularies
       can :read, DynamicFieldCategory # Need to allow this so we can render EnabledDynamicField pages.
+      can :read, BatchExport, user_id: user.id
+      can :destroy, BatchExport, user_id: user.id
```
cgalarza: Per my other comment I would add `can :create, BatchExport`

lib/hyacinth/storage/abstract_storage.rb
```diff
@@ -63,7 +64,11 @@ def delete(location)
       # Returns the first compatible storage adapter for the given location, or nil if no compatible storage adapter is found.
       def storage_adapter_for_location(location)
         adapter = @storage_adapters.find { |ad| ad.handles?(location) }
-        Rails.logger.error("Tried to find storage adapter for #{location}, but no adapter was found.")
+        if adapter.nil?
```
cgalarza:
```suggestion
        if adapter = @storage_adapters.find { |ad| ad.handles?(location) }
```

barmintor: probably `unless`?

app/graphql/mutations/batch_export/create_batch_export.rb
elohanlon: Good idea.  Will do!

spec/requests/graphql/mutations/batch_export/delete_batch_export_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  include_examples 'requires user to have correct permissions for graphql request' do
+    let(:variables) { { input: { id: id } } }
+    let(:request) { graphql query, variables }
+  end
+
+  context 'when user is logged in' do
+    before do
+      login_as user, scope: :user
+      batch_export # create the batch export
```
cgalarza: I don't think you need this because referencing `id` will generate the `batch_export`

lib/hyacinth/storage/abstract_storage.rb
cgalarza: yup, totally right, i'll fix it in the suggestion

app/models/ability.rb
elohanlon: 👍

spec/requests/graphql/mutations/batch_export/delete_batch_export_spec.rb
elohanlon: Ah, right!  Will remove, thanks!

lib/hyacinth/storage/abstract_storage.rb
elohanlon: Method updated as per our gchat conversation.  Thanks!

PR #231, opened 2020-02-20
digital object type facet is functional (HYACINTH-492)

barmintor:
- import Card in DigitalObjectSearch to handle empty result set


barmintor:
Still needs query string routing support, but working in React state.

barmintor:
fixes a React import bug
fixes a pagination bug with digital object search results
makes query filters linkable in digital object search
introduces an input object to facilitate translation of the front-end search params to solr params uniformly across app

PR #233, opened 2020-02-21
Hyacinth 467 csv export (resque setup + BatchExportJob)

elohanlon:


README.md
```diff
@@ -107,3 +107,28 @@ If you have an IDE that supports jsconfig.json files (e.g. Visual Studio Code),
...
+
+# stop development solr
+bundle exec rake solr:stop
+```
+
+This development solr runs on a port specified in config/solr_wrapper.yml (8983 by default).
+
+When you run the CI task (`bundle exec rake hyacinth:ci`), it temporarily spins up another solr in parallel while running tests, and this runs on a separate port (8993 by default).
+
+One important thing to know is that solr's built-in start script defines a separate port for solr to listen on for stop commands, and this port is equal to your selected port **MINUS 1000**.  This normally isn't an issue, but note that if you set config/solr_wrapper.yml values for your development and test environments that are exactly 1000 apart, you'll run into problems.  So, for example, don't set the development port to 8983 and the test port to 9983.
```
barmintor: Well, this explains a lot.

PR #241, opened 2020-02-28
No longer using permissionActions module for storing app-wide permission data; Now just retrieving permission data where it's needed

elohanlon:


elohanlon:
Please compare with alternative: https://github.com/cul/ldpd-hyacinth/pull/242

elohanlon:
Talked this over with @barmintor and @cgalarza and we've chosen to go with this PR instead of #242.

PR #245, opened 2020-02-28
Hyacinth 526 assign attributes

elohanlon:


app/models/concerns/digital_object_concerns/attribute_assignment.rb
```diff
@@ -0,0 +1,45 @@
...
+    include DigitalObjectConcerns::AttributeAssignment::State
+    include DigitalObjectConcerns::AttributeAssignment::StructuredChildren
+
+    # A powerful method that can set many of this object's properties in one go, based on the given digital_object_data hash.
+    # This method will raise errors if it is given invalid data (e.g. references to projects or publish targets that don't exist),
+    # so be ready to handle those exceptions. All of the deliberately-thrown exceptions will extend Hyacinth::Exceptions::HyacinthError.
+    # @param digital_object_data [Hash] A hash of data used to update many of this object's properties.
+    # @param merge_dynamic_fields [boolean] If true, merges given dynamic_field_data Hash into into existing dynamic_field_data.
+    #        If false, replaces existing dynamic_field_data with new dynamic_field_data Hash.
+    def assign_attributes(new_digital_object_data, merge_dynamic_fields)
```
cgalarza: Would it be bad to have a default value for `merge_dynamic_fields`?

cgalarza: could we maybe make it a named param and force one to be provided if `dynamic_field_data` is present? It just seems weird to provide the setting when we aren't setting `dynamic_field_data`. We can also make this a ticket and implement later.

spec/models/concerns/digital_object_concerns/attribute_assignment/parent_uids_spec.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe DigitalObjectConcerns::AttributeAssignment::OptimisticLockToken do
```
barmintor:
```suggestion
RSpec.describe DigitalObjectConcerns::AttributeAssignment::ParentUids do
```

spec/models/concerns/digital_object_concerns/attribute_assignment/rights_spec.rb
```diff
@@ -44,7 +44,7 @@
     end
 
     it 'works as expected on dynamic_field_data instance variable' do
-      digital_object_with_sample_data.set_rights('rights' => rights)
+      digital_object_with_sample_data.assign_rights('rights' => rights)
```
barmintor: does assign_rights have the same optional merge behavior as the dynamic field data?

cgalarza: no it doesn't, afaik

barmintor: Should we ticket that? Seems like a potential gotcha down the line.

cgalarza: are you thinking that merge is probably going to be necessary when importing through a CSV? Because I don't see a use case for merging when the data is coming from the form.

barmintor: Yes - for the basic fields for example (if they want to bulk manage copyright status) - but also for API uniformity in two very similar methods/operations.

PR #246, opened 2020-03-01
begin refactoring solr indexing to support rights eval

barmintor:
Some of this is slapdash, but starting to refactor the solr document building to support rights evaluation better.

### DIFF
--- a/app/graphql/types/digital_object_interface.rb
+++ b/app/graphql/types/digital_object_interface.rb
@@ -7,10 +7,6 @@ module Types
 
     description 'A digital object'
 
-    TITLE_DYNAMIC_FIELD_GROUP_NAME = 'title'
-    TITLE_SORT_PORTION_DYNAMIC_FIELD_NAME = 'sort_portion'
-    TITLE_NON_SORT_PORTION_DYNAMIC_FIELD_NAME = 'non_sort_portion'
-
     orphan_types Types::DigitalObject::ItemType, Types::DigitalObject::AssetType, Types::DigitalObject::SiteType
 
     field :id, ID, null: false, method: :uid
@@ -43,7 +39,7 @@ module Types
     end
 
     def title
-      DigitalObjectInterface.title_for(object)
+      object.generate_title
     end
 
     def resources
@@ -55,19 +51,6 @@ module Types
           resource.as_json
         )
       end
-    end
-
-    def self.title_for(object, sortable = false)
-      val = '[No Title]'
-
-      title_field_group = object.dynamic_field_data[TITLE_DYNAMIC_FIELD_GROUP_NAME]
-      if title_field_group.present? && (title_field = title_field_group[0]).present?
-        val = title_field[TITLE_SORT_PORTION_DYNAMIC_FIELD_NAME]
-        non_sort_portion = title_field[TITLE_NON_SORT_PORTION_DYNAMIC_FIELD_NAME]
-        val = "#{non_sort_portion} #{val}" if non_sort_portion && !sortable
-      end
-
-      val
     end
 
     def number_of_children
--- a/app/graphql/types/digital_object_interface.rb
+++ b/app/graphql/types/digital_object_interface.rb
@@ -57,7 +57,7 @@ module Types
       end
     end
 
-    def self.title_for(object, indexable = false)
+    def self.title_for(object, sortable = false)
       val = '[No Title]'
 
       title_field_group = object.dynamic_field_data[TITLE_DYNAMIC_FIELD_GROUP_NAME]
--- a/app/graphql/types/digital_object_interface.rb
+++ b/app/graphql/types/digital_object_interface.rb
@@ -64,7 +64,7 @@ module Types
       if title_field_group.present? && (title_field = title_field_group[0]).present?
         val = title_field[TITLE_SORT_PORTION_DYNAMIC_FIELD_NAME]
         non_sort_portion = title_field[TITLE_NON_SORT_PORTION_DYNAMIC_FIELD_NAME]
-        val = "#{non_sort_portion} #{val}" if non_sort_portion && !indexable
+        val = "#{non_sort_portion} #{val}" if non_sort_portion && !sortable
       end
 
       val
--- a/app/graphql/types/digital_object_interface.rb
+++ b/app/graphql/types/digital_object_interface.rb
@@ -57,14 +57,14 @@ module Types
       end
     end
 
-    def self.title_for(object)
+    def self.title_for(object, indexable = false)
       val = '[No Title]'
 
       title_field_group = object.dynamic_field_data[TITLE_DYNAMIC_FIELD_GROUP_NAME]
       if title_field_group.present? && (title_field = title_field_group[0]).present?
         val = title_field[TITLE_SORT_PORTION_DYNAMIC_FIELD_NAME]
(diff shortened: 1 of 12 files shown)


barmintor:
I've got it up on dev, but there are some hitches with the rights metadata - generally the field configs are out of synch there. Might be good to clear it out and migrate the prod data to it?

lib/hyacinth/adapters/digital_object_search_adapter/solr/document_adapter.rb
```diff
@@ -0,0 +1,82 @@
...
+    module DigitalObjectSearchAdapter
+      class Solr::DocumentAdapter
+        def solr_document_for(digital_object)
+          solr_document = merge_core_fields_for(digital_object)
+          merge_dynamic_fields_for!(digital_object, solr_document)
+          merge_rights_fields_for!(digital_object, solr_document)
+          solr_document
+        end
+
+        def merge_core_fields_for(digital_object, solr_document = {})
```
elohanlon: Are we planning on ever calling this method with a second param that isn't `{}`? If not, is it worth getting rid of the second parameter and renaming it to `core_fields_for`?

lib/hyacinth/adapters/digital_object_search_adapter/solr/document_adapter.rb
```diff
@@ -0,0 +1,82 @@
...
+            'identifier_ssim' => digital_object.identifiers.to_a,
+            'title_ssi' => indexable_title,
+            'primary_project_ssi' => digital_object.primary_project&.string_key,
+            'projects_ssim' => project_keys_for(digital_object)
+          )
+          add_keywords(indexable_title, solr_document)
+          solr_document
+        end
+
+        def merge_dynamic_fields_for(digital_object, solr_document = {})
```
elohanlon: Same question here about this `merge_` method.

lib/hyacinth/adapters/digital_object_search_adapter/solr/document_adapter.rb
```diff
@@ -0,0 +1,82 @@
...
+          # add flag for content at all if field is not facetable
+          collection_values = digital_object.dynamic_field_data.fetch('collection', [])
+          collection_values.each do |collection_value|
+            (solr_document['collection_ssim'] ||= []) << collection_value.dig('term', 'pref_label')
+            add_keywords(collection_value.dig('term', 'pref_label'), solr_document)
+          end
+          solr_document
+        end
+
+        def merge_rights_fields_for(digital_object, solr_document = {})
```
elohanlon: And same question here about this `merge_` method.

lib/hyacinth/adapters/digital_object_search_adapter/solr/document_adapter.rb
barmintor: My goals here were:
1. Principle of least surprise about hash-like APIs
2. Reusability/thread-safety of more stateless wrapping class
3. Reusability across solr "profiles" (item/asset/site)
4. Ease of testing

So this seemed like the most straight-forward way to accomplish those - though I might be influenced by prior art in the Hydra world.

app/graphql/types/digital_object_interface.rb
```diff
@@ -57,14 +57,14 @@ def resources
       end
     end
 
-    def self.title_for(object)
+    def self.title_for(object, indexable = false)
```
elohanlon: Do you mind if we're a little more literal with the second param, calling it something like `sortable`?  I found it slightly confusing at first because there isn't a reason why the full title _isn't_ indexable -- we just wouldn't want to index it in a sortable title field.

barmintor: That seems like a good idea. Can you `suggest a change`?

elohanlon: Will do!

elohanlon:
```suggestion
    def self.title_for(object, sortable = false)
```

app/graphql/types/digital_object_interface.rb
```diff
@@ -57,14 +57,14 @@ def resources
...
       val = '[No Title]'
 
       title_field_group = object.dynamic_field_data[TITLE_DYNAMIC_FIELD_GROUP_NAME]
       if title_field_group.present? && (title_field = title_field_group[0]).present?
         val = title_field[TITLE_SORT_PORTION_DYNAMIC_FIELD_NAME]
         non_sort_portion = title_field[TITLE_NON_SORT_PORTION_DYNAMIC_FIELD_NAME]
-        val = "#{non_sort_portion} #{val}" if non_sort_portion
+        val = "#{non_sort_portion} #{val}" if non_sort_portion && !indexable
```
elohanlon:
```suggestion
        val = "#{non_sort_portion} #{val}" if non_sort_portion && !sortable
```

app/graphql/types/digital_object_interface.rb
cgalarza: It really feels like this method should be moved to DigitalObject.

barmintor: I think the only reason it hasn't is that we were talking about making it an actual core field

cgalarza: This might be a good time to move it because when that change happens the implementation of this method can be updated and graphql or the indexing code wouldn't know the difference.

barmintor:
```suggestion
    def self.title_for(object, sortable = false)
```

lib/hyacinth/adapters/digital_object_search_adapter/solr/document_adapter.rb
elohanlon: 👍

lib/hyacinth/adapters/digital_object_search_adapter/solr/document_adapter.rb
```diff
@@ -0,0 +1,82 @@
...
+        def merge_core_fields_for(digital_object, solr_document = {})
+          merge_core_fields_for!(digital_object, solr_document.dup)
+        end
+
+        def merge_core_fields_for!(digital_object, solr_document = {})
+          indexable_title = ::Types::DigitalObjectInterface.title_for(digital_object, true)
+          solr_document.merge!(
+            'id' => digital_object.uid,
+            'digital_object_type_ssi' => digital_object.digital_object_type,
+            'uid_ssi' => digital_object.uid,
```
elohanlon: Maybe not needed, since we have the value in `id`? (I realize this is a ported thing and not a new thing.)

elohanlon: Also, there's no test checking for its presence.

lib/hyacinth/adapters/digital_object_search_adapter/solr/document_adapter.rb
```diff
@@ -0,0 +1,82 @@
+# frozen_string_literal: true
+
+module Hyacinth
+  module Adapters
+    module DigitalObjectSearchAdapter
+      class Solr::DocumentAdapter
```
cgalarza: This might be a weird and picky comment, but it feels like we are overusing the word "adapter". Because when i first saw the class name DocumentAdapter I thought it was at the same level with `DigitalObjectSearchAdapter` but its actually more of a helper class. Could we call it something more descriptive like Solr::DocumentCreation or something like that.

PR #250, opened 2020-03-06
Refactoring downloads controller

cgalarza:
This is in preparation for adding a batch_imports download endpoint. For which we have to provide a download to the original file and a file without the successful rows.

### DIFF
--- a/app/controllers/api/v1/downloads_controller.rb
+++ b/app/controllers/api/v1/downloads_controller.rb
@@ -12,7 +12,7 @@ module Api
       # GET /downloads/digital_object/:id/:resource_name
       def digital_object
         id = digital_object_params[:id]
-        subresource = digital_object_params[:resource_name]
+        resource_name = digital_object_params[:resource_name]
 
         digital_object = DigitalObject::Base.find(id)
         authorize! :read, digital_object
--- a/app/controllers/api/v1/downloads_controller.rb
+++ b/app/controllers/api/v1/downloads_controller.rb
@@ -16,7 +16,7 @@ module Api
 
         digital_object = DigitalObject::Base.find(id)
         authorize! :read, digital_object
-        resource = digital_object.resources[subresource]
+        resource = digital_object.resources[resource_name]
         raise Hyacinth::Exceptions::NotFound if resource.nil? || resource.location.blank?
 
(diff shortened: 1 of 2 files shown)


cgalarza:
@elohanlon I tried making the `download` method take all the parameters needed, but the total number of parameters was 6 and rubocop was not happy. 

Another thing we could do at the end of each action method is call `set_download_header` and `download` separately.

elohanlon:
@cgalarza This seems like a case where the benefits of overriding the rubocop complaint (via `# rubocop:disable` comment) may outweigh the downsides of having to set a bunch of instance variables.  The instance variable method is a little more error prone. Would you be open to doing that?

elohanlon:
But also, calling `set_download_header` and `download` separately is a good idea.  If that keeps rubocop happy, it's definitely still better than relying on the setting of a bunch of instance variables.

cgalarza:
Okay, sounds good, i'll go for calling the two methods separately.

cgalarza:
Just pushed an update.

app/controllers/api/v1/downloads_controller.rb
```diff
@@ -9,79 +9,64 @@ class DownloadsController < ApplicationApiController
...
-          while (chunk = io.read(DOWNLOAD_BUFFER_BYTE_SIZE))
-            response.stream.write(chunk)
-          end
-        ensure
-          response.stream.close
-        end
+      # GET /downloads/digital_object/:id/:resource_name
+      def digital_object
+        id = digital_object_params[:id]
+        subresource = digital_object_params[:resource_name]
```
elohanlon:
```suggestion
        resource_name = digital_object_params[:resource_name]
```

app/controllers/api/v1/downloads_controller.rb
```diff
@@ -9,79 +9,64 @@ class DownloadsController < ApplicationApiController
...
+
+        digital_object = DigitalObject::Base.find(id)
+        authorize! :read, digital_object
+        resource = digital_object.resources[subresource]
```
elohanlon:
```suggestion
        resource = digital_object.resources[resource_name]
```

elohanlon:
Looks great!

PR #251, opened 2020-03-06
Generalizing the uploads controller; Generalizing file upload UI

cgalarza:


cgalarza:
Deploying to dev right now to test asset uploads.

PR #252, opened 2020-03-10
Batch Import create/update/delete mutation; Batch Import UI

cgalarza:
closes HYACINTH-513, HYACINTH-516, HYACINTH-528

### DIFF
--- a/app/controllers/api/v1/downloads_controller.rb
+++ b/app/controllers/api/v1/downloads_controller.rb
@@ -44,6 +44,26 @@ module Api
         download(storage, file_location)
       end
 
+      # GET /downloads/batch_import/:id(/:option)
+      def batch_import
+        batch_import = load_and_authorize_batch_import!
+
+        if batch_import_params[:option] == 'without_successful_imports'
+          send_data(
+            batch_import.csv_without_successful_imports,
+            type: 'text/csv',
+            filename: "without-successful-rows-#{batch_import.original_filename}"
+          )
+        else
+          storage = Hyacinth::Config.batch_import_storage
+          size = storage.size(file_location).to_s
+          file_name = batch_import.original_filename || "import-#{batch_import.id}.csv"
+
+          set_download_headers(batch_import.updated_at, 'text/csv', size, file_name)
+          download(storage, batch_import.file_location)
+        end
+      end
+
       private
 
         def digital_object_params
@@ -52,6 +72,10 @@ module Api
 
         def batch_export_params
           params.permit(:id)
+        end
+
+        def batch_import_params
+          params.permit(:id, :option)
         end
 
         # File download.
@@ -80,6 +104,14 @@ module Api
           safe_label = label.gsub(' ', '%20').gsub(',', '%2C')
           # The two adjacent single quotes in the line below look weird, but are correct:
           (attachment ? 'attachment; ' : 'inline') + "; filename*=utf-8''" + safe_label
+        end
+
+        def load_and_authorize_batch_import!
+          batch_import = BatchImport.find(batch_import_params[:id])
+          authorize! :read, batch_import
+          file_location = batch_import.file_location
+          raise Hyacinth::Exceptions::NotFound if file_location.blank?
+          batch_import
         end
     end
   end
--- a/app/controllers/api/v1/downloads_controller.rb
+++ b/app/controllers/api/v1/downloads_controller.rb
@@ -49,14 +49,14 @@ module Api
         batch_import = load_and_authorize_batch_import!
 
         if batch_import_params[:option] == 'without_successful_imports'
+          response.headers["Last-Modified"] = batch_import.updated_at.httpdate
           send_data(
             batch_import.csv_without_successful_imports,
-            type: 'text/csv',
-            filename: "without-successful-rows-#{batch_import.original_filename}"
+            type: 'text/csv', filename: "without-successful-imports-#{batch_import.original_filename}"
           )
         else
           storage = Hyacinth::Config.batch_import_storage
-          size = storage.size(file_location).to_s
+          size = storage.size(batch_import.file_location).to_s
           file_name = batch_import.original_filename || "import-#{batch_import.id}.csv"
 
           set_download_headers(batch_import.updated_at, 'text/csv', size, file_name)
--- a/app/javascript/hyacinth_v1/components/batch_imports/BatchImportIndex.js
+++ b/app/javascript/hyacinth_v1/components/batch_imports/BatchImportIndex.js
@@ -1,0 +1,114 @@
+import React, { useState } from 'react';
+import { Card, Button } from 'react-bootstrap';
+import * as moment from 'moment';
+import { useQuery, useMutation } from '@apollo/react-hooks';
+import { startCase } from 'lodash';
+
+import { Can } from '@hyacinth_v1/utils/abilityContext';
+import ContextualNavbar from '../shared/ContextualNavbar';
+import GraphQLErrors from '../shared/GraphQLErrors';
+import PaginationBar from '../shared/PaginationBar';
+import { batchImportsQuery, updateBatchImportMutation, deleteBatchImportMutation } from '../../graphql/batchImports';
+
+function BatchImportIndex() {
+  const limit = 10;
+  const [offset, setOffset] = useState(0);
+
+  const {
+    loading, error, data, refetch,
+  } = useQuery(batchImportsQuery, { variables: { limit, offset } });
+
+  const [updateBatchImport, { error: updateError }] = useMutation(updateBatchImportMutation);
+  const [deleteBatchImport, { error: deleteError }] = useMutation(deleteBatchImportMutation);
+
+  if (loading) return (<></>);
+  if (error) return (<GraphQLErrors errors={error} />);
+
+  const onPageNumberClick = (newOffset) => {
+    setOffset(newOffset);
+    refetch();
+  };
+
+  const onCancel = (id) => {
+    const variables = { input: { id, cancelled: true } };
+    updateBatchImport({ variables }).then(refetch);
+  };
+
+  const onDelete = (id) => {
+    const variables = { input: { id } };
+    deleteBatchImport({ variables }).then(refetch);
+  };
+
+  const { batchImports: { nodes: batchImports, totalCount: totalBatchImports } } = data;
+
+  return (
+    <>
+      <ContextualNavbar
+        title="Batch Imports"
+        rightHandLinks={[
+          { link: 'batch_imports/new', label: 'New Batch Import' },
+        ]}
+      />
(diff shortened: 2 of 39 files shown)


app/controllers/api/v1/downloads_controller.rb
```diff
@@ -44,6 +44,26 @@ def batch_export
         download(storage, file_location)
       end
 
+      # GET /downloads/batch_import/:id(/:option)
+      def batch_import
```
elohanlon: @cgalarza Could you please add a test for this in downloads_spec.rb?

cgalarza: Sure!

app/controllers/api/v1/downloads_controller.rb
```diff
@@ -44,6 +44,26 @@ def batch_export
...
+        batch_import = load_and_authorize_batch_import!
+
+        if batch_import_params[:option] == 'without_successful_imports'
+          send_data(
```
elohanlon: Most of the time, this `send_data` call will be sending a relatively small file, but sometimes it might be a much larger CSV with a lot of columns.  It would be good to refactor this to stream like the other downloads methods, using DOWNLOAD_BUFFER_BYTE_SIZE to limit the amount of data in memory at any given moment.  This is a sort of tricky situation though, given how we need to read and filter the csv without successful rows before providing it as a download.

Here here's one possible option:
- Convert `csv_without_successful_imports` to a method that does a buffered low-memory-usage write to a temporary file on disk, and have it yield a block with a reference to that file, and automatically clean up the temporary file afterward in an `ensure`.
- In the block that you run, read the temporary file and stream it via the same `response.stream.write(chunk)` strategy used in the `download` method.  Ideally it would be nice to re-use more of the logic in the download method, and maybe we can with some additional refactoring, but the above steps would be a first-pass way of keeping the memory usage low for this download case.

app/controllers/api/v1/downloads_controller.rb
```diff
@@ -44,6 +44,26 @@ def batch_export
...
+            batch_import.csv_without_successful_imports,
+            type: 'text/csv',
+            filename: "without-successful-rows-#{batch_import.original_filename}"
```
elohanlon: We call it "without-successful-rows" in Hyacinth 2, but I don't mind if we change it to "without-successful-imports" in Hyacinth 3 if you don't.

app/javascript/hyacinth_v1/components/batch_imports/BatchImportIndex.js
```diff
@@ -0,0 +1,114 @@
...
+      />
+      <GraphQLErrors errors={updateError || deleteError} />
+      { batchImports.length === 0 && (<p className="text-center">No Batch Imports have been created.</p>) }
+      {
+        batchImports.map(batchImport => (
+          <Card key={batchImport.id} className="mb-3">
+            <Card.Header>
+              {`Import ID: ${batchImport.id}`}
+              <div className="float-right">
+                {moment(batchImport.createdAt).format('MMMM Do YYYY, h:mm:ss a')}
```
elohanlon: Isn't `moment` great?  I really like this module.

app/javascript/hyacinth_v1/components/batch_imports/BatchImportShow.js
```diff
@@ -0,0 +1,169 @@
...
+      stopPolling();
+    }
+    setPolling(prevPolling => !prevPolling);
+  };
+
+  useEffect(() => {
+    if (!data) { return; }
+
+    const { batchImport: { status } } = data;
+    if (polling && status !== 'in_progress' && status !== 'pending') {
```
elohanlon: This is great!

app/javascript/hyacinth_v1/components/batch_imports/BatchImports.js
```diff
@@ -0,0 +1,27 @@
...
+import DigitalObjectImports from './digital_object_imports/DigitalObjectImports';
+
+function BatchImports() {
+  return (
+    <Switch>
+      <Route exact path="/batch_imports" component={BatchImportIndex} />
+      <Route exact path="/batch_imports/new" component={BatchImportNew} />
+      <Route path="/batch_imports/:id/digital_object_imports" component={DigitalObjectImports} />
+
+      {/* need to check  permissions somehow */}
```
elohanlon:
```suggestion
      {/* TODO: need to check  permissions somehow */}
```
I just added "TODO" to make this easier to find later with other TODO notes.  Also, if this isn't ticketed yet, it would be good to create a ticket for it.

elohanlon: Is the problem here that we don't have easy access to the current user's id in this context?  Maybe we want to make it more globally accessible across the client-side code?

This same type of permission check may apply to DigitalObjectImports.js routes too.

app/javascript/hyacinth_v1/components/shared/ReadableDate.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import * as moment from 'moment';
+
+function ReadableDate({ date, format }) {
```
elohanlon: Great component idea!

app/javascript/hyacinth_v1/components/batch_imports/BatchImportIndex.js
elohanlon: I just saw got up to your ReadableDate component in the PR, so I'll add this suggestion:
```suggestion
                <ReadableDate date={batchImport.createdAt} />
```
(plus the import at the top of the file)

app/javascript/hyacinth_v1/components/batch_imports/BatchImportIndex.js
```diff
@@ -0,0 +1,114 @@
+import React, { useState } from 'react';
+import { Card, Button } from 'react-bootstrap';
+import * as moment from 'moment';
+import { useQuery, useMutation } from '@apollo/react-hooks';
+import { startCase } from 'lodash';
+
+import { Can } from '@hyacinth_v1/utils/abilityContext';
+import ContextualNavbar from '../shared/ContextualNavbar';
+import GraphQLErrors from '../shared/GraphQLErrors';
+import PaginationBar from '../shared/PaginationBar';
```
elohanlon:
```suggestion
import PaginationBar from '../shared/PaginationBar';
import ReadableDate from '../shared/ReadableDate';
```

spec/requests/graphql/mutations/batch_import/delete_batch_import_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+      end
+
+      it 'returns errors' do
+        expect(response.body).to be_json_eql(%(
+         "Cannot destroy batch import while imports are in_progress or pending"
+        )).at_path('errors/0/message')
+      end
+    end
+
+    context "when deleting an batch import that doesn't exist" do
```
elohanlon:
```suggestion
    context "when deleting a batch import that doesn't exist" do
```

spec/requests/graphql/mutations/batch_import/delete_batch_import_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    let(:variables) { { input: { id: id } } }
+    let(:request) { graphql query, variables }
+  end
+
+  context 'when user is logged in' do
+    before do
+      sign_in user
+    end
+
+    context 'when deleting an batch import that exists' do
```
elohanlon:
```suggestion
    context 'when deleting a batch import that exists' do
```

app/controllers/api/v1/downloads_controller.rb
cgalarza: We currently use `send_data` in prod and we haven't seen a problem yet (that I know of). Refactoring this can be delayed. I think we should just ticket this and work on it later when we have time.

app/controllers/api/v1/downloads_controller.rb
cgalarza: Are you talking about the filename prefix?

app/javascript/hyacinth_v1/components/batch_imports/BatchImportIndex.js
cgalarza: Whoops, I'll update this to use the component.

app/javascript/hyacinth_v1/components/batch_imports/BatchImports.js
cgalarza: I think that we are _not_ going to be able to successfully check permissions for BatchImports before we render the page. In order to successfully check permissions I need to be able to retrieve the actual object (BatchImport) itself in order to check what user it belongs to. In order to retrieve the object the user needs to already have permission to the object. At some point, I thought we would be able to check permissions via the user id, but then I realized that we don't have that information to do that if the user doesn't have access to the import. I think we should just remove this TODOs and depend on the graphql errors that will be raised and displayed if a user doesn't have access.

app/controllers/api/v1/downloads_controller.rb
cgalarza: Testing this definitely caught at least one problem. Thanks for the reminder.

app/controllers/api/v1/downloads_controller.rb
elohanlon: Sounds good! Ticketed!

app/controllers/api/v1/downloads_controller.rb
elohanlon: Yeah exactly.  We could do: `"without-successful-imports-#{batch_import.original_filename}"`

app/controllers/api/v1/downloads_controller.rb
elohanlon: 👍

app/javascript/hyacinth_v1/components/batch_imports/BatchImports.js
elohanlon: Sounds good to me!

app/controllers/api/v1/downloads_controller.rb
cgalarza: ahhh! I have no strong feelings. Would you prefer "without-successful-imports-#{batch_import.original_filename}"?

elohanlon: I have a slight preference for it just because of the future scenario of creating batch imports that aren't from CSVs (like from voyager-to-hyacinth syncs).

cgalarza: Okay, I'll make the change.

elohanlon: Thanks!

PR #253, opened 2020-04-03
Right display/form using dynamic field definitions

cgalarza:
- Adding rake task that batch loads rights fields as dynamic fields
- Adding `metadata_form` flag to DynamicFieldCategory to describe what form the fields should be displayed in.
- Creating rights read display to show selected values
- Updating the rights form to pull field configuration for the rights dynamic fields
- No longer using GraphQL types to represent the rights data structure. Rights is now a json type. By removing this we removed a lot of code on the graphql side
- Adding basic validation of rights structure to the digital object model
- Removed API endpoint for rights

Note: When we deploy this PR to dev or prod, we need to run the rake task that creates the fields.


elohanlon:
Looks great!  And I really like all the deleted code!

PR #257, opened 2020-04-20
Validating Dynamic Field Data

cgalarza:
Checking against the defined dynamic fields to ensure that descriptive_metadata and rights are properly structured and that the values for each field are valid.

Added a class to easily load dynamic field categories, groups and fields.

Rehydrating and dehydrating term data from descriptive_metadata and rights when its read-in from storage.

Note: This PR does not cover creating and minting terms. That will be a separate PR.


app/validators/digital_object/dynamic_fields_validator.rb
```diff
@@ -0,0 +1,97 @@
...
+      value.is_a?(String) ? false : ['must be a string']
+    end
+
+    def errors_for_textarea_field(configuration, value)
+      errors_for_string_field(configuration, value)
+    end
+
+    # TODO: Add more specific checks for this field.
+    def errors_for_date_field(configuration, value)
+      errors_for_string_field(configuration, value)
```
cgalarza, line 66: Should we be validating the dates the come in? if so, what should the format be? I assume (YYYY-MM-DD).

app/validators/digital_object/dynamic_fields_validator.rb
```diff
@@ -0,0 +1,97 @@
...
+    end
+
+    # TODO: Add more specific checks for this field. Select value should be one of the configured values.
+    def errors_for_select_field(configuration, value)
+      errors_for_string_field(configuration, value)
```
cgalarza, line 71: Should we be validating the values in select fields? I worry that its going to lead to a lot of migration issues, but if we want to enforce select values now would be the best time to do it.

elohanlon: Yeah, I think validating select values now is a good idea.  Allowing bad values in during migration will probably lead to worse headaches later on.

app/validators/digital_object/dynamic_fields_validator.rb
elohanlon: YYYY-MM-DD is correct.  We're basing the format on the date portion of the ISO 8601 standard.

barmintor: Test

cgalarza: I am following the documentation here:

https://www.w3.org/TR/NOTE-datetime 

and supporting year (YYYY), year and month (YYYY-MM) and complete date (YYYY-MM-DD). 

Should I also be validating that the date is a real date?

The regex I plan on using is `^\d{4}(-(0[1-9]|1[0-2])(-(0[1-9]|[1,2][0-9]|3[0,1]))?)?$` which minimally checks that the date and month could be valid.

app/validators/digital_object/dynamic_fields_validator.rb
cgalarza: Okay, will do.

app/validators/digital_object/dynamic_fields_validator.rb
elohanlon: `Date.parse(date_string)` will raise an ArgumentError if you pass an invalid `date_string` value like `'2020-02-41'`.  That's probably an inexpensive thing that would be good to throw in as an additional step after the regex, catching that error.  It'll help to defend against user typos (which will hopefully be rare, so we won't have to worry about the expense of catching many errors).  I tested it and it supports negative years for BCE, so that's good too.

cgalarza: The regex doesn't support negative year, should it?

elohanlon: It should.  We have a few BCE items in our digital collections.

cgalarza: Done.

app/validators/digital_object/dynamic_fields_validator.rb
```diff
@@ -0,0 +1,113 @@
...
+
+      errors.empty? ? false : errors
+    end
+
+    # Returns nil if vocabulary is not represented in hash
+    def custom_fields_for(vocabulary)
+      vocabulary_to_custom_fields_map.fetch(vocabulary, nil)
+    end
+
+    def vocabulary_to_custom_fields_map
```
cgalarza, line 110: Ideally, we could read in this data from cache. Is hyacinth configured properly to use caching? I know we've gone back and forth a few times about what caching store to use.

elohanlon: I just checked and it's not configured yet for server environments.  Our `config.cache_store` config option is commented out in `deployed.rb`.

I think our best bet is to set up redis as a cache that works across passenger processes in the deployed environment.  That way when one thread invalidates the cache, it's invalidated for all processes.

In devevelopment we have `config.cache_store = :memory_store`, but that's a per-process cache, so we might want to use `:file_store` instead (which would cover cases like my own, when I'm sometimes running the local puma server with multiple workers so one request doesn't block another).

app/models/digital_object/base.rb
```diff
@@ -58,9 +58,9 @@ class Base
     # Identifiers
     metadata_attribute :identifiers, Hyacinth::DigitalObject::TypeDef::JsonSerializableSet.new.default(-> { Set.new })
     # Descriptive Metadata
-    metadata_attribute :descriptive_metadata, Hyacinth::DigitalObject::TypeDef::JsonSerializableHash.new.default(-> { Hash.new })
+    metadata_attribute :descriptive_metadata, Hyacinth::DigitalObject::TypeDef::DynamicFieldData.new(:descriptive_metadata).default(-> { Hash.new })
```
barmintor, line 61: 👍

lib/hyacinth/dynamic_fields_map.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+module Hyacinth
+  module DynamicFieldsMap
+    # Generates map of dynamic fields for the given metadata form.
+    #
+    # @param [String|Array<String>] for_metadata_form limits map to fields related forms
+    def self.generate(for_metadata_form)
```
barmintor, line 8: would it be easier to use the splat operator here?

barmintor:
I didn't follow up on the test issue - looks like maybe a hash is being filtered if values are `blank?` ?

app/models/digital_object/base.rb
elohanlon: 👍

app/validators/digital_object/dynamic_fields_validator.rb
```diff
@@ -0,0 +1,113 @@
+# frozen_string_literal: true
+
+# This class provided the logic for validating dynamic fields against a map generated by DynamicFieldMap.
+class DigitalObject::DynamicFieldsValidator < ActiveModel::EachValidator
+  CORE_TERM_FIELDS = ['uri', 'alt_labels', 'pref_label', 'term_type', 'authority', 'term_type'].freeze
```
elohanlon, line 5: How would you feel about putting this constant in the Term class?  It's not wrong to have it here, but it feels to me like something that should be defined in that context -- sort of "owned by" the Term class.  This is just a slight preference, so I'm find keeping it here if you'd prefer having it here.

app/validators/digital_object/dynamic_fields_validator.rb
```diff
@@ -0,0 +1,113 @@
...
+
+  private
+
+    def generate_errors(digital_object, attribute, value, map)
+      errors_for(map, value, attribute).each { |a| digital_object.errors.add(a[0], a[1]) }
+    end
+
+    # Returns errors that should be recorded for any of the fields in the data given.
+    # Errors include, invalid fields appearing in the data given, field values containing the wrong type of data.
```
elohanlon, line 14:
```suggestion
    # Errors include: invalid fields appearing in the data given, field values containing the wrong type of data.
```

app/validators/digital_object/dynamic_fields_validator.rb
```diff
@@ -0,0 +1,113 @@
...
+    end
+
+    def errors_for_date_field(_configuration, value)
+      return ['must be a string'] unless value.is_a?(String)
+      return ['must be in YYYY-MM-DD format'] unless value.match?(/^-?\d{4}(-(0[1-9]|1[0-2])(-(0[1-9]|[1,2][0-9]|3[0,1]))?)?/)
+
+      begin
+        Date.parse(value)
+      rescue
+        return ['invalid date']
```
elohanlon, line 71:
```suggestion
        return ['is an invalid date']
```
Suggesting a phrasing change to match the style of some other error messages.

lib/hyacinth/adapters/term_search_adapter/solr.rb
```diff
@@ -35,7 +35,7 @@ def find(vocabulary, uri)
         end
 
         # Solr query. Returns solr json.
-        def search
+        def search(post: false)
```
elohanlon, line 38: I wonder if it would be better to just post 100% of the time if we know we'll need to sometimes?  I don't know if there's an advantage to doing `get` sometimes and `post` other times.

lib/hyacinth/digital_object/type_def/dynamic_field_data.rb
```diff
@@ -0,0 +1,117 @@
...
+          @type = type.to_sym
+        end
+
+        def from_serialized_form_impl(json_var)
+          return nil if json_var.nil?
+          raise ArgumentError, "Expected hash, but got: #{json_var.class}" unless json_var.is_a?(Hash)
+
+          terms_map = extract_terms(field_map, json_var)
+
+          search_query = terms_map.map { |vocab, terms|
```
elohanlon, line 25: Maybe it would be worth moving the batch search functionality into the term search adapter itself (in a new method like `batch_search` that takes an argument that's a hash like {names: ['uri1', 'uri2'], collections: ['uri3', 'uri4']), since it'll be useful to look up a bunch of terms in a vocabulary for both validation purposes AND, later, new term registration or existing-term-enhancement purposes.

lib/hyacinth/adapters/term_search_adapter/solr.rb
cgalarza: I was wondering about this actually. That way we don't have to worry about specifying it in certain places.

lib/hyacinth/digital_object/type_def/dynamic_field_data.rb
```diff
@@ -0,0 +1,117 @@
...
+            "(vocabulary:\"#{vocab}\" AND uri:(#{uris}))"
+          }.join(' OR ')
+
+          search_results = Hyacinth::Config.term_search_adapter.search(post: true) do |solr_params|
+            solr_params.q(search_query, escape: false)
+          end
+
+          # Retrieve all custom fields for vocabularies so we know which fields should be add to the hash.
+          # This ensure that there's always a value for the all the custom fields defined and does not add
+          # any custom fields that have been removed from the Vocabulary since indexing.
```
elohanlon, line 36: Does this comment mean that we're expecting that `Hyacinth::Config.term_search_adapter.search` could theoretically return search results that include no-longer-existing custom fields or are missing existing custom fields?  If so, that sounds like a job that the search adapter should be doing rather than handling it only in the typedef, so we benefit from this inconsistency-handling app-wide.  (Unless I'm misunderstanding?)

lib/hyacinth/digital_object/type_def/dynamic_field_data.rb
```diff
@@ -0,0 +1,117 @@
...
+
+              term_hash.slice!('uri')
+              term_hash.merge!(full_term)
+            end
+          end
+
+          json_var
+        end
+
+        def to_serialized_form_impl(json_var)
```
elohanlon, line 68: I love how straightforward and understandable this method is.  Nice implementation.

lib/hyacinth/digital_object/type_def/dynamic_field_data.rb
```diff
@@ -0,0 +1,117 @@
...
+          json_var = json_var.deep_dup
+
+          # Dehydrate URI terms. Remove all other fields in terms hash except for 'uri'
+          extract_terms(field_map, json_var).values.sum([]).each { |t| t.slice!('uri') }
+
+          json_var
+        end
+
+        def field_map
+          Hyacinth::DynamicFieldsMap.generate(TYPE_TO_FORM_TYPE[@type])
```
elohanlon, line 82: Once we enforce uniqueness among top level DynamicFieldGroup string keys AND start doing more data structure caching, it may make sense to just have one fully cached all-dynamic-fields map that's invalidated whenever any dynamic field is modified -- and that'll simplify some code in this class -- but until then I can see the benefits of retrieving smaller per-form field maps, so I'm good with this. 👍

lib/hyacinth/dynamic_fields_loader.rb
```diff
@@ -2,7 +2,57 @@
...
+    #                   ]
+    #                 }
+    #               }
+    #             ]
+    #           }
+    #         ]
+    #       }
+    # @param [opts] options to use when generating fields
+    # @option opts [Boolean] :load_vocabularies option to load any vocabularies fields are dependent on
+    def self.load_fields!(field_definitions, opts = {})
```
elohanlon, line 33: This also seems potentially useful for loading fields between Hyacinth instances (i.e. syncing prod fields to a new dev instance).

lib/hyacinth/dynamic_fields_loader.rb
```diff
@@ -2,7 +2,57 @@
...
+        # Load Item and Asset Rights Fields
+        field_definitions[:dynamic_field_categories].each do |category_config|
+          display_label = category_config.delete(:display_label)
+          child_dynamic_field_groups = category_config.delete(:dynamic_field_groups)
+
+          category = DynamicFieldCategory.find_or_create_by!(display_label: display_label) do |c|
+            c.assign_attributes(category_config)
+          end
+
+          create_dynamic_field_groups!(category, child_dynamic_field_groups, opts)
```
elohanlon, line 45: It doesn't have to be for this pull request, but how would you feel if we renamed this module `DynamicFieldsCreator` instead of `DynamicFieldsLoader` (and renamed load_* to create_*)?  Reason is, when I hear "load" it makes me think "retrieving and loading something into memory" rather than saving something to permanent storage.

lib/tasks/hyacinth/setup.rake
```diff
@@ -74,70 +74,23 @@ namespace :hyacinth do
...
-          filter_label: '',
-          controlled_vocabulary: '',
-          select_options: '{}',
-          additional_data_json: '{}'
-        }
-      ]
+      field_definitions = {
+        dynamic_field_categories: [{
+          display_label: "Descriptive Metadata",
+          dynamic_field_groups: [
```
elohanlon, line 80: I added these to my pull request too, haha.

spec/hyacinth/digital_object/type_def/dynamic_field_data_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+            'other_data' => 'something'
+          }
+        }
+      ]
+    }
+  end
+
+  # Load field definitions and create terms.
+  before do
+    Hyacinth::DynamicFieldsLoader.load_fields!(field_definitions)
```
elohanlon, line 75: Is it worth using `load_vocabularies: true` here to cover the vocabulary creation cases below, and then you can just change the `name_vocab = Vocabulary.create(...` line to `name_vocab = Vocabulary.find_by(string_key: 'name')`?

elohanlon: Oh, I see.  You're also making use of the `genre_vocab` and `name_role_vocab` references later.  I guess it doesn't save you much if you have to call find_by on those too.

spec/hyacinth/digital_object/type_def/dynamic_field_data_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+    name_role_vocab = Vocabulary.create(string_key: 'name_role', label: 'Name Role')
+    genre_vocab = Vocabulary.create(string_key: 'genre', label: 'Genre')
+
+    Term.create!(pref_label: 'Person, Random', term_type: 'temporary', vocabulary: name_vocab, custom_fields: { uni: 'abc123' })
+    Term.create!(uri: 'http://id.loc.gov/vocabulary/relators/aut', pref_label: 'Author', alt_labels: ['writer'], authority: 'marcrelator', term_type: 'external', vocabulary: name_role_vocab)
+    Term.create!(uri: 'http://vocab.getty.edu/aat/300048715', pref_label: 'articles', authority: 'aat', term_type: 'external', vocabulary: genre_vocab)
+  end
+
+  describe '#from_serialized_form_impl' do
+    let(:expected_serialization) do
```
elohanlon, line 89: For completeness, it would be good to also include some non-term fields in this test to make sure they're deserialized as expected (and not modified in any way).

spec/hyacinth/digital_object/type_def/dynamic_field_data_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+      }
+    end
+
+    it 'adds term data to hash' do
+      expect(type_def.from_serialized_form_impl(descriptive_metadata)).to include(expected_serialization)
+    end
+  end
+
+  describe '#to_serialized_form_impl' do
+    let(:expected_serialization) do
```
elohanlon, line 134: It would also be good to also include at least one non-controlled-term field in this test to make sure it's serialized as expected.

spec/hyacinth/dynamic_fields_map_spec.rb
```diff
@@ -0,0 +1,169 @@
...
+      subject(:map) { described_class.generate('item_rights') }
+
+      it 'returns empty map' do
+        expect(map).to eql({})
+      end
+    end
+
+    context 'when generating map for asset_rights form' do
+      subject(:map) { described_class.generate('asset_rights') }
+      before { Hyacinth::DynamicFieldsLoader.load_rights_fields! }
```
elohanlon, line 17: This `before` step definitely does test the `Hyacinth::DynamicFieldsMap`'s `generate` method, but I think it might be better to write a test with mock data so that we don't need to update this test if/when the hard-coded set of rights fields change.  Swapping it with `load_fields!(...)` and some mock json should do it, and would better isolate the success of this test from the success of the `load_rights_fields!` implementation.

spec/models/concerns/digital_object_concerns/validations/descriptive_fields_validator_spec.rb
```diff
@@ -0,0 +1,298 @@
...
+
+    it 'returns errors' do
+      expect(item.valid?).to be false
+      expect(item.errors.messages).to include(
+        'descriptive_metadata.name[0].role[0].term': ['has invalid key, "other_field" in hash', 'has invalid key, "other_field_2" in hash']
+      )
+    end
+  end
+
+  context 'when field does not exists' do
```
elohanlon, line 141:
```suggestion
  context 'when field does not exist' do
```

app/validators/digital_object/dynamic_fields_validator.rb
cgalarza: I am going to ticket this and work on it as a separate task.

lib/hyacinth/digital_object/type_def/dynamic_field_data.rb
cgalarza: Yes, if a custom_field is added/removed from a vocabulary all the terms for that vocabulary aren't updated to reflect that change. I wouldn't expect it often, but it could happen. It seems like the ideal solution would be to reindex the terms again after a vocabularies custom field change. We could update the `custom_fields` hash in `Hyacinth::Config.term_search_adapter.search`, but we would additionally also have to update the custom field's indexed fields.

lib/hyacinth/dynamic_fields_loader.rb
cgalarza: Yep, that's what I was thinking!

spec/hyacinth/digital_object/type_def/dynamic_field_data_spec.rb
cgalarza: Right!

lib/tasks/hyacinth/setup.rake
cgalarza: Yea, it's crazy how many lines it reduces.

lib/hyacinth/dynamic_fields_loader.rb
cgalarza: I am open about the name but I would rather not use "DynamicFieldCreator" because its not always creating a new field.

lib/hyacinth/digital_object/type_def/dynamic_field_data.rb
cgalarza: I'll look into a refactor. I need think more about how creating and update terms is going to happen.

app/validators/digital_object/dynamic_fields_validator.rb
elohanlon: Sounds good!

lib/hyacinth/adapters/term_search_adapter/solr.rb
elohanlon: Yeah, I think that would be the best way to go then.

lib/hyacinth/digital_object/type_def/dynamic_field_data.rb
elohanlon: Great, thanks.

lib/hyacinth/digital_object/type_def/dynamic_field_data.rb
elohanlon: Rather than reindexing the terms after a custom field change, what if we disallow deletion until the used values are cleared out?

The reason I like that approach is that it resembles a few other Hyacinth cases (that aren't yet implemented) for batch-scanning records when certain global changes occur:
- If someone attempts to delete a dynamic field, scan through DigitalObject solr docs to determine whether that field is in use and reject deletion if it is.
- If someone attempts to delete a project, do the same thing to make sure the project isn't in use.
- If someone wants to delete a controlled term from a vocabulary, do the same thing, making sure that term isn't referenced by any records.

And now we would have this other case for making sure that if someone attempts to delete a controlled term's custom field, we first determine if any terms are using that custom field.

Are custom fields with blank values absent from the solr document?  If so, then I think what I'm proposing would work and we wouldn't have to do any kind of additional reindexing of terms *after* deletion.  So to recap:

To clear a vocabulary's custom field that has NO occurrences in any terms:
1) delete operation first checks if it's present in any terms, which is isn't
2) custom field is deleted from vocabulary.  done.

To clear a vocabulary's custom field that HAS occurrences in some terms:
1) delete operation first checks if it's present in any terms, which it is. deletion rejected.
2) a Hyacinth user is responsible for clearing the field from the terms that use it (knowable via solr query) in whatever way we see fit (i.e. actually delete it OR migrate the value to a new custom field if applicable)
3) the updating of a term cascades and updates all of the digital objects that reference it (via reindexing of those digital objects) -- so this process depends on the change-term-and-reindex-digital-objects functionality we don't have yet, but that we need.
4) we retry the vocabulary custom field deletion again, and it succeeds because no terms reference the custom field anymore.

What do you think?  If this sounds reasonable, I can ticket it.

And then it would potentially still make sense to move the logic I mentioned in my earlier comment to the term search adapter.

lib/hyacinth/digital_object/type_def/dynamic_field_data.rb
```diff
@@ -0,0 +1,117 @@
...
+            terms.each do |term_hash|
+              solr_term_hash = search_results['response']['docs'].find do |result|
+                result['uri'] == term_hash['uri'] && result['vocabulary'] == vocab
+              end
+
+              next if solr_term_hash.blank?
+
+              full_term = {}
+
+              { 'authority' => '', 'pref_label' => '', 'alt_labels' => [], 'term_type' => '' }.map do |key, default|
```
elohanlon, line 53: Do we actually want a term_type default of `''`?  I think if a term comes in with no type, that would cause all sorts of problems and an empty string wouldn't make it any better.  Maybe not necessary to worry about that default here.

lib/hyacinth/dynamic_fields_loader.rb
elohanlon: Great!  I'm not tied to DynamicFieldCreator, but would mostly just prefer to avoid `Loader`.

Do you mean that you don't like the "Creator" part because it only creates fields that don't already exist?

lib/hyacinth/dynamic_fields_map.rb
cgalarza: Yes, added that in.

app/validators/digital_object/dynamic_fields_validator.rb
cgalarza: Updated.

spec/hyacinth/digital_object/type_def/dynamic_field_data_spec.rb
cgalarza: done

spec/hyacinth/digital_object/type_def/dynamic_field_data_spec.rb
cgalarza: done.

lib/hyacinth/digital_object/type_def/dynamic_field_data.rb
cgalarza: I would rather do this refactor as part of the create and update terms addition. I think there's other code that needs to be extracted out in addition to the method you pointed out.

lib/hyacinth/digital_object/type_def/dynamic_field_data.rb
cgalarza: One thing to consider: custom fields with blank values aren't always absent from the database record and the solr record if they are empty (at least not right now). Because we store custom fields in a hash if a `blank?` value has been set the key of that field is most likely present in the custom fields hash and has therefore been indexed. We index the custom fields json into a solr field. In order to assure that there are no blank values in the custom fields we would have to apply logic to that effect. 

I think the rests works but its very theoretical now. What are we going to implement in the short term? Remove the ability to delete custom fields?

lib/hyacinth/dynamic_fields_loader.rb
cgalarza: Yes.

lib/hyacinth/digital_object/type_def/dynamic_field_data.rb
cgalarza: I realize that some of these fields will never be empty, but I am just trying to assure that the empty fields do have a value. What's your suggestion? Should I just pull the fields that are always populated out of the logic and merge them separately?

cgalarza:
@elohanlon Where do you think is the best approach to "rehydrating terms" after save? Should it be part of the before_save call back that creates or updates terms? If so, I can try to PR those changes to this PR.

spec/hyacinth/dynamic_fields_map_spec.rb
cgalarza: Done.

lib/hyacinth/digital_object/type_def/dynamic_field_data.rb
elohanlon: Ticket sounds good!

lib/hyacinth/digital_object/type_def/dynamic_field_data.rb
elohanlon: For the short-term having no UI ability to delete custom fields works.  I think we may want to clear out empty custom field values in the near future, and then we'll work through the other aspects of this problem later on.

lib/hyacinth/dynamic_fields_loader.rb
elohanlon: "Creator" doesn't bother me because in conditions when it actually does something, that something is a "create" operation, but I guess if you want to express the conditional nature of the functionality in the class name (rather than calling it "Creator" and maybe just expressing "_if_exist" in a method name), you could call it `DynamicFieldConditionalCreator`?

cgalarza: That sounds really clunky. I would honestly just rather keep the original name. I think load is a general term that shouldn't be tied specifically to loading things into memory.

lib/hyacinth/digital_object/type_def/dynamic_field_data.rb
cgalarza: Okay, sounds good. But the logic here would stay the same, correct? or are you suggesting changes to this PR?

lib/hyacinth/dynamic_fields_loader.rb
elohanlon: All I mean is that load just doesn't sound accurate for what it's doing, since it's creating rather than loading.  Load also doesn't communicate the conditional thing you were concerned about, since it doesn't load things that have already been created.  But if you have a strong preference for load, feel free to keep it.  It was just a minor thing for me.

lib/hyacinth/digital_object/type_def/dynamic_field_data.rb
elohanlon: I was just suggesting that we remove `'term_type' => ''` from this hash since I didn't think it would do anything.

lib/hyacinth/digital_object/type_def/dynamic_field_data.rb
elohanlon: I'm good with your suggestion to make this future work, so no changes to this PR.

elohanlon:
@cgalarza Yeah, doing rehydration in before_save along with the create/update logic sounds good to me.  Combining those things should reduce the number of solr queries we need to do.  We can do one solr query to look for existing terms in the data to rehydrate AND determine which terms in the data don't yet exist (when they're missing from the solr response).

app/validators/digital_object/dynamic_fields_validator.rb
cgalarza: I can move it out!

lib/hyacinth/adapters/term_search_adapter/solr.rb
cgalarza: Will do. I realized that because we add the projects filter automatically the searches are pretty long.

lib/hyacinth/digital_object/type_def/dynamic_field_data.rb
cgalarza: 👍

lib/hyacinth/digital_object/type_def/dynamic_field_data.rb
cgalarza: It is fetching the key and moving it over to the full_term hash, so I can find an alternate way to do that for the fields that should always be populated.

lib/hyacinth/adapters/term_search_adapter/solr.rb
elohanlon: 👍

app/validators/digital_object/dynamic_fields_validator.rb
elohanlon: Thanks!

lib/hyacinth/digital_object/type_def/dynamic_field_data.rb
elohanlon: Sounds good!

PR #258, opened 2020-04-20
DigitalObject#dynamic_field_data -> DigitalObject#descriptive

cgalarza:
Renaming :dynamic_field_data to :descriptive as discussed.

For clarity, in places where :descriptive data was stored in a variable I named those variables `descriptive_metadata`. 

HYACINTH-559


cgalarza:
PR closed because it used the wrong field name.

PR #259, opened 2020-04-20
Development environment reset task + updated reindex task

elohanlon:
New development task (hyacinth:development:reset) clears all local data and sets up new basic setup data via hyacinth:setup:* tasks.

Also fixed purge_all_digital_objects task so that it no longer gets stuck on parent objects that have un-purged children.  Purge order doesn't matter when purging all objects.

### DIFF
--- a/lib/tasks/hyacinth.rake
+++ b/lib/tasks/hyacinth.rake
@@ -5,21 +5,22 @@ namespace :hyacinth do
     DigitalObjectRecord.find_in_batches(batch_size: (ENV['BATCH_SIZE'] || 1000).to_i) do |records|
       records.each do |record|
         ::DigitalObject::Base.find(record.uid).index(false)
-      rescue => e
+      rescue
         puts "Error while reindexing #{record.uid}. See raised error below:"
-        raise e
+        raise
       end
       Hyacinth::Config.digital_object_search_adapter.solr.commit
     end
   end
 
   task purge_all_digital_objects: :environment do
-    puts Rainbow("This will delete ALL digital objects in Rails.env #{Rails.env} and cannot be undone. Are you sure you want to do this? (yes/no)").red.bright
+    puts Rainbow("This will delete ALL digital objects in the selected Rails environment (#{Rails.env}) and cannot be undone. "\
+      "Please confirm that you want to continue by typing in the selected Rails environment (#{Rails.env}):").red.bright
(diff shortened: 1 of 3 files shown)


lib/tasks/hyacinth.rake
```diff
@@ -3,15 +3,20 @@
 namespace :hyacinth do
   task reindex: :environment do
     DigitalObjectRecord.find_in_batches(batch_size: (ENV['BATCH_SIZE'] || 1000).to_i) do |records|
-      records.each { |record| ::DigitalObject::Base.find(record.uid).index(false) }
+      records.each do |record|
+        ::DigitalObject::Base.find(record.uid).index(false)
+      rescue => e
+        puts "Error while reindexing #{record.uid}. See raised error below:"
+        raise e
```
barmintor, line 10: IIRC you can actually just use an unmapped `rescue` and `raise` - the no arg raise I think raises the last exception

lib/tasks/hyacinth.rake
```diff
@@ -3,15 +3,20 @@
...
+      end
       Hyacinth::Config.digital_object_search_adapter.solr.commit
     end
   end
 
   task purge_all_digital_objects: :environment do
     puts Rainbow("This will delete ALL digital objects in Rails.env #{Rails.env} and cannot be undone. Are you sure you want to do this? (yes/no)").red.bright
     print '> '
-    response = STDIN.gets.chomp
+    response = ENV['yes'] || STDIN.gets.chomp
```
barmintor, line 19: I would ticket this separately, but I wonder if the gravity of it might even make having the user verify the name of the environment they are deleting here instead of a simple 'yes'

lib/tasks/hyacinth.rake
elohanlon: I didn't realize you could do that!  I'll make that change.  Thanks!

lib/tasks/hyacinth.rake
elohanlon: Good call.  For the sake of safety, I'll make that change right now.  Should be easy.  Thanks!

elohanlon:
Thanks for reviewing!  I'll merge!

Even though the github travis check display is still showing up as pending, the tests actually passed on the travis side.  Maybe some github-to-travis communication weirdness.

PR #262, opened 2020-04-28
Batch import processing

elohanlon:


app/graphql/mutations/batch_import/create_batch_import.rb
```diff
@@ -4,24 +4,32 @@ class Mutations::BatchImport::CreateBatchImport < Mutations::BaseMutation
   argument :priority, Enums::BatchImportPriorityEnum, required: false
   argument :signed_blob_id, String, "Signed blob ID generated via `/api/v1/uploads`", required: true
 
-  field :batch_import, Types::BatchImportType, null: false
+  field :batch_import, Types::BatchImportType, null: true
```
barmintor, line 7: `:batch_import` can be null?

barmintor: I see, on error. Hmm.

elohanlon: Yeah, if validation of the submitted CSV data fails (and then errors are reported).

app/graphql/types/query_type.rb
```diff
@@ -91,7 +91,6 @@ class QueryType < Types::BaseObject
     end
 
     def digital_objects(**arguments)
-      # TODO: consider object read permissions via projects
```
barmintor, line 94: 👍

app/javascript/hyacinth_v1/components/batch_imports/BatchImportIndex.js
```diff
@@ -30,13 +30,19 @@ function BatchImportIndex() {
   };
 
   const onCancel = (id) => {
-    const variables = { input: { id, cancelled: true } };
-    updateBatchImport({ variables }).then(refetch);
+    // eslint-disable-next-line no-alert
+    if (window.confirm(`Are you sure you want to cancel Export Job ${id}?`)) {
```
barmintor, line 34:
```suggestion
    if (window.confirm(`Are you sure you want to cancel Import Job ${id}?`)) {
```

barmintor: I think another one below in this file

app/javascript/hyacinth_v1/components/shared/TopNavbar.js
```diff
@@ -57,13 +57,13 @@ function TopNavbar(props) {
 
             <NavDropdown.Divider />
 
-            <LinkContainer to="/batch_exports">
```
barmintor, line 60: 🤔

app/javascript/hyacinth_v1/components/batch_imports/BatchImportIndex.js
elohanlon: Good catch, thanks!

app/javascript/hyacinth_v1/components/shared/TopNavbar.js
elohanlon: ?

app/javascript/hyacinth_v1/components/batch_imports/BatchImportIndex.js
```diff
@@ -30,13 +30,19 @@ function BatchImportIndex() {
...
+      const variables = { input: { id, cancelled: true } };
+      updateBatchImport({ variables }).then(refetch);
+    }
   };
 
   const onDelete = (id) => {
-    const variables = { input: { id } };
-    deleteBatchImport({ variables }).then(refetch);
+    // eslint-disable-next-line no-alert
+    if (window.confirm(`Are you sure you want to delete Export Job ${id}?`)) {
```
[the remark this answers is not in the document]
elohanlon: Just realized that I actually meant to say "Batch Import" rather than "* Job", so I'm updating that now in my local copy.  Thanks for catching this though, since it definitely shouldn't say "Export".

app/javascript/hyacinth_v1/graphql/fileUploads.js
```diff
@@ -1,9 +1,14 @@
 import * as ActiveStorage from 'activestorage';
 
-const uploadFile = (file, url) => {
+const uploadFile = (file, url, onProgress) => {
   const delegate = {
-    directUploadWillCreateBlobWithXHR: (data) => { console.debug(data); },
-    directUploadWillStoreFileWithXHR: (data) => { console.debug(data); },
+    directUploadWillCreateBlobWithXHR: (request) => { console.debug(request); },
+    directUploadWillStoreFileWithXHR: (request) => {
+      request.upload.addEventListener('progress', (event) => {
```
barmintor, line 7: faint memory ringing - is there a memory issue with disposal of these event listeners?

app/javascript/hyacinth_v1/components/shared/TopNavbar.js
elohanlon: I switched the order of the links to match Hyacinth 2.  We accidentally had them reversed here.

app/jobs/digital_object_import_processing_job.rb
```diff
@@ -0,0 +1,87 @@
...
+      apply_recursive_failure!(digital_object_import)
+      return
+    end
+
+    # Mark this digital object import as successful
+    digital_object_import.success!
+
+    # Now we'll find and destroy all ImportPrerequisites that this DigitalObjectImport was a
+    # prerequisite for, but store the returned object values so we can queue their processing next.
+    destroyed_import_prerequisites = ImportPrerequisite.where(prerequisite_digital_object_import: digital_object_import).destroy_all
```
cgalarza, line 33: okay, i see where this model came in handy.

app/models/digital_object_import.rb
```diff
@@ -1,11 +1,12 @@
 # frozen_string_literal: true
 
 class DigitalObjectImport < ApplicationRecord
-  enum status: { pending: 0, in_progress: 1, success: 2, failure: 3 }
+  enum status: { pending: 0, queued: 1, in_progress: 2, success: 3, failure: 4 }
 
   serialize :import_errors, Array
 
   belongs_to :batch_import
+  has_many :import_prerequisites, dependent: :destroy
```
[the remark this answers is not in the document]
cgalarza: I see where this came in handy later.

app/javascript/hyacinth_v1/components/batch_imports/BatchImportNew.js
```diff
@@ -43,31 +88,38 @@ function BatchImportNew() {
...
-      <GraphQLErrors errors={createError} />
+      <GraphQLErrors errors={createError || validateError || startError} />
 
       <div className="m-2">
         <Alert variant="info">
           Important Note: Please avoid using Microsoft Excel for CSV editing,
           imports, or exports. Excel doesn&apos;t handle UTF-8 properly.
         </Alert>
-
+        {validationResult && (<Alert variant="success">This Batch Import appears to be valid.</Alert>) }
```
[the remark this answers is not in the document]
elohanlon: Haha yeah.

app/models/batch_import.rb
```diff
@@ -81,6 +82,64 @@ def csv_without_successful_imports
...
+      next false if key.start_with?('_')
+      dynamic_field_data[key] = import_json[key]
+      true
+    end
+
+    # Assign dynamic_field_data to 'dynamic_field_data' key in import_json
+    import_json['dynamic_field_data'] = dynamic_field_data
+
+    # Remove leading underscore from all remaining underscore-prefixed keys
+    import_json.transform_keys! { |key| key.start_with?('_') ? key[1..-1] : key }
```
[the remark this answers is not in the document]
elohanlon: It does, yeah.  The current plan is for rights fields in a spreadsheet to be prefixed with the rights typedef field -- right now `_rights`, and I guess `_rights_metadata` when we update it to that?

I'll need to update this code when I merge master into this branch before merging the PR, changing `dynamic_field_data` to `descriptive_metadata`.

app/models/digital_object_import.rb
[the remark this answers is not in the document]
elohanlon: 👍

app/models/import_prerequisite.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class ImportPrerequisite < ApplicationRecord
+  # A digital_object_import actually is required on save, but it can be supplied as a
+  # digital_object_import_id rather than a digital_object_import.
+  belongs_to :digital_object_import, required: false
+  belongs_to :batch_import
+  belongs_to :prerequisite_digital_object_import, class_name: 'DigitalObjectImport'
+
+  # Note, below, that we're requiring digital_object_import_id and prerequisite_digital_object_import_id
```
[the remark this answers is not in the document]
elohanlon: I originally thought so too, but it turns out that the model wouldn't save.  It you do a `presence` check for `digital_object_import`, `presence` apparently validates the presence of the `digital_object_import` rather than `digital_object_import_id`.

The reason this was a problem is that in `batch_import_start_job.rb`, I have a potentially large number of `digital_object_import_id` and `prerequisite_digital_object_import_id` numeric values and I didn't want to do a ton of ActiveRecord lookups for all of those objects.  That's why I needed to use IDs.  One alternative would have been to do a bunch of literal SQL insert queries instead, but I wasn't hesitant to totally bypass all ActiveRecord validations.  I guess it might not be the worst thing, since it's a fairly straightforward insert operation on guaranteed-existing records.  Do you think that would be worth it? 

One alternative that I considered was

lib/hyacinth/batch_import/index_prerequisite_map.rb
```diff
@@ -0,0 +1,237 @@
...
+      #   7 => [2],    # 7 requires 2
+      #   10 => [9]    # 10 requires 9
+      # }
+      def self.generate(csv_file)
+        # First, we'll generate two maps:
+        identifiers_to_row_numbers = {}
+        row_numbers_to_parent_identifiers = {}
+
+        ::BatchImport.csv_file_to_hierarchical_json_hash(csv_file) do |json_hash_for_row, csv_row_number|
+          identifiers = json_hash_for_row['identifiers'] || []
```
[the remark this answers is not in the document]
elohanlon: Yeah, that's exactly what it is -- the generic "identifiers" array attribute on a DigitalObject.

lib/hyacinth/batch_import/index_prerequisite_map.rb
```diff
@@ -0,0 +1,237 @@
...
+          end
+
+          # Raise an error if we encountered any unresolvable identifiers
+          if unresolvable_identifiers.present?
+            raise Hyacinth::Exceptions::BatchImportError,
+                  "The following parent identifiers could not be found in CSV data or "\
+                  "among existing Hyacinth objects: #{unresolvable_identifiers.join(', ')}"
+          end
+
+          # If there are any duplicate values in parent_row_numbers, that indicates that the
```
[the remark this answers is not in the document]
elohanlon: The parent_uids attribure on DigitalObject::Base is a Set, so it shouldn't make it into our data model.  But also, if we encounter a duplicate parent reference in a user-submitted CSV then it's definitely a mistake, and data that we don't want to send downstream to other prerequisite-related operations, so that's why I'm raising the error here.

It's specifically about this kind of case:
_parent_digital_objects[0].identifier = 'parent1'
_parent_digital_objects[1].identifier = 'parent2'
_parent_digital_objects[2].identifier = 'parent1'

And this (probably extremely rare) other case:
_parent_digital_objects[0].uid = 'uid1'
_parent_digital_objects[1].uid = 'uid2'
_parent_digital_objects[2].identifier = 'identifier-that-points-to-record-with-uid1'

app/javascript/hyacinth_v1/graphql/fileUploads.js
elohanlon: This is a good concern, and I've been doing some reading and it seems that it's not a problem in today's modern browsers -- with plain JavaScript, at least.  Use of jQuery may increase the chance of memory leaks because of the different way that jQuery binds events (we'd want to look more into that case), but fortunately we're not using jQuery here.  React does some fancy things behind the scenes with the way it manages the DOM, but since React promotes the use of `onClick` (rather than the usual `onclick`) and hasn't required explicit cleanup of those event handlers, I'm guessing that pattern also avoids memory leaks.

One other thing that makes this particular case even better for garbage collection is that the event listener is bound to a request object rather than a DOM element.  One of the major reasons for memory leaks in older browsers (especially IE) was a garbage collection issue where event listeners held onto references to DOM elements, which caused issues with cleanup of those elements, and I believe this led to circular reference cleanup issues.

Mentioning @cgalarza here just as an FYI, since this is good for all of us to know about going forward.

barmintor: I guess another minor performance point: It's kind of extravagant to have 3 inline lambdas with the same body, isn't it?

elohanlon: It is, though I think not doing so makes it harder to pass `onProgress` to the `addEventListener` inner context (unless we can do something with `bind`, maybe?).  Do you have a preferred refactor in mind?

Here's a slightly cleaned up version, at least.  I removed some stuff that appeared to be unused.  What do you think?

```
const uploadFile = (file, url, onProgress) => {
  const delegate = {
    directUploadWillStoreFileWithXHR: (request) => {
      request.upload.addEventListener('progress', (event) => {
        onProgress(Math.round(event.loaded * 100 / event.total));
      });
    },
  };
  const directUpload = new ActiveStorage.DirectUpload(file, url, delegate);
  return new Promise((resolve, reject) => {
    directUpload.create((error, props) => {
      if (error) {
        reject(error);
      } else {
        resolve(props);
      }
    });
  });
};
```

app/models/batch_import.rb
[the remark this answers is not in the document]
cgalarza: Ahhh okay, this works for me!

lib/hyacinth/batch_import/index_prerequisite_map.rb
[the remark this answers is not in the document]
cgalarza: OHH okay, I totally did not understand what this check was doing. Thanks for explaining!

app/models/import_prerequisite.rb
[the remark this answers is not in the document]
cgalarza: I definitely prefer your solution to bypassing ActiveRecord validation.

I just added a presence check on digital_object_import for batch_import and create a new digital object import by setting batch_import_id and it worked. I didn't get any errors. So, I am wondering if ActiveRecord is getting confused somehow. 

Maybe the `inverse_of` should be set, like described here: https://guides.rubyonrails.org/active_record_validations.html#presence

It also seems that since `require: true` is the default for belongs_to relationships the additional presence check isn't necessary because `require: true` automatically requires a presence check. 
https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-belongs_to

lib/hyacinth/batch_import/index_prerequisite_map.rb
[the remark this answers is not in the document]
cgalarza: This might not be something we need to worry about, but if someone adds a uid that points at a different object in Hyacinth, in theory bad things would happen? Like information would get mapped to the wrong object?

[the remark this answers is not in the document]
elohanlon: Yeah, you're right, I don't think we need to worry about that.  I don't think we can handle that kind of user mistake since if they put the wrong _uid field value for a row then they're basically saying "this is the UID for the record I want to update for this row" instead of their intended one.

app/models/import_prerequisite.rb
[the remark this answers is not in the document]
elohanlon: Aha!  I was actually able to just change back to normal associations and it works.  I'm not sure what was going on before, but the code should now look more like you'd expect.  Tests are passing, so it looks like things are good.

app/models/batch_import.rb
```diff
@@ -81,6 +82,66 @@ def csv_without_successful_imports
...
+  #   Sample return value for failed validation:
+  #   [false, ['Something went wrong!', 'Something else went wrong!']]
+  #   Sample return value for successful validation:
+  #   [true, []]
+  def self.pre_validate_blob(blob)
+    pre_validation_errors = []
+    blob.open do |file|
+      # TODO: Make sure there aren't any duplicate headers
+      csv_file_to_hierarchical_json_hash(file) do |json_hash_for_row, _csv_row_number|
+        # TODO: Make sure that only valid fields are present in the dynamic field data properties
```
cgalarza, line 109: This work should be ticketed, I don't think it's represented anywhere.

elohanlon: HYACINTH-602

cgalarza: thanks!

spec/jobs/digital_object_import_processing_job_spec.rb
```diff
@@ -2,7 +2,185 @@
...
+
+        it "has the expected dyanmic field data" do
+          expect(created_digital_object.dynamic_field_data).to eq(
+            'title' => [{ 'sort_portion' => 'The', 'non_sort_portion' => 'Cool Item' }],
+            'abstract' => [{ 'abstract_value' => 'some abstract' }]
+          )
+        end
+
+        it "is marked as created and updated by the expected user" do
+          expect(created_digital_object.created_by).to eq(nil)
```
cgalarza, line 51: Is the created_by/updated_by user not set? I am probably missing something, sorry.

elohanlon: I'm glad you pointed this out!  That test was definitely not testing the right thing, and revealed a bug.  Just fixed it with my latest commit.  Thanks!

elohanlon:
@cgalarza I believe I've addressed all of your feedback, but let me know if there's anything else.  Thanks for all of the helpful comments!

PR #265, opened 2020-05-14
Adding orderBy to digitalObjects query

cgalarza:


app/graphql/types/query_type.rb
```diff
@@ -29,6 +29,7 @@ class QueryType < Types::BaseObject
 
     field :digital_objects, DigitalObjectInterface.results_type, null: true, extensions: [Types::Extensions::SolrSearch, Types::Extensions::MapToDigitalObjects] do
       description "List and searches all digital objects"
+      argument :order_by, Inputs::DigitalObject::OrderByInput, required: false, default_value: { field: 'score', direction: 'desc' }
```
elohanlon, line 32: We initially had score-based default sorting in Hyacinth a long time ago, but users ended requesting and preferring title-based default sorting instead so we switched to that.

app/graphql/enums/digital_object/order_fields_enum.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class Enums::DigitalObject::OrderFieldsEnum < Types::BaseEnum
+  value 'RELEVANCE', 'sorting by relevance score', value: 'score'
```
elohanlon, line 4: The all-caps sort field names stood out to me as an uncommon practice, based on other APIs I've used.  Is it okay if we go for lower case?  Or is there a reason that you prefer caps?  Also, you have lower-case asc and desc values in params.rb (`VALID_SORT_DIRECTION = ['asc', 'desc'].freeze`).

elohanlon: One other thought about the translation of these enums to fields like `updated_at_dtsi`: It seems like we're coupling GraphQL too tightly with solr here.  I'm wondering if the translation to solr fields should happen elsewhere.

app/graphql/enums/order_direction_enum.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+class Enums::OrderDirectionEnum < Types::BaseEnum
+  value 'ASC', 'ascending', value: 'asc'
+  value 'DESC', 'descending', value: 'desc'
```
elohanlon, line 5: Similar thought here as above about the enums being capitalized.

app/graphql/enums/digital_object/order_fields_enum.rb
```diff
@@ -0,0 +1,7 @@
...
+  value 'LAST_MODIFIED', 'sorting by last modified at', value: 'updated_at_dtsi'
+  value 'TITLE', 'sorting by non-sort title', value: 'sort_title_ssi'
```
elohanlon, line 6: Updated comment:

Seeing this capital enum made me look up whether or not enums should be capitalized with GraphQL. And yes, they should! http://spec.graphql.org/June2018/#sec-Enum-Value

So you have that right, but we may want to update some of our other enums that aren't all-caps values.

app/graphql/types/query_type.rb
cgalarza: Do you think that should be the default for the API and UI?

app/graphql/enums/digital_object/order_fields_enum.rb
cgalarza: I'll create a ticket about updating the Enums to caps and I can do that work in a subsequent PR.

Yea, I don't love putting the solr keys there but I don't have a great suggestion for how to decouple it. This problem will come up in other places too (filters, etc). I would rather leave it as is for now and iterate on it later.

cgalarza: RE: Solr keys problem, I got a few ideas last night but I would like to see where all the seams are before deciding on an approach.

app/graphql/inputs/digital_object/order_by_input.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module Inputs
+  class DigitalObject::OrderByInput < Types::BaseInputObject
+    description 'Digital Object Sort Parameters'
+
+    argument :field, Enums::DigitalObject::OrderFieldsEnum, required: false, default_value: 'score'
```
barmintor, line 7: are the default_value's here not supposed to be the enum values?

app/graphql/types/query_type.rb
barmintor: If we can separate the two, I would use the Hyc2 default in the UI.

lib/hyacinth/adapters/digital_object_search_adapter/solr/document_generator.rb
```diff
@@ -23,9 +23,12 @@ def merge_core_fields_for!(digital_object, solr_document = {})
             'digital_object_type_ssi' => digital_object.digital_object_type,
             'doi_ssi' => digital_object.doi,
             'identifier_ssim' => digital_object.identifiers.to_a,
-            'title_ssi' => indexable_title,
+            'title_ssi' => digital_object.generate_title,
```
barmintor, line 26: if this isn't separately used as an index, could it just be `title_ss` now?

app/graphql/inputs/digital_object/order_by_input.rb
cgalarza: That's what I thought at first, but its supposed to be the actual value the enum is representing.

app/graphql/types/query_type.rb
cgalarza: I expect that from the UI we would always set the sort parameters, so we can definitely specify a different sort value for the UI and API.

lib/hyacinth/adapters/digital_object_search_adapter/solr/document_generator.rb
cgalarza: If a user searches the full title of an object (with the non sort portion) would it still show up in the search? Would it only show up in the search if the non sort portion had only stop words in it?

barmintor: it depends on how we want to do the title search, but I am assuming that would be a separate solr field, indexed as text from the configured fields. But it's possible that `digital_object.generate_title` should be passed to `add_keywords` instead of `indexable_title`. Here I was just saying that if we only use the `title_ssi` field to show, it doesn't need to be separately indexed (unless I am missing a use case!)

cgalarza: ahh okay that makes sense. So I should convert title_ssi to title_ss and then change the title that is indexed for keyword search. I can make that change.

cgalarza: Done.

app/graphql/types/query_type.rb
cgalarza: @elohanlon: I am going to merge this request, but if you think the API default should be different let me know and I can change it in a subsequent request. I noted that the default for the UI should be title sort here: https://issues.cul.columbia.edu/browse/HYACINTH-571

elohanlon: @cgalarza Default in UI-only sounds good to me!  Thank you for the ticket.

app/graphql/enums/digital_object/order_fields_enum.rb
elohanlon: 👍

PR #268, opened 2020-05-22
Term model ensuring vocabulary isn't locked before creating/updating/deleting

cgalarza:
Adding Term model before_save and before_destroy callbacks that ensure the related vocabulary is not locked. If the vocabulary is locked no changes should be made to any terms in the vocabulary.

HYACINTH-604


app/models/term.rb
```diff
@@ -141,6 +142,11 @@ def pref_label_unchanged_for_temp_term
       errors.add(:pref_label, 'cannot be updated for temp terms') if pref_label_changed?
     end
 
+    # Raise error if vocabulary is locked
+    def vocabulary_locked!
+      raise Hyacinth::Exceptions::VocabularyLocked, 'term cannot be created/updated/deleted while vocabulary is locked' if vocabulary.locked?
```
elohanlon, line 147: This suggestion might be too overly-cautious, but should this check happen within a with_lock block to reload the vocabulary instance and make sure that you're working with the latest state of `vocabulary.locked`, in case it changed since the Term was instantiated?  As in:

```
# with_lock forces a vocabulary.reload and also locks on the vocabulary row in the db
vocabulary.with_lock do
  raise Hyacinth::Exceptions::VocabularyLocked, 'term cannot be created/updated/deleted while vocabulary is locked' if vocabulary.locked?
end
```
The downside is that it means an extra SQL query for every term save.  I guess the worst thing that could happen if we don't use `with_lock` is that someone could instantiate a term, then an external process could lock the vocabulary, and then the term would successfully save because the term's local instance of `vocabulary` would have a `locked` value of true.

elohanlon: Depending on what you and @barmintor decide, it may or may not be worth adding an additional test for the case I mentioned (term creation, separate locking of vocab that's not on the term's instance, then attempt to save).

cgalarza: I see locking and unlocking vocabularies as a very rare occurrence. I am not sure its worth the overhead of reloading the object. I am not against it tho. What do you think @barmintor?

barmintor: I think this is worth thinking about, but we should ticket it separately. I see the concern, but it's not clear to me that this is the right place to enforce the locking and the performance hit might be substantial.

cgalarza: Ticketed: HYACINTH-615

PR #277, opened 2020-06-29
add specs for Solr field name generators

barmintor:
- spec Hyacinth::DigitalObject::Facets
- spec Hyacinth::DigitalObject::SolrKeys

### DIFF
--- a/spec/hyacinth/digital_object/facets_spec.rb
+++ b/spec/hyacinth/digital_object/facets_spec.rb
@@ -23,7 +23,7 @@ describe Hyacinth::DigitalObject::Facets do
         Hyacinth::DigitalObject::SolrKeys.for_dynamic_field(path)
       end
     end
-    it { expect(described_class.all_solr_keys - expected).to eql([]) }
+    it { expect(described_class.all_solr_keys).to match_array(expected) }
   end
   describe '.all_facetable_fields' do
     let!(:expected) { DynamicField.where(is_facetable: true).all }
--- a/spec/hyacinth/digital_object/facets_spec.rb
+++ b/spec/hyacinth/digital_object/facets_spec.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe Hyacinth::DigitalObject::Facets do
+  let(:descriptive_category) { FactoryBot.create(:dynamic_field_category) }
+  let(:rights_category) { FactoryBot.create(:dynamic_field_category, display_label: 'Rights', metadata_form: 'item_rights') }
+  let(:name_group) { FactoryBot.create(:dynamic_field_group, parent: descriptive_category) }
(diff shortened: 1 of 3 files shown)


app/graphql/types/extensions/solr_search.rb
```diff
@@ -35,11 +36,11 @@ def after_resolve(object:, arguments:, value:, context:, memo:)
       # parse the facet objects or return an empty array
       def facets(facet_counts)
         return [] unless facet_counts['facet_fields']
-        display_label_map = Hyacinth::DigitalObject::Facets.facet_display_label_map
+        # display_label_map = Hyacinth::DigitalObject::Facets.facet_display_label_map
         facet_counts['facet_fields'].map do |facet_field, value_counts|
           {
             field_name: facet_field,
-            display_label: display_label_map[facet_field] || facet_field.split('_')[0...-1].map(&:titlecase).join(' '),
+            display_label: facet_display_label_map[facet_field] || facet_field.split('_')[0...-1].map(&:titlecase).join(' '),
```
elohanlon, line 43: Doesn't this end up invoking the `facet_display_label_map` method once per map iteration instead of caching the result and reusing it?

barmintor: 😂 yes, yes it does. what can I say I love geometric performance problems

barmintor: I guess we should *not* delegate to the other class.

elohanlon: 👍

spec/hyacinth/digital_object/facets_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  describe '.all_solr_keys' do
+    let!(:expected) do
+      [term_field, role_field, rights_field].map do |df|
+        path = df.path[1..-1].map(&:string_key) << df.string_key
+        Hyacinth::DigitalObject::SolrKeys.for_dynamic_field(path)
+      end
+    end
+    it { expect(described_class.all_solr_keys - expected).to eql([]) }
```
cgalarza, line 26:
```suggestion
    it { expect(described_class.all_solr_keys).to match_array(expected) }
```

cgalarza: Might be easier to read?

barmintor: I think we'd have to add sorting as well, right?

cgalarza: oh yea, potentially

cgalarza: oh actually! It order shouldn't matter: https://rubydoc.info/github/rspec/rspec-expectations/RSpec/Matchers:match_array

It uses contains_exactly and that method is not dependent on order.

barmintor: Oh! Ok, let's do it.

PR #279, opened 2020-07-30
Move EnabledDynamicFields API to GraphQL

barmintor:
HYACINTH-417: project/enabledFields graphql endpoint
HYACINTH-620: create digital object mutation (necessary to remove enabled fields REST API)
HYACINTH-621: update digital object mutation (necessary to remove enabled fields REST API)


barmintor:
Hmm, first few commit should be squashed

app/graphql/inputs/enabled_dynamic_field_input.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Inputs
+  class EnabledDynamicFieldInput < Types::BaseInputObject
```
elohanlon, line 4: On the ruby-graphql Input Object page (https://graphql-ruby.org/type_definitions/input_objects), they show an example subclass that ends in "Attributes" (Types::PostAttributes), but later in the page also show an example subclass that ends in "Input" (Types::DateRangeInput).  We've used both forms in HYC3, but I'd like to recommend that we standardize on one.  If you agree, do you prefer one convention over the other?

barmintor: Hmm... on the one hand the document illustrates a certain lack of common practice :) But some convention makes it easier to focus your mental energy. If we were going to standardize, I think I prefer *Input - which I think also aligns with the generated bits of the schema, right?

elohanlon: *Input sounds good to me, and you're right about it aligning well.  If okay, I'll create a ticket as a TODO for later?

The other thing I like about having "Input" in the name is that because an input object can never be returned as a result, it'll help us to quickly differentiate between any similarly-named Types that actually can be returned as a result (e.g. when searching through files in an IDE).

app/graphql/inputs/enabled_dynamic_field_input.rb
```diff
@@ -0,0 +1,17 @@
...
+    description 'Attributes for an enabled dynamic field in a project for a digital obect type'
+
+    argument :dynamic_field, Inputs::IdReference, required: true
```
elohanlon, line 7: Is it possible to do `argument :dynamic_field, ID, required: true` instead of the above `Inputs::IdReference`?  (And same question for `field_sets` below.)

elohanlon: Though with my suggestion, that may involve changing the name of the arguments to `dynamic_field_id` and `field_set_ids`, if we were passing IDs instead of objects.  But if we're only ever passing an object with one id attribute, would it be worth just naming the fields to indicate that they only accept IDs?  And then just use IDs?

elohanlon: Ah, I see a pattern with the `Project::StringKey` input object.

elohanlon: Good pattern!  Withdrawing my suggestion!

app/graphql/mutations/digital_object/create_digital_object.rb
```diff
@@ -0,0 +1,31 @@
...
+      field :errors, [Types::Errors::FieldedInput], null: false
+
+      def resolve(project:, digital_object_type:, descriptive_metadata:, identifiers:)
+        project = Project.find_by!(project.to_h)
+        ability.authorize! :create_objects, project
+
+        digital_object = Hyacinth::Config.digital_object_types.key_to_class(digital_object_type).new
+        digital_object.primary_project = project
+        digital_object.assign_attributes(
+          'identifiers' => Set.new(identifiers), 'descriptive_metadata' => descriptive_metadata
```
elohanlon, line 21: The assign_attributes method accepts identifiers as an array and converts them into a set internally (in internal assign_identifiers method), so no need to create a new Set here.

app/graphql/mutations/digital_object/create_digital_object.rb
```diff
@@ -0,0 +1,31 @@
...
+        )
+        if digital_object.save(update_index: true, user: context[:current_user])
```
elohanlon, line 23: The save method no longer accepts an `update_index` key because we call index in an after_save callback -- though as I'm noticing as I search through the current code base, we're still unnecessarily passing in a few other cases (unrelated to this PR).

app/graphql/mutations/digital_object/update_descriptive_metadata.rb
```diff
@@ -0,0 +1,31 @@
...
+      field :errors, [Types::Errors::FieldedInput], null: false
+
+      def resolve(id:, descriptive_metadata:, identifiers:)
+        digital_object = ::DigitalObject::Base.find(id)
+        ability.authorize! :update, digital_object
+        digital_object.assign_attributes(
+          'identifiers' => Set.new(identifiers),
+          'descriptive_metadata' => descriptive_metadata
+        )
+        if digital_object.save(update_index: true, user: context[:current_user])
```
elohanlon, line 20: Same note here about `update_index` not being a valid option anymore.

app/graphql/mutations/digital_object/update_descriptive_metadata.rb
```diff
@@ -0,0 +1,31 @@
...
+      argument :identifiers, [String], required: true
+
+      field :digital_object, Types::DigitalObjectInterface, null: false
+      field :errors, [Types::Errors::FieldedInput], null: false
+
+      def resolve(id:, descriptive_metadata:, identifiers:)
+        digital_object = ::DigitalObject::Base.find(id)
+        ability.authorize! :update, digital_object
+        digital_object.assign_attributes(
+          'identifiers' => Set.new(identifiers),
```
elohanlon, line 17:
```suggestion
          'identifiers' => identifiers,
```
Same note here about not needing to convert identifiers to a Set.

app/graphql/types/query_type.rb
```diff
@@ -160,6 +178,23 @@ def dynamic_field(id:)
...
     end
 
+    def enabled_dynamic_fields(project:, digital_object_type:)
+      project = Project.find_by!(project.to_h)
+      ability.authorize!(:read, project)
+      join_query = "SELECT dynamic_fields.id AS df_id, enabled_dynamic_fields.* FROM dynamic_fields"\
+                   " LEFT OUTER JOIN enabled_dynamic_fields ON enabled_dynamic_fields.dynamic_field_id = dynamic_fields.id"\
+                   " AND enabled_dynamic_fields.project_id = #{project.id} AND enabled_dynamic_fields.digital_object_type = '#{digital_object_type}'"
+      field_sets = project.field_sets.where(enabled_dynamic_fields: { digital_object_type: digital_object_type }).includes(:enabled_dynamic_fields)
+      DynamicField.connection.select_all(join_query).map do |result|
```
elohanlon, line 188: Good use of raw SQL query.  :+1:

Do you think it would be worth moving the SQL query lines and `DynamicField.connection.select_all(join_query)` to a clearly-named static method on the DynamicField model (that would take two arguments -- project id and digital object type)?  It would compress this method a bit and would improve skim-ability.

app/javascript/hyacinth_v1/components/digital_objects/metadata/MetadataForm.js
```diff
@@ -78,78 +125,75 @@ function MetadataForm(props) {
...
-    );
-  };
+  const keepEnabledFields = (enabledFieldIds, children) => children.map((c) => {
+    switch (c.type) {
+      case 'DynamicFieldGroup':
+        c.children = keepEnabledFields(enabledFieldIds, c.children);
+        return c.children.length > 0 ? c : null;
+      case 'DynamicField':
+        if (enabledFieldIds.includes(c.id)) return c;
+        // if the ids are fetched in queries ID scalars and JSON, strings must be compared
```
elohanlon, line 135: Is this comment missing a word?  I feel like I'm reading it wrong.

app/javascript/hyacinth_v1/components/digital_objects/metadata/MetadataForm.js
```diff
@@ -78,78 +125,75 @@ function MetadataForm(props) {
...
+        if (enabledFieldIds.includes(String(c.id))) return c;
+        return null;
+      default:
+        return c;
+    }
+  }).filter(c => c !== null);
 
   // TODO: Replace effect below with GraphQL when we have a GraphQL DynamicFieldCategories API
```
elohanlon, line 143: The referenced effect is gone!  Okay to delete comment?

app/javascript/hyacinth_v1/components/digital_objects/new/DigitalObjectNewForm.js
```diff
@@ -30,15 +30,14 @@ function DigitalObjectNewForm() {
...
-
   return (
     <>
       <ContextualNavbar
         title={`New ${capitalize(initialDigitalObject.digitalObjectType)}`}
         rightHandLinks={[{ link: '/digital_objects', label: 'Back to Digital Objects' }]}
       />
       <DigitalObjectSummary digitalObject={initialDigitalObject} />
-      <MetadataForm formType="new" digitalObject={initialDigitalObject} />
+      <MetadataForm formType="new" digitalObject={initialDigitalObject} project={project} createType={digitalObjectType} />
```
elohanlon, line 40: Instead of passing in project and createType, are you able to internally use (within the MetadataForm) the primaryProject and digitalObjectType fields of the passed-in digitalObject?

app/javascript/hyacinth_v1/components/shared/forms/FormButtons.js
```diff
@@ -21,7 +21,7 @@ function FormButtons(props) {
       </Col>
 
       <Col sm="auto">
-        <ResponsiveSubmitButton saveData={onSave} formType={formType} />
+        <ResponsiveSubmitButton key={`${formType}-submit`} saveData={onSave} formType={formType} />
```
elohanlon, line 24: Why is the key attribute necessary here if this element doesn't have any siblings?

app/javascript/hyacinth_v1/components/shared/forms/buttons/ResponsiveSubmitButton.js
```diff
@@ -1,48 +1,45 @@
...
-  }
-}
+  return (
+    <Button variant={variant} type="submit" onClick={onClick} disabled={isSaving} {...rest}>
+      {savingText}
+    </Button>
+  );
+};
 
 export default withRouter(ResponsiveSubmitButton);
```
elohanlon, line 45: Is withRouter still used now that this is a functional component?

app/javascript/hyacinth_v1/components/shared/forms/buttons/ResponsiveSubmitButton.js
```diff
@@ -1,48 +1,45 @@
...
-    this.setState({ isSaving: true });
-    const { saveData } = this.props;
-
-    saveData()
-      .then(() => this.setState({ success: true }))
-      .catch(() => this.setState({ success: false }))
-      .then(() => setTimeout(() => this.setState({ isSaving: false, success: null }), 1000));
+    const exec = () => {
+      setIsSaving(true);
+      saveData();
```
elohanlon, line 15: I could be wrong, but is saveData usually an asynchronous function that returns a promise?  Meaning that the two lines below this one will run immediately after the function fires and not wait until the async process completes?

app/javascript/hyacinth_v1/components/shared/forms/buttons/SubmitButton.js
```diff
@@ -2,17 +2,15 @@ import React from 'react';
...
+  // switch to functional components + hooks
+  const { formType, staticContext, ...rest } = props;
+  return (
+    <Button variant="primary" type="submit" {...rest}>
+      {formType === 'new' ? 'Create' : 'Update'}
+    </Button>
+  );
+};
 
 export default withRouter(SubmitButton);
```
elohanlon, line 16: Same question here about whether withRouter should still be used now that this is a functional component.  On this branch, this `withRouter ` occurrence and the other one I mentioned above are the only ones that are left in the app.  I think okay to remove?

app/models/dynamic_field_category.rb
```diff
@@ -10,14 +10,17 @@ class DynamicFieldCategory < ActiveRecord::Base
...
-  def as_json(_options = {})
-    {
+  def as_json(options = {})
+    json = {
       id: id,
       type: self.class.name,
       display_label: display_label,
       sort_order: sort_order,
-      children: children
+      children: children.map(&:as_json)
```
elohanlon, line 19: I thought that nested as_json calls worked in such a way as to automatically call as_json on each level of discovered nested object, so you may not need to explicitly call as_json on the children here.  Might be worth a check.

app/models/dynamic_field.rb
```diff
@@ -32,22 +38,12 @@ module Type
...
-      is_keyword_searchable: is_keyword_searchable,
-      is_title_searchable: is_title_searchable,
-      is_identifier_searchable: is_identifier_searchable,
-      controlled_vocabulary: controlled_vocabulary
-    }
+  def as_json(options = {})
+    json = EXPORTABLE_ATTRIBUTES.map { |k| [k, self.send(k)] }.to_h
+    json[:type] = self.class.name
+    return json unless options[:camelize]
+    json.deep_transform_keys! { |key| key.to_s.camelize(:lower).to_sym }
```
elohanlon, line 45: I like the way you implemented this with the `options[:camelize]` option. 👍

spec/requests/graphql/mutations/digital_object/create_digital_object_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+          }
+        ]
+      }]
+    }
+  end
+
+  before do
+    Hyacinth::DynamicFieldsLoader.load_fields!(field_definitions)
+    search_params = { digital_object_type: 'item' }
+    EnabledDynamicField.where(search_params).delete_all
```
elohanlon, line 30: It's not clear to me why the above `delete_all` line runs before each test.  Wouldn't the db be cleared before each test anyway?

spec/requests/graphql/mutations/digital_object/create_digital_object_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+      let(:expected_identifiers) do
+        variables[:input][:identifiers]
+      end
+
+      before do
+        sign_in_project_contributor to: :create_objects, project: project
+        graphql query, variables
+      end
+
+      it "return a single item with the expected metadata fields" do
```
elohanlon, line 78:
```suggestion
      it "returns a single item with the expected metadata fields" do
```

spec/requests/graphql/mutations/digital_object/update_descriptive_metadata_mutation_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+          }
+        ]
+      }]
+    }
+  end
+
+  before do
+    Hyacinth::DynamicFieldsLoader.load_fields!(field_definitions)
+    search_params = { digital_object_type: 'item' }
+    EnabledDynamicField.where(search_params).delete_all
```
elohanlon, line 31: Same question here about why `delete_all` is necessary.

spec/requests/graphql/query/enabled_dynamic_fields_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+  let!(:enabled_dynamic_field) do
+    FactoryBot.create(:enabled_dynamic_field, project: project, digital_object_type: digital_object_type,
+      dynamic_field: dynamic_field, required: false, default_value: default_value, field_sets: [field_set])
+  end
+
+  include_examples 'requires user to have correct permissions for graphql request' do
+    let(:request) { graphql query(project.string_key, digital_object_type) }
+  end
+
+  context 'when logged in user is an administrator' do
```
elohanlon, line 20: It might be more useful to change the "administrator" check to "project administrator" or "user who has project permission" because administrators have access to everything all the time.  But if this is just a generic check that's verifying that "someone who has access" gets the expected the response, then maybe that doesn't matter.  Best case scenario, a lower level user check is more likely to catch a permission bug.

app/graphql/mutations/digital_object/create_digital_object.rb
barmintor: Ah, this is an artifact from when I was assigning identifiers separately.

app/graphql/types/query_type.rb
barmintor: I think something like that is a good idea, but I'm not sure what class to put it on or what to call it - when I start thinking about the naming is awkward. `EnabledDynamicField.or_not_for_project` or `DynamicField.with_enabling_if_enabled`

app/javascript/hyacinth_v1/components/shared/forms/FormButtons.js
barmintor: It's not - it's debugging cruft.

app/javascript/hyacinth_v1/components/shared/forms/buttons/ResponsiveSubmitButton.js
barmintor: I don't think it is - this is a place where we could use some js testing :-(

app/models/dynamic_field_category.rb
barmintor: I don't think they do in the way you're talking about - I think that rubygraphql calls as_json down the line if it has a relevant type, but if you want to ensure it is called for a JSON scalar in this context you have to do it manually.

spec/requests/graphql/mutations/digital_object/create_digital_object_spec.rb
barmintor: Yep.

app/javascript/hyacinth_v1/components/digital_objects/metadata/MetadataForm.js
barmintor, line 135:
```suggestion
        // if the ids are fetched in queries mixing ID scalars and JSON, strings must be compared
```

app/javascript/hyacinth_v1/components/digital_objects/metadata/MetadataForm.js
barmintor:
```suggestion
```

app/javascript/hyacinth_v1/components/shared/forms/FormButtons.js
barmintor:
```suggestion
        <ResponsiveSubmitButton saveData={onSave} formType={formType} />
```

spec/requests/graphql/mutations/digital_object/create_digital_object_spec.rb
barmintor:
```suggestion
```

spec/requests/graphql/mutations/digital_object/update_descriptive_metadata_mutation_spec.rb
barmintor:
```suggestion
```

spec/requests/graphql/query/enabled_dynamic_fields_spec.rb
barmintor: probably a good guiding principle to test at the lowest level of access expected to work.

app/javascript/hyacinth_v1/components/digital_objects/new/DigitalObjectNewForm.js
barmintor: Possibly - need to review what's in the stub object and how the new/edit modes differ.

app/javascript/hyacinth_v1/components/digital_objects/metadata/MetadataForm.js
barmintor: Added a word :-)

app/javascript/hyacinth_v1/components/shared/forms/buttons/ResponsiveSubmitButton.js
barmintor: ok, removed.

app/javascript/hyacinth_v1/components/digital_objects/new/DigitalObjectNewForm.js
barmintor: Removed redundant parameters.

app/javascript/hyacinth_v1/components/shared/forms/buttons/ResponsiveSubmitButton.js
barmintor: Restored asynch processing and commented on halting state updates to prevent memory leaks.

app/javascript/hyacinth_v1/components/shared/forms/buttons/SubmitButton.js
barmintor: Removed.

spec/requests/graphql/query/enabled_dynamic_fields_spec.rb
barmintor: lowered privileges used in spec to project controbutor

app/javascript/hyacinth_v1/components/digital_objects/metadata/MetadataForm.js
elohanlon: Thanks! :-)

app/javascript/hyacinth_v1/components/shared/forms/buttons/ResponsiveSubmitButton.js
elohanlon: 👍

app/models/dynamic_field_category.rb
elohanlon: Ah!  You're right.  I just tried it and confirmed.

spec/requests/graphql/query/enabled_dynamic_fields_spec.rb
elohanlon: 👍

app/models/dynamic_field_category.rb
barmintor: yeah, sad times.

app/graphql/inputs/enabled_dynamic_field_input.rb
elohanlon: Ticket created!

app/graphql/types/query_type.rb
barmintor: opened HYACINTH-646 to track this refactor.

elohanlon: Thanks!

PR #286, opened 2020-08-31
HYACINTH-590

elohanlon:
- Updates the DigitalObjects query result type name
- Fixes a typo in hyacinth.template.yml


elohanlon:
Thank you for approving!  Merging now!

PR #288, opened 2020-09-01
Hyacinth 564

JackBlackLight:


JackBlackLight:
I think this meets the ticket requirements. The redirectToSearch module represents the code that would have been duplicated in DigitalObjectSearch and DigitalObjectInterface.

app/javascript/hyacinth_v1/components/digital_objects/DigitalObjectInterface.js
```diff
@@ -1,22 +1,53 @@
...
+  const backToSearchHandler = () => {
+    const latestSearchQuery = JSON.parse(latestSearchQueryString);
+    redirectToSearch(history, latestSearchQuery);
+  };
+
+  const backToSearchButton = () => {
+    const button = latestSearchQueryString ?
+          <div>
+                <Button
+                  variant="Dark"
```
elohanlon, line 30:
```suggestion
                  variant="dark"
```

app/javascript/hyacinth_v1/components/digital_objects/DigitalObjectInterface.js
```diff
@@ -1,22 +1,53 @@
...
+  const latestSearchQueryString = sessionStorage.getItem('searchQueryParams');
+
+
+  const backToSearchHandler = () => {
+    const latestSearchQuery = JSON.parse(latestSearchQueryString);
+    redirectToSearch(history, latestSearchQuery);
+  };
+
+  const backToSearchButton = () => {
+    const button = latestSearchQueryString ?
```
elohanlon, line 27: I'd recommend directly returning the output of the ternary operator rather than assigning to local variable `button`.

app/javascript/hyacinth_v1/components/digital_objects/DigitalObjectInterface.js
```diff
@@ -1,22 +1,53 @@
...
+import { Button } from 'react-bootstrap';
 import DigitalObjectSummary from './DigitalObjectSummary';
+import { redirectToSearch } from '../../utils/redirectToSearch';
 
 function DigitalObjectInterface(props) {
   const { digitalObject, children } = props;
   const { id, title, digitalObjectType } = digitalObject;
+  const history = useHistory();
+  const latestSearchQueryString = sessionStorage.getItem('searchQueryParams');
+
```
elohanlon, line 19: Very minor style thing, but could you remove some of the extra blank lines in this file?

app/javascript/hyacinth_v1/components/digital_objects/DigitalObjectInterface.js
elohanlon: Actually, for semantic html reasons (and for proper accessibility, and for consistent UI styling and hover state in the navbar) we should use an <a> tag here instead of a button because it leads to a different URL.  I was about to suggest just using an <a> tag here in the JSX, but just learned that the react bootstrap library automatically renders an <a> if you include an href attribute on the button (https://react-bootstrap.github.io/components/buttons/#button-tags).  This would mean an additional change to the `backToSearchHandler` situation: you'd want a method in the same module that actually returns the new url (e.g. `backToSearchPath()`) rather than performing the history update.

And then you would change the ContextualNavbar render to:
```
<ContextualNavbar
        title="Create Batch Import"
        rightHandLinks={[
          { link: backToSearchPath, label: 'Back to Search' },
        ]}
      />
```

And the `backToSearchButton` method can go away altogether.

app/javascript/hyacinth_v1/components/digital_objects/DigitalObjectList.js
```diff
@@ -10,16 +10,39 @@ import { startCase } from 'lodash';
   to display a list of digital objects.
 */
 
+
+const storeSearchQueryParams =  (orderBy, totalCount, perPage, offset, pageNumber, q, filters) => {
+  if (q || filters.length > 0) {
+    window.sessionStorage.setItem('searchQueryParams',
+    JSON.stringify(
```
elohanlon, line 17: We should set you up with in-IDE js linting because there a couple of js linter errors that I'm seeing when I run this branch in my local copy.  We're using eslint, which is like rubocop for javascript.  Our eslint settings come from this file: https://github.com/cul/ldpd-hyacinth/blob/master/.eslintrc

elohanlon (changes requested):
I have a few suggested changes, but this is good!  Thank you for adding this feature!

app/javascript/hyacinth_v1/utils/redirectToSearch.js
```diff
@@ -0,0 +1,19 @@
+import FilterArrayParam from './filterArrayParam';
+import * as qs from 'query-string';
+import {
+  NumberParam, StringParam, withDefault, encodeQueryParams
+} from 'use-query-params';
+
+export const queryParamsConfig = {
```
elohanlon, line 7: I tested this branch locally and I'm seeing warnings in the js console like `encodeQueryParams.js:18 Encoding parameter totalCount as string since it was not configured.`.  You should be able to see the same thing if you open the js console after doing a search and clicking on a digital object show page.

I think this was a problem with the code before your changes, so nothing to do with your new code, but would you mind updating `queryParamsConfig` in `redirectToSearch.js` so that it covers all of the params being passed in here to `const queryParams`?  That should hopefully get rid of the console warnings.

JackBlackLight: One remaining parameter, 'query', triggers this warning. I need to test it out, but I think this could be resolved if searchParams used the key 'q' rather than 'query'.

elohanlon: Sounds good!  Also, the latest commits you added look good.  Feel free to re-request a review from me on this PR when you're finished with the q/query change and I'll check it out.

JackBlackLight: The q/query issue seems to have required only a change in one spot. I think we're done here.

elohanlon: 👍

elohanlon:
@JackBlackLight I was just testing this locally and I discovered a small issue.  If you go to `/ui/v1/digital_objects` and then click on a search result WITHOUT applying a query or any facet filters, it doesn't show the "Back To Search" link on the resulting page.

Going to `/ui/v1/digital_objects` and then changing the per-page count and the sort field -- but once again not choosing a query or facet filter -- also results in a missing "Back To Search" link on the resulting page.

JackBlackLight:
@elohanlon, great catch. I will correct that and push commits tomorrow. Thanks!

JackBlackLight:
@elohanlon, my latest commit addresses coming from search without query or facet params.

elohanlon:
@JackBlackLight Thanks!  I just tested out the above case again.  This fixed the previous issue, but I just found a new one.  I added a suggestion comment to the PR that should take care of it.  Here's what was happening:

If you go to `/ui/v1/digital_objects?pageNumber=3&perPage=2` and then click on a result and press "Back to search", you get:
`/ui/v1/digital_objects?offset=4&orderBy=TITLE%20ASC&pageNumber=3&perPage=2&totalCount=11`

The `offset` and `totalCount`params aren't user-facing ones, so we'll want to remove those before generating the "Back to Search" url.  The orderBy is fine to keep in there because it's a user-facing one and just represents the default search mode.

I'll approve this PR in advance of your review of my suggestion comment.  Thanks for your work on this!

app/javascript/hyacinth_v1/components/digital_objects/DigitalObjectInterface.js
```diff
@@ -7,15 +7,30 @@ import Tabs from '../shared/tabs/Tabs';
...
 import DigitalObjectSummary from './DigitalObjectSummary';
+import { encodeAndStringifySearch } from '../../utils/encodeAndStringifySearch';
 
 function DigitalObjectInterface(props) {
   const { digitalObject, children } = props;
   const { id, title, digitalObjectType } = digitalObject;
+  const latestSearchQueryString = sessionStorage.getItem('searchQueryParams');
+
+  const backToSearchPath = () => {
+    const latestSearchQuery = JSON.parse(latestSearchQueryString);
```
elohanlon, line 18:
```suggestion
    const latestSearchQuery = JSON.parse(latestSearchQueryString);
    
    // Delete a couple of search parameters that shouldn't appear in a user-facing search url
    delete latestSearchQuery.offset;
    delete latestSearchQuery.totalCount;
```

elohanlon (approved):

PR #289, opened 2020-09-02
obscure publish target API token when user is not admin (HYACINTH-454)

barmintor:


app/graphql/types/publish_target_type.rb
```diff
@@ -2,6 +2,8 @@
 
 module Types
   class PublishTargetType < Types::BaseObject
+    OBSCURED_API_KEY = "****************"
```
elohanlon, line 5: Oh, I get it.  Like that Molly Ringwald movie "Sixteen Asterisks."

spec/requests/graphql/query/publish_target_spec.rb
```diff
@@ -12,7 +12,32 @@
...
+              "doiPriority": 100,
+              "isAllowedDoiTarget": false
+            }
+          }
+        })).at_path('data')
+      end
+    end
+  end
+
+  context 'when logged in user has read permissions' do
```
elohanlon, line 40:
```suggestion
  context 'when logged in user is not an admin, but has read permissions' do
```
(Optional suggestion -- feel free to take it or leave it.)

elohanlon (approved):
Looks great!  I only had one optional suggestion, but feel free to merge any time with or without it. 👍

PR #290, opened 2020-09-03
Add typeOfSearch to Types::SearchAttributes (HYACINTH-581)

barmintor:
Opening a draft PR while I work on HYACINTH-582

### DIFF
--- a/app/graphql/enums/filter_match_enum.rb
+++ b/app/graphql/enums/filter_match_enum.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class Enums::FilterMatchEnum < Types::BaseEnum
+  value 'ABSENT', 'absent (negated present)', value: 'absent'
+  value 'CONTAINS', 'contains', value: 'contains'
+  value 'MATCHES', 'matches', value: 'matches'
+  value 'OMITS', 'omits (negated contains)', value: 'omits'
+  value 'PRESENT', 'present', value: 'present'
+  value 'VARIES', 'varies (negated matches)', value: 'varies'
+end
--- a/app/graphql/enums/filter_match_enum.rb
+++ b/app/graphql/enums/filter_match_enum.rb
@@ -1,10 +1,12 @@
 # frozen_string_literal: true
 
 class Enums::FilterMatchEnum < Types::BaseEnum
-  value 'ABSENT', 'absent (negated present)', value: 'absent'
-  value 'CONTAINS', 'contains', value: 'contains'
-  value 'MATCHES', 'matches', value: 'matches'
-  value 'OMITS', 'omits (negated contains)', value: 'omits'
-  value 'PRESENT', 'present', value: 'present'
-  value 'VARIES', 'varies (negated matches)', value: 'varies'
+  value 'CONTAINS', 'contains', value: 'CONTAINS'
+  value 'DOES_NOT_CONTAIN', 'does not contain', value: 'DOES_NOT_CONTAIN'
+  value 'DOES_NOT_EQUAL', 'does not equal', value: 'DOES_NOT_EQUAL'
+  value 'DOES_NOT_EXIST', 'does not exist', value: 'DOES_NOT_EXIST'
+  value 'DOES_NOT_START_WITH', 'does not start with', value: 'DOES_NOT_START_WITH'
+  value 'EQUALS', 'equals', value: 'EQUALS'
+  value 'EXISTS', 'exists', value: 'EXISTS'
+  value 'STARTS_WITH', 'starts with', value: 'STARTS_WITH'
 end
--- a/lib/solr/params.rb
+++ b/lib/solr/params.rb
@@ -10,8 +10,10 @@ module Solr
       'DOES_NOT_CONTAIN' => { field_template: '-%s:(%s)', value_template: '*%s*' },
       'DOES_NOT_EQUAL' => { field_template: '-%s:(%s)' },
       'DOES_NOT_EXIST' => { field_template: '-%s:%s', value: '*' },
+      'DOES_NOT_START_WITH' => { field_template: '-%s:(%s)', value_template: '%s*' },
       'EQUALS' => { field_template: '%s:(%s)' },
-      'EXISTS' => { field_template: '%s:%s', value: '*' }
+      'EXISTS' => { field_template: '%s:%s', value: '*' },
+      'STARTS_WITH' => { field_template: '%s:(%s)', value_template: '%s*' }
     }.freeze
     VALID_SORT_DIRECTION = ['asc', 'desc'].freeze
 
--- a/lib/solr/params.rb
+++ b/lib/solr/params.rb
@@ -4,6 +4,15 @@ module Solr
   class Params
     attr_reader :parameters
 
+    VALID_BOOLEAN_OPERATORS = { and: ' AND ', or: ' OR ' }.freeze
+    VALID_FILTER_MATCHES = {
+      'absent' => { field_template: '-%s:%s', value: '*' },
+      'contains' => { field_template: '%s:(%s)', value_template: '*%s*' },
+      'matches' => { field_template: '%s:(%s)' },
+      'omits' => { field_template: '-%s:(%s)', value_template: '*%s*' },
+      'present' => { field_template: '%s:%s', value: '*' },
+      'varies' => { field_template: '-%s:(%s)' }
+    }.freeze
     VALID_SORT_DIRECTION = ['asc', 'desc'].freeze
 
     def initialize
@@ -16,10 +25,13 @@ module Solr
       }
     end
 
-    def fq(field, values, boolean_operator = :or)
-      raise ArgumentError unless [:and, :or].include?(boolean_operator)
-      escaped_values = Array.wrap(values).map { |value| Solr::Utils.escape(value).gsub(' ', '\ ') }
-      @parameters[:fq] << "#{field}:(#{escaped_values.join(boolean_operator == :and ? ' AND ' : ' OR ')})" unless escaped_values.blank?
+    def fq(field, value_or_values, match_operator = 'matches', boolean_operator = :or)
+      raise(ArgumentError, "Invalid match operator: #{match_operator}") unless VALID_FILTER_MATCHES.include?(match_operator)
+      raise(ArgumentError, "Invalid boolean operator: #{boolean_operator}") unless VALID_BOOLEAN_OPERATORS.include?(boolean_operator)
+      escaped_values = Array.wrap(value_or_values).map { |value| Solr::Utils.escape(value).gsub(' ', '\ ') }
+      filter_data = VALID_FILTER_MATCHES[match_operator]
+      value = filter_data[:value] || escaped_values.map { |ev| filter_data[:value_template]&.%(ev) || ev }.join(VALID_BOOLEAN_OPERATORS[boolean_operator])
+      @parameters[:fq] << format(filter_data[:field_template], field, value) unless value.blank?
       self
     end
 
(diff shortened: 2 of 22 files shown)


barmintor:
Ok, I think it's ready to discuss!

app/graphql/enums/filter_match_enum.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class Enums::FilterMatchEnum < Types::BaseEnum
+  value 'ABSENT', 'absent (negated present)', value: 'absent'
+  value 'CONTAINS', 'contains', value: 'contains'
+  value 'MATCHES', 'matches', value: 'matches'
+  value 'OMITS', 'omits (negated contains)', value: 'omits'
```
elohanlon, line 7: That's a cool addition.

app/graphql/enums/filter_match_enum.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class Enums::FilterMatchEnum < Types::BaseEnum
+  value 'ABSENT', 'absent (negated present)', value: 'absent'
+  value 'CONTAINS', 'contains', value: 'contains'
+  value 'MATCHES', 'matches', value: 'matches'
```
elohanlon, line 6: Can we stick with Hyacinth 2's "EQUALS" instead of changing it to "MATCHES"?  "MATCHES" makes me think we're supporting regular expressions.

app/graphql/enums/filter_match_enum.rb
```diff
@@ -0,0 +1,10 @@
...
+  value 'OMITS', 'omits (negated contains)', value: 'omits'
+  value 'PRESENT', 'present', value: 'present'
+  value 'VARIES', 'varies (negated matches)', value: 'varies'
```
elohanlon, line 9: In Hyacinth 2, we call this "DOES_NOT_EQUAL".  Can we keep that term?  I feel like "varies" is a confusing term to pick in the custom filter selector (e.g. "original file path > varies > carnegie")

app/graphql/enums/filter_match_enum.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class Enums::FilterMatchEnum < Types::BaseEnum
```
elohanlon, line 3: Thank you for going above and beyond the minimum requirements for HYACINTH-582 with all of these filters!  It's great having so many types.  If (and only if) it would be easy to do, would you be okay with throwing 'STARTS_WITH' in here too, since that's the only HYC2 one that's missing at this point?

app/graphql/enums/filter_match_enum.rb
barmintor: I struggled with this in my head, since this enum is kind of a mish-mash of verbs, adjectives and phrases 😟 I don't know that the user-facing labels have to be the same as the enum values, but sure - especially if we use 'equals' instead of 'matches'.

barmintor: so `EQUALS/DOES_NOT_EQUAL`, `CONTAINS/DOES_NOT_CONTAIN`, `STARTS_WITH/DOES_NOT_START_WITH`, `HAS_A_VALUE/DOES_NOT_HAVE_A_VALUE`?

elohanlon: Thanks!  That sounds good!  DOES_NOT_START_WITH isn't required, but if it's easy to add I'm good with having it in there too.

I could live with `HAS_A_VALUE/DOES_NOT_HAVE_A_VALUE`, but I have a slight preference for `IS_PRESENT/IS_ABSENT` because it might be less error prone for API users:

"collection > has a value > [am I allowed to match a value here?]"
vs.
"collection > is present > [hopefully wouldn't make as much sense to put anything after 'is present'?]"

In the end though, it's probably just personal preference and being used to the HYC2 "present" vs. "absent" terms.

spec/solr/params_spec.rb
```diff
@@ -11,14 +11,14 @@
...
       params.fq('animals', ['dogs + cats', 'llamas', 'alpacas'])
       expect(params.to_h).to include(fq: ['animals:(dogs\ \+\ cats OR llamas OR alpacas)'])
     end
 
-    it 'properly handles and escapes multiple values with the :and param passed, AND-ing them together' do
-      params.fq('animals', ['dogs + cats', 'llamas', 'alpacas'], :and)
-      expect(params.to_h).to include(fq: ['animals:(dogs\ \+\ cats AND llamas AND alpacas)'])
+    it 'properly handles and escapes multiple values with the match param passed, OR-ing them together' do
+      params.fq('animals', ['dogs + cats', 'llamas', 'alpacas'], 'contains')
+      expect(params.to_h).to include(fq: ['animals:(*dogs\ \+\ cats* OR *llamas* OR *alpacas*)'])
```
elohanlon, line 21: With this change, it seems like we're no longer testing our AND-ing ability anywhere.

spec/solr/params_spec.rb
```diff
@@ -11,14 +11,14 @@
       expect(params.to_h).to include(fq: ['animals:(dogs\ \+\ cats)'])
     end
 
-    it 'properly handles and escapes multiple values and defaults to OR-ing them together' do
+    it 'properly handles and escapes multiple values and defaults to match function' do
```
elohanlon, line 14:
```suggestion
    it 'properly handles and escapes multiple values and defaults to OR-ing them together with equals comparison' do
```
Maybe something like that?

spec/solr/params_spec.rb
barmintor: Remember when I was asking about multiple values and filters? I don't think there is an API for it, I think it only existed with the tests. Am I wrong?

elohanlon: Right.  Looking at the code here reminded me that we added that to support a future feature, but no current API exists.

elohanlon: "Find me all items in collection x OR project y"
vs.
"Find me all items with name a AND name b"

app/graphql/enums/filter_match_enum.rb
barmintor: `PRESENTS/DOES_NOT_PRESENT` 😬

barmintor: Mostly I just want it to be clear what the negated matches are, and what they're negating.

elohanlon: AHH! Anything but `PRESENTS/DOES_NOT_PRESENT`!  Haha.

👍  for negated match clarity.

Maybe `IS_PRESENT/IS_NOT_PRESENT`?

spec/solr/params_spec.rb
```diff
@@ -11,14 +11,14 @@
       expect(params.to_h).to include(fq: ['animals:(dogs\ \+\ cats)'])
     end
 
-    it 'properly handles and escapes multiple values and defaults to OR-ing them together' do
+    it 'properly handles and escapes multiple values and defaults to OR-ing them together with equals comparison' do
```
elohanlon, line 14:
```suggestion
    it 'properly handles and escapes multiple values with default equals comparison, OR-ing them together' do
```

spec/solr/params_spec.rb
```diff
@@ -11,14 +11,14 @@
...
       expect(params.to_h).to include(fq: ['animals:(dogs\ \+\ cats OR llamas OR alpacas)'])
     end
 
-    it 'properly handles and escapes multiple values with the :and param passed, AND-ing them together' do
-      params.fq('animals', ['dogs + cats', 'llamas', 'alpacas'], :and)
-      expect(params.to_h).to include(fq: ['animals:(dogs\ \+\ cats AND llamas AND alpacas)'])
+    it 'properly handles and escapes multiple values with the match param passed, OR-ing them together' do
+      params.fq('animals', ['dogs + cats', 'llamas', 'alpacas'], 'CONTAINS')
+      expect(params.to_h).to include(fq: ['animals:(*dogs\ \+\ cats* OR *llamas* OR *alpacas*)'])
     end
```
elohanlon, line 22:
```suggestion
    it 'properly handles and escapes multiple values with the match param passed, OR-ing them together' do
      params.fq('animals', ['dogs + cats', 'llamas', 'alpacas'], 'CONTAINS')
      expect(params.to_h).to include(fq: ['animals:(*dogs\ \+\ cats* OR *llamas* OR *alpacas*)'])
    end

    it 'allows the same fq key to be added twice, thereby AND-ing across the separate fqs' do
      params.fq('animals', ['cool cats', 'hot dogs'], 'CONTAINS')
      params.fq('animals', ['cool llamas', 'hot alpacas'], 'CONTAINS')
      expect(params.to_h).to include(fq: ['animals:(*cool\ cats* OR *hot\ dogs*)', 'animals:(*cool\ llamas* OR *hot\ alpacas*)'])
    end
```

elohanlon, line 22:
```suggestion
    end
    
    it 'allows the same fq key to be added twice, thereby AND-ing across the separate fqs' do
      params.fq('animals', ['cool cats', 'hot dogs'], 'CONTAINS')
      params.fq('animals', ['cool llamas', 'hot alpacas'], 'CONTAINS')
      expect(params.to_h).to include(fq: ['animals:(*cool\ cats* OR *hot\ dogs*)', 'animals:(*cool\ llamas* OR *hot\ alpacas*)'])
    end
```

spec/hyacinth/adapters/digital_object_search_adapter/solr_spec.rb
```diff
@@ -10,10 +10,50 @@
 
   context "#solr_params_for" do
     let(:solr_params) { adapter.solr_params_for(search_params) }
-    context "with multiple filter values on the same filter" do
-      let(:search_params) { { 'animals' => ['dogs', 'cats'] } }
+    context "with multiple filters on the same field" do
+      let(:search_params) do
+        {
+          'animals' => [
+            ['aardvarks', 'DOES_NOT_EXIST'], ['celocanths', 'CONTAINS'], ['elephants', 'EQUALS'],
```
elohanlon, line 17: `['aardvarks', 'DOES_NOT_EXIST']` and `['panthers', 'EXISTS']`

I wasn't expecting to see any search values associated with these match types.

elohanlon: Is this test meant to demonstrate that if you supply a value that it will be ignored by these two match types?  This might be confusing for someone looking at the tests as usage examples.

lib/solr/params.rb
```diff
@@ -16,15 +27,19 @@ def initialize
       }
     end
 
-    def fq(field, values, boolean_operator = :or)
-      raise ArgumentError unless [:and, :or].include?(boolean_operator)
-      escaped_values = Array.wrap(values).map { |value| Solr::Utils.escape(value).gsub(' ', '\ ') }
-      @parameters[:fq] << "#{field}:(#{escaped_values.join(boolean_operator == :and ? ' AND ' : ' OR ')})" unless escaped_values.blank?
+    def fq(field, value_or_values, match_operator = 'EQUALS', boolean_operator = :or)
```
elohanlon, line 30: To confirm: Since this `boolean_operator` param (plus internal method logic) is still here, does that mean that we are still supporting the possibility of passing AND to the `fq` method?  The reason I ask is because that `params.fq('animals', ['dogs + cats', 'llamas', 'alpacas'], :and)` test was deleted and we talked about supporting AND-ing through the addition of multiple, separate fqs.

We should either delete the boolean_operator param or re-add the deleted method to maintain test coverage.
 I'm good to go with whichever way you'd prefer.

elohanlon (changes requested):
@barmintor Thank you for all of the updates.  I have a couple of minor questions and a couple of suggestions, but otherwise this looks good!

elohanlon (approved):

PR #291, opened 2020-09-30
HYACINTH-534: Add export filter functionality to limit which fields are exported during batch export

elohanlon:


elohanlon:
This PR applies default filters to our current batch export function and paves the way for eventually allowing users to specify their own inclusion and exclusion filters when we build user-controlled filtering into the API and UI.

lib/hyacinth/jobs/batch_export_job/export_filter.rb
```diff
@@ -0,0 +1,102 @@
...
+    module BatchExportJob
+      class ExportFilter
+        DEFAULT_CORE_INCLUSION_FILTERS = [
+          '_uid',
+          '_asset_type',
+          '_digital_object_type',
+          '_doi',
+          /_identifiers\[\d+\]/,
+          /_parent_digital_objects\[\d+\]\..+/,
+          '_primary_project.string_key',
```
elohanlon, line 14: Just a quick note here to say that we'll definitely be updating the DEFAULT_CORE_INCLUSION_FILTERS at some point, but this represents a basic initial set.  We can talk with the Hyacinth User Group to figure out which additional fields would be useful to include in batch exports in the future.

elohanlon:
@barmintor I just added a couple of additional commits to address a couple of other related tickets.  Ready for review any time!

lib/hyacinth/jobs/batch_export_job/export_filter.rb
```diff
@@ -0,0 +1,102 @@
...
+          descriptive_field_group_string_keys.map { |string_key| /#{string_key}\[\d+\]\..+/ }
+        end
+
+        # Given an input batch export file, generates an output batch export file that only includes
+        # headers that match the configured inclusion and exclusion filters.
+        def generate_filtered_export(input_csv_file_path, output_csv_file_path)
+          ordered_indexes_to_keep = nil
+          CSV.open(output_csv_file_path, 'wb') do |output_csv|
+            CSV.foreach(input_csv_file_path, 'rb') do |row|
+              if ordered_indexes_to_keep.nil?
```
JackBlackLight, line 80: @elohanlon, it looks like ordered_indexes_to_keep will always be nil here.

elohanlon: @JackBlackLight, it's nil the first time that the foreach loop runs, but will be assigned a value during that first iteration and that value will be used during every following iteration.

JackBlackLight (approved):

PR #297, opened 2020-12-10
verify that DynamicFieldGroup path collisions are rescued and correct…

barmintor:
…able (HYACINTH-568)

- rescue ActiveRecord::RecordNotUnique
- indicate in errors hash when DynamicFieldGroup path uniqueness is violated

### DIFF
--- a/app/models/concerns/dynamic_field_structure/string_key.rb
+++ b/app/models/concerns/dynamic_field_structure/string_key.rb
@@ -17,7 +17,12 @@ module DynamicFieldStructure
       # Validate that the string_key is unique between all siblings.
       def unique_string_key
         sibling_string_keys = siblings.map(&:string_key)
-        errors.add(:string_key, 'is already in use by a sibling field or field group') if sibling_string_keys.include?(string_key)
+        path_error if sibling_string_keys.include?(string_key)
+      end
+
+      def path_error
+        errors.add(:string_key, 'is already in use by a sibling field or field group')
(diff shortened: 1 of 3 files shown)


app/models/concerns/dynamic_field_structure/string_key.rb
```diff
@@ -17,7 +17,12 @@ module StringKey
       # Validate that the string_key is unique between all siblings.
       def unique_string_key
```
elohanlon, line 18: @barmintor If we're relying on a unique constraint to validate this rare (but important to protect against) error, and handling it for error display, I'm now wondering if we even need to keep `validate :unique_string_key` or the `unique_string_key` method at all.

app/models/dynamic_field_group.rb
```diff
@@ -29,6 +29,11 @@ class DynamicFieldGroup < ActiveRecord::Base
   validate  :non_circular_relationship
   validates_with StringKey::AbsentInSiblingFieldsValidator
 
+  def save
```
elohanlon, line 32: @barmintor I had another idea for how we could handle HYACINTH-568.  I'm going to open a PR to this branch in a sec so you can check it out and see what you think.  It involves two extra SQL queries, but since we don't update field definitions too often I'm hoping it doesn't seem too bad, and it allows us to avoid overriding `save`.

elohanlon (approved):

PR #299, opened 2020-12-15
Update unique_string_key to ensure unique top level string_key values across all immediate descendants of DynamicFieldCategories

elohanlon:


spec/models/dynamic_field_group_spec.rb
```diff
@@ -160,14 +160,7 @@
     end
 
     context 'when creating a group that shares the same string_key as a non-sibling group' do
-      let(:dynamic_field_group) { FactoryBot.create(:dynamic_field_group) }
-      let(:dynamic_field) { FactoryBot.build(:dynamic_field, string_key: 'name', dynamic_field_group: dynamic_field_group) }
```
elohanlon, line 164: I think this was copied from another test.  It didn't seem to be testing anything relevant to this group, so that's why I removed it.  (We can re-add if I'm mistaken!)

spec/models/dynamic_field_group_spec.rb
```diff
@@ -358,12 +351,13 @@
       let(:dynamic_field) { FactoryBot.create(:dynamic_field, dynamic_field_group: dynamic_field_group, display_label: "Dynamic Field Display Label") }
       it 'updates the path' do
         dynamic_field_group.string_key = new_string_key
-        dynamic_field_group.save && dynamic_field_group.reload
+        expect(dynamic_field_group.save).to eq(true)
```
elohanlon, line 354: I added a few of these `expect(dynamic_field_group.save).to eq(true)` lines below because they helped me more quickly narrow down what was failing later in the tests, when they briefly failed as I was updating code.

barmintor (approved):
Approved!

PR #300, opened 2021-01-13
Hyacinth 682 js error navigating parent child

JackBlackLight:
2 related fixes here: a filter param formatting issue that triggered the ticketed JS error, and a change that suppresses search result paging in the digital object show footer when not applicable.


elohanlon (approved):

elohanlon:
Thank you for the updates, @JackBlackLight !

PR #303, opened 2021-02-09
Resource generation and Derivativo integration

elohanlon:
This PR covers a wide range of related tasks around resource generation and integration with the new Derivativo 2 app (plus a couple of other fixes/improvements along the way).

Closes the following tickets:

HYACINTH-699
HYACINTH-714
HYACINTH-715
HYACINTH-716
HYACINTH-717
HYACINTH-718
HYACINTH-719
HYACINTH-720
HYACINTH-721
HYACINTH-722
HYACINTH-723
HYACINTH-724
HYACINTH-725
HYACINTH-726
HYACINTH-727
HYACINTH-728
HYACINTH-729
HYACINTH-730
HYACINTH-731
HYACINTH-732
HYACINTH-733
HYACINTH-745
HYACINTH-746


app/graphql/mutations/create_asset.rb
```diff
@@ -1,26 +1,30 @@
...
   field :asset, Types::DigitalObject::AssetType, null: true
 
-  def resolve(signed_blob_id:, parent_id:)
+  def resolve(file_location:, parent_id:)
     parent = DigitalObject::Item.find(parent_id)
     ability.authorize! :create_objects, parent.primary_project
-    blob = ActiveStorage::Blob.find_signed(signed_blob_id)
+
+    # At this time, non-admins can only perform blob-based asset creation
+    raise GraphQL::ExecutionError, 'You are only authorized to create assets from ActiveStorage blob uploads.' if !file_location.start_with?('blob://') && ability.cannot?(:manage, :all)
```
barmintor, line 16:
```suggestion
    raise GraphQL::ExecutionError, 'You are only authorized to create assets from ActiveStorage blob uploads.' unless file_location.start_with?('blob://') || ability.can?(:manage, :all)
```

barmintor: I might need more coffee, but the AND'd negated conditions took me a second.

app/graphql/mutations/digital_object/resource/create_resource.rb
```diff
@@ -0,0 +1,56 @@
...
+
+        field :digital_object, Types::DigitalObjectInterface, null: true
+        field :user_errors, [Types::Errors::FieldedInput], null: false
+
+        def resolve(id:, resource_name:, file_location:)
+          digital_object = ::DigitalObject::Base.find(id)
+          ability.authorize! :update, digital_object
+
+          # At this time, non-admins can only perform blob-based asset creation
+          raise GraphQL::ExecutionError, 'You are only authorized to create resources from ActiveStorage blob uploads.' if !file_location.start_with?('blob://') && ability.cannot?(:manage, :all)
```
barmintor, line 19:
```suggestion
          raise GraphQL::ExecutionError, 'You are only authorized to create resources from ActiveStorage blob uploads.' unless file_location.start_with?('blob://') || ability.can?(:manage, :all)
```

app/models/resource_request.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+class ResourceRequest < ApplicationRecord
+  enum status: { pending: 0, in_progress: 1, success: 2, failure: 3, cancelled: 4 }
+  enum job_type: {
```
barmintor, line 5: so one approach to enums like this that you see historically is to define the enum values on bit switches - 0, 1, 2, 4, etc. - so that a given client value can encode multiple enum values (for example, being able to indicate fulltext on office or video access jobs). It might not be relevant here, and it has other implications about the total set of values for the enum, but I thought I would mention it.

lib/hyacinth/digital_object/resource.rb
```diff
@@ -45,6 +45,29 @@ def to_serialized_form
...
+        BestType.pcdm_type.for_mime_type(media_type) == 'Audio' ||
+          BestType.pcdm_type.for_file_name(original_filename) == 'Audio'
+      end
+
+      def pdf?
+        media_type == 'application/pdf' || original_filename.ends_with?('.pdf')
+      end
+
+      def text_or_office_document?
+        media_type.match?(/text|msword|ms-word|officedocument|powerpoint|excel|iwork/)
```
barmintor, line 69: could we get this from BestType returned values?

lib/hyacinth/digital_object/resource_attributes.rb
```diff
@@ -8,11 +8,11 @@ def self.included(base)
       end
 
       def resource_attributes
-        self.class.resource_attributes
+        self.class.send(__method__) # class method name same as instance method
       end
 
       def resource_import_attributes
-        self.class.resource_attributes
```
barmintor, line 15: this is now delegating to a different class method (before both methods delegated to the `resource_attributes` class method), right?

app/graphql/mutations/create_asset.rb
elohanlon: Good suggestion!

app/graphql/mutations/digital_object/resource/create_resource.rb
elohanlon: 👍

lib/hyacinth/digital_object/resource.rb
elohanlon: Good idea!  I checked the BestType pcdm type lookup rules and I'll need to make a couple of minor updates to fully cover the cases here, but that's not hard to do.  I'll update this code now to use BestType and then separately update BestType.

lib/hyacinth/digital_object/resource_attributes.rb
elohanlon: That's correct.  That was actually a mistake before.  The instance method should behave the same way as its same-name class method.

PR #304, opened 2021-02-18
HYACINTH-562 restricts email changes to user managers

JackBlackLight:


app/graphql/mutations/update_user.rb
```diff
@@ -47,6 +44,15 @@ def changing_password?(attributes)
       [:current_password, :password, :password_confirmation].any? { |k| attributes.include?(k) && attributes[k].present? }
     end
 
+    def prune_restricted_values(user, attributes)
+      attributes.delete(:is_admin) unless ability.can?(:manage, :all)
```
elohanlon, line 48:
```suggestion
      # Only administrators can make other users into administrators
      attributes.delete(:is_admin) unless ability.can?(:manage, :all)
```
Just a comment suggestion for clarity, since we have the other helpful comments below.

app/javascript/hyacinth_v1/components/users/UserEdit.js
```diff
@@ -44,6 +44,14 @@ function UserEdit() {
     return activatable;
   };
 
+  const canEditEmail = () => {
+    let editEmail = ability.can('manage', 'all');
+    if (!editEmail) {
+      editEmail = (ability.can('manage', 'User') && !isAdmin);
+    }
+    return editEmail;
+  };
```
elohanlon, line 53: It would be good to double check me on this suggestion, but would this work?

```suggestion
  const canEditEmail = () => {
    return ability.can('manage', 'all') || ability.can('manage', 'User');
  };
```

If you can `manage all` then you're an admin, otherwise only allow if you can `manage User` (which means you're a user admin).  Otherwise this would result in false.

app/javascript/hyacinth_v1/components/users/UserEdit.js
```diff
@@ -44,6 +44,14 @@ function UserEdit() {
     return activatable;
   };
 
+  const canEditEmail = () => {
+    let editEmail = ability.can('manage', 'all');
+    if (!editEmail) {
+      editEmail = (ability.can('manage', 'User') && !isAdmin);
```
JackBlackLight, line 50: const canEditEmail = () => ability.can('manage', 'all') || ability.can('manage', 'User')  && !isAdmin;

app/javascript/hyacinth_v1/components/users/UserEdit.js
JackBlackLight, line 53: const canEditEmail = () => ability.can('manage', 'all') || ability.can('manage', 'User')  && !isAdmin;

app/javascript/hyacinth_v1/components/users/UserEdit.js
```diff
@@ -44,6 +44,14 @@ function UserEdit() {
     return activatable;
   };
 
+  const canEditEmail = () => {
```
JackBlackLight, line 47: const canEditEmail = () => ability.can('manage', 'all') || ability.can('manage', 'User')  && !isAdmin;

app/javascript/hyacinth_v1/components/users/UserEdit.js
JackBlackLight: Confusingly, the !isAdmin value refers to the user record being edited rather than the user who is editing. So this line does what we want:

  const canEditEmail = () => ability.can('manage', 'all') || (ability.can('manage', 'User')  && !isAdmin);

Would it make sense to change the name of that variable, maybe to !isAdminRecord?

elohanlon: Oh, right!  I just took a look at the whole file and I see what you mean.  I'm okay with keeping it as just `isAdmin` because of the way the other top variables refer to the current record rather than the ability of the viewer.  And I like your suggested one-line change if you're okay with that instead of the `if` statement.  (For me, it's slightly faster to read because it's more compact.)

elohanlon: I'll approve this PR now and leave the line change up to you.  I'm good either way.  Thanks for the PR!

elohanlon (approved):

PR #308, opened 2021-03-09
HYACINTH 735

elohanlon:
And HYACINTH-412 because it was also related to error handling.
And HYACINTH-738 because it was related too.

### DIFF
--- a/spec/graphql/hyacinth_schema_spec.rb
+++ b/spec/graphql/hyacinth_schema_spec.rb
@@ -3,7 +3,7 @@
 require 'rails_helper'
 
 RSpec.describe HyacinthSchema, solr: true do
-  subject(:graphql_request) { HyacinthSchema.execute(query, context: context, variables: variables) }
+  let(:graphql_request) { HyacinthSchema.execute(query, context: context, variables: variables) }
   let(:context) { { current_user: FactoryBot.create(:user), ability: ability } }
   let(:ability) do
     abil = double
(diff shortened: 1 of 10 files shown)


spec/graphql/hyacinth_schema_spec.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe HyacinthSchema, solr: true do
+  subject(:graphql_request) { HyacinthSchema.execute(query, context: context, variables: variables) }
```
JackBlackLight, line 6: A question of interest here. I hadn't seen 'subject' in action before. I'm curious about the advantage it confers.

JackBlackLight (approved):
I just had one informational question. I'm approving now.

spec/graphql/hyacinth_schema_spec.rb
elohanlon: Ah!  Good question, and I'm glad you caught that, thanks!

I didn't actually mean to use subject there.  That came from copying that `HyacinthSchema.execute` line from another test file.  In effect, `subject` acts the same here as `let` since it's a named subject, but I think it's actually better to avoid `subject` and go with `let` instead.  I'll push another commit to change that.  I found this StackOveflow post that explains more about `subject` vs. `let` and makes a pretty good case for avoiding the use of subject in most cases: https://stackoverflow.com/a/38459039

elohanlon:
Thanks for the review, @JackBlackLight !

PR #310, opened 2021-03-24
Hyacinth 737 user error messages in digital object forms

JackBlackLight:
@elohanlon, the MetadataForm.js approach to error handling was slightly different from the other forms, and I wasn't able to integrate UserErrors into that code, so I'm handling it in parallel, using ErrorList, as you suggested.


app/javascript/hyacinth_v1/components/digital_objects/metadata/MetadataForm.js
```diff
@@ -59,9 +60,13 @@ function MetadataForm(props) {
   } = digitalObject;
   const [descriptiveMetadata, setDescriptiveMetadata] = useState({});
   const [createDigitalObject, { error: createErrors }] = useMutation(createDigitalObjectMutation);
-  const [updateDescriptiveMetadata, { error: updateErrors }] = useMutation(
+  const [updateDescriptiveMetadata, { data: updateData }] = useMutation(
```
elohanlon, line 63: You're right to add `data: updateData` here, but we still want to keep and make use of `error: updateErrors` in case there are any top-level, unexpected graphql errors raised.

I've been using the term "userErrors" (which you have below, for errors that appear within the top level "data" object) to indicate that they're errors that the user should be able to fix on their own, but it's still useful to show the top level errors so that users can report those error messages to developers if they come up.

app/javascript/hyacinth_v1/components/digital_objects/metadata/MetadataForm.js
```diff
@@ -163,7 +168,7 @@ function MetadataForm(props) {
 
   if (enabledFieldsLoading || fieldGraphLoading) return (<></>);
 
-  const anyErrors = (enabledFieldsError || fieldGraphError || createErrors || updateErrors);
+  const anyErrors = (enabledFieldsError || fieldGraphError || createErrors );
```
elohanlon, line 171: Related to above comment, you'll want to undo this line change.

app/javascript/hyacinth_v1/components/digital_objects/rights/rights_form/ItemRightsForm.js
```diff
@@ -36,7 +36,10 @@ function ItemRightsForm(props) {
 
   const [rights, setRights] = useHash(mergeDefaultValues(fieldConfiguration, initialRights));
 
-  const [updateRights, { error: updateError }] = useMutation(updateRightsMutation);
+  const [updateRights, { data: updateData }] = useMutation(updateRightsMutation);
+
+  // One day, maybe enable optionalChaining JS feature in babel to simplify lines like the one below.
+  const userErrors = (updateData && updateData.updateRights && updateData.updateRights.userErrors) || [];
```
elohanlon, line 42: Same note here about wanting to include `error: updateError` and still show it in with `GraphQLErrors` component (which only renders if updateErrors are actually present).

app/javascript/hyacinth_v1/components/digital_objects/rights/rights_form/AssetRightsForm.js
```diff
@@ -29,7 +29,11 @@ function AssetRightsForm(props) {
 
   const [rights, setRights] = useHash(merge({}, defaultAssetRights, initialRights));
 
-  const [updateRights, { error: updateError }] = useMutation(updateRightsMutation);
+  const [updateRights, { data: updateData }] = useMutation(updateRightsMutation);
```
elohanlon, line 32: Same note here about wanting to include `error: updateError` and still show it in with `GraphQLErrors` component (which only renders if updateErrors are actually present).

app/javascript/hyacinth_v1/components/digital_objects/metadata/MetadataForm.js
```diff
@@ -59,9 +60,13 @@ function MetadataForm(props) {
   } = digitalObject;
   const [descriptiveMetadata, setDescriptiveMetadata] = useState({});
   const [createDigitalObject, { error: createErrors }] = useMutation(createDigitalObjectMutation);
```
elohanlon, line 62: @JackBlackLight Looking at this again, I realized that we'll also want to handle userErrors that come from the createDigitalObject mutation.  Since we don't create AND update in the same operation, we can safely merge both create and update user errors into the same array (since one set will always be blank).

elohanlon (approved):
@JackBlackLight Looks good!  FYI My VS Code js linter caught an issue where `const variables` was declared in an outer and inner scope (not related to your new code), so I pushed a commit to fix.  Will approve and merge now!

PR #316, opened 2021-07-06
Verification tests for non-ASCII UTF-8 (HYACINTH-522)

barmintor:
- spec graphql query with non-ascii utf8 data
- spec graphql update with non-ascii utf8
- spec batch import from csv with non-ascii utf8 data
- spec batch export of items with non-ascii utf8 data


elohanlon (approved):
Looks great!

PR #323, opened 2021-07-15
Hyacinth 775 global publish target UI

elohanlon:


app/javascript/hyacinth_v1/components/shared/forms/buttons/ProgressButton.js
```diff
@@ -0,0 +1,93 @@
...
+  onSuccess: PropTypes.func,
+  onError: PropTypes.func,
+};
+
+ProgressButton.defaultProps = {
+  onSuccess: () => {},
+  onError: () => {},
+};
+
+export default ProgressButton;
```
JackBlackLight, line 93: great to be seeing 'success' now, instead of that misleading yellow 'could not save', on the dynamic field form as well!

app/javascript/hyacinth_v1/components/publish_targets/PublishTargets.js
```diff
@@ -0,0 +1,48 @@
...
+        requiredAbility={{ action: 'create', subjectType: 'PublishTarget' }}
+        component={PublishTargetNew}
+      />
+
+      <ProtectedRoute
+        exact
+        path="/publish_targets/:stringKey"
+        component={PublishTargetShow}
+        requiredAbility={params => ({ action: 'read', subject: 'PublishTarget', publishTarget: { stringKey: params.stringKey } })}
+      />
```
JackBlackLight, line 33: curious as to why the publish target stringKey is needed here. can read ability vary based on which public target is being requested?

JackBlackLight (approved):
This looks great, @elohanlon. I had one request for more info, but am approving now.

app/javascript/hyacinth_v1/components/shared/forms/buttons/ProgressButton.js
elohanlon: Yes!  I'm so glad that's gone now!

app/javascript/hyacinth_v1/components/publish_targets/PublishTargets.js
elohanlon: Good question!  With the capabilities of our ability checking system, we technically can assign read permission on a publish target by publish target basis if we wanted to.  Do we actually do that?  No, not at this time.  But I like to be as specific as possible in case we ever change our minds in the future.

A resource like a DigitalObject actually does have id-specific checks, but for a PublishTarget we're more likely to say "user X has read permission on any PublishTarget".  Fortunately, that broader ability automatically covers the more specific stringKey-specific case in checks like the one above.

To be even more accurate though, we're currently only allowing admins to work with publish targets, and for that reason we don't even assign PublishTarget permissions explicitly.  The high level admin permission cascades down like this:

- Admin permission automatically grants broad PublishTarget access for any operation type
- Broad PublishTarget access grants broad PublishTarget read access
- Broad PublishTarget read access automatically grants specific PublishTarget read permission for a given stringKey

elohanlon:
@JackBlackLight Thank you for the review and approval!

PR #324, opened 2021-07-16
HYACINTH-630 - Implement permission-independent dynamic field validations

elohanlon:


app/validators/digital_object/dynamic_fields_validator.rb
```diff
@@ -51,6 +42,25 @@ def field_errors(configuration, value)
...
+        value.each_with_index do |v, i|
+          errors.concat errors_for(field_map_subtree[:children], v, "#{new_path}[#{i}]")
+        end
+      else
+        errors.append([new_path, "must contain an array"])
+      end
+      errors
+    end
+
+    def violates_repeat_value_constraint(field_or_group_key, value)
```
elohanlon, line 59:
```suggestion
    def violates_repeat_value_constraint(group_key, value)
```
Recommending change because this method only expects a field group and wouldn't be receiving a field.

app/validators/digital_object/dynamic_fields_validator.rb
```diff
@@ -51,6 +42,25 @@ def field_errors(configuration, value)
       send("errors_for_#{configuration[:field_type]}_field", configuration, value)
     end
 
+    def errors_for_field_group(value, field_or_group_key, new_path, field_map_subtree)
```
elohanlon, line 45:
```suggestion
    def errors_for_field_group(value, group_key, new_path, field_map_subtree)
```

app/validators/digital_object/dynamic_fields_validator.rb
```diff
@@ -51,6 +42,25 @@ def field_errors(configuration, value)
...
+      errors = []
+      if value.is_a?(Array)
+        errors.append([new_path, "is not repeatable"]) if violates_repeat_value_constraint(field_or_group_key, value)
```
elohanlon, line 48:
```suggestion
        errors.append([new_path, "is not repeatable"]) if violates_repeat_value_constraint(group_key, value)
```

app/validators/digital_object/dynamic_fields_validator.rb
```diff
@@ -51,6 +42,25 @@ def field_errors(configuration, value)
...
+          errors.concat errors_for(field_map_subtree[:children], v, "#{new_path}[#{i}]")
+        end
+      else
+        errors.append([new_path, "must contain an array"])
+      end
+      errors
+    end
+
+    def violates_repeat_value_constraint(field_or_group_key, value)
+      dynamic_field_group = DynamicFieldGroup.find_by(string_key: field_or_group_key)
```
elohanlon, line 60:
```suggestion
      dynamic_field_group = DynamicFieldGroup.find_by(string_key: group_key)
```

app/validators/digital_object/enabled_dynamic_fields_validations.rb
```diff
@@ -0,0 +1,67 @@
...
+                                project: digital_object.primary_project,
+                                digital_object_type: digital_object.digital_object_type).first
+      no_match = false
+    end
+    return ['field must be enabled'] if no_match
+    false
+  end
+
+  def check_required_fields(dynamic_field_paths, digital_object, errors)
+    required_enabled_fields = EnabledDynamicField.where(project: digital_object.primary_project,
```
elohanlon, line 38:
```suggestion
    required_fields = DynamicField.where(
      id: EnabledDynamicField.where(
        project: digital_object.primary_project, digital_object_type: digital_object.digital_object_type, required: true
      ).pluck(:dynamic_field_id)
    )
```
Since you're only using the underling dynamic_field property for each of the EnabledDynamicField objects, you can boost efficiency here by retrieving all of the relevant fields in one go rather than calling the `.dynamic_field` method on each (which I believe will lazy load the field, resulting in an extra SQL query for each iteration).

app/validators/digital_object/enabled_dynamic_fields_validations.rb
```diff
@@ -0,0 +1,67 @@
...
+    data.each do |df_group, children|
+      dynamic_field_paths = collect_field_paths(df_group, children, dynamic_field_paths)
+      dynamic_field_paths.each do |dynamic_field_path|
+        if (e = not_enabled?(dynamic_field_path, digital_object))
+          errors.concat e.map { |i| [dynamic_field_path, i] }
+        end
+      end
+    end
+
+    errors = check_required_fields(dynamic_field_paths, digital_object, errors)
```
elohanlon, line 20: This line assigns a new value to the `errors` object, but internally the `check_required_fields` method also modifies the passed-in errors object.

app/validators/digital_object/enabled_dynamic_fields_validations.rb
```diff
@@ -0,0 +1,67 @@
...
+
+    errors
+  end
+
+  def not_enabled?(dynamic_field_path, digital_object)
+    no_match = true
+    dynamic_field = DynamicField.find_by(path: dynamic_field_path)
```
elohanlon, line 27: This SQL query and the one below it appear to run separately for every field present in the record, which can get expensive for digital objects with a lot of fields.  Instead, I'd recommend collecting a single list of all fields present and comparing it to a single list of all enabled fields, and seeing what doesn't match up.

app/validators/digital_object/enabled_dynamic_fields_validations.rb
```diff
@@ -0,0 +1,67 @@
...
+  def check_required_fields(dynamic_field_paths, digital_object, errors)
+    required_enabled_fields = EnabledDynamicField.where(project: digital_object.primary_project,
+                                  digital_object_type: digital_object.digital_object_type, required: true)
+
+    required_enabled_fields.each do |enabled_field|
+      errors.concat ['is required'].map { |i| [enabled_field.dynamic_field.path, i] } unless dynamic_field_paths.include?(enabled_field.dynamic_field.path)
+    end
+
+    errors
+  end
```
elohanlon, line 46:
```suggestion
  # For the given project and digital_object type, checks the given present_dynamic_field_paths list
  # to see if it contains all of the required fields.
  #
  # @param dynamic_field_paths [Array<String>] An array of dynamic field paths to check.
  # @param project [Array<String>] An array of dynamic field paths to check.
  # @return [Array<Array<String>>] An array errors of the format: [['field name', 'error message], ['field name', 'error message]]
  def check_missing_required_fields(project, digital_object_type, present_dynamic_field_paths)
    required_field_paths = DynamicField.where(
      id: EnabledDynamicField.where(
        project: project, digital_object_type: digital_object_type, required: true
      ).pluck(:dynamic_field_id)
    ).pluck(:path)

    [required_field_paths - present_dynamic_field_paths].each do |missing_required_field_path|
      [missing_required_field_path, 'is required']
    end
  end
```
In an earlier comment, I mentioned how this method modifies errors and also returns it.  Here's a proposed new version of the method that only returns a set of errors, and also gives you a more efficient record selection method (that avoids extra queries that I think would have been run each each iteration for `enabled_field.dynamic_field`).  I changed the method signature because after removing `errors`, I realized that the method didn't actually require an entire digital_object instance to work -- it only needed a reference to a project and digital_object_type string.  That should make it easier to test, and less coupled to a more complex object.

app/validators/digital_object/enabled_dynamic_fields_validations.rb
elohanlon: After adding that other comment with the suggested new implementation for `check_required_fields`, I'm realizing that there's a parallel here for checking enabled fields.  You may want to do something similar.  In fact, you could even retrieve all of the enabled DynamicField instances just once in the `enabled_field_errors` method, and then pass them into method and `check_required_fields`.  That would reduce the SQL queries even more.

Step 1: Get list of all EnabledDynamicFields for the project/digital_object_type combo.  Make sure to have `.includes(:dynamic_field)` on that ActiveRecord query to grab all of the Fields in one go with that query.
Step 2: Take that list of EnabledDynamicFields (which includes DynamicFields) and run `enabled_dynamic_fields.map {|edf| edf.dynamic_field } to get the list of DynamicFields, and check it against the fields present in data (i.e. replacement for not_enabled? check).
Step 3: Also take that list of EnabledDynamicFields (which includes DynamicFields) and run `enabled_dynamic_fields.select {|edf| edf.required }.map{|edf| edf.dynamic_field }` to get the list of required dynamic fields, and check it against the fields present in data (i.e. replacement for `check_required_fields`).

app/validators/digital_object/enabled_dynamic_fields_validations.rb
elohanlon: Before making this suggested change, definitely check out my other comments because what I proposed here may not be exactly what you want.

spec/models/concerns/digital_object_concerns/validations/dynamic_fields_validator_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+      ]
+    }
+  end
+
+  before do
+    Hyacinth::DynamicFieldsLoader.load_fields!(field_definitions, load_vocabularies: true)
+    item.assign_descriptive_metadata({ 'descriptive_metadata' => descriptive_metadata }, false)
+  end
+
+  context 'when new value is being added to a descriptive field' do
```
elohanlon, line 35: Since the only thing inside this context is a single context (no `let`, `before`, second `context`, etc.), I'd remove the outer context and modify the inner message as needed.

elohanlon:
@JackBlackLight I can't formally approve or request changes on this PR because I initiated the PR (I should have asked you to create the new one instead, and will next time), but I'm leaving a comment here to let you know that I'm done reviewing and I've left some suggestions/comments.

elohanlon:
Approved and merged!
(I say "Approved" here in words because I can't technically "Approve" it via GitHub mechanism since I created the PR.)

PR #326, opened 2021-07-19
Hyacinth 774 enable checkboxes edit view only

JackBlackLight:


app/javascript/hyacinth_v1/components/projects/enabled_dynamic_fields/EnabledDynamicFieldForm.js
```diff
@@ -223,7 +228,9 @@ export const EnabledDynamicFieldForm = (props) => {
   const [disabled] = useState(formType !== 'edit');
```
elohanlon, line 228: The use of `useState` doesn't seem necessary for this `disabled` variable, since it's just derived from the formType prop.  (I know that's not your change, but I just noticed it here and wanted to mention it.)

Maybe it would be better to just remove this line altogether and make use of formType since it's only used once in the entire functional component.

app/javascript/hyacinth_v1/components/projects/enabled_dynamic_fields/EnabledDynamicFieldForm.js
```diff
@@ -287,6 +294,10 @@ export const EnabledDynamicFieldForm = (props) => {
       .then(historyPromise);
   };
 
+  const onSuccessHandler = (result) => {
```
elohanlon, line 297: Good idea adding this!  No reason to keep the user on the page after they hit save.

app/javascript/hyacinth_v1/components/projects/enabled_dynamic_fields/EnabledDynamicFieldForm.js
```diff
@@ -324,4 +337,40 @@ EnabledDynamicFieldForm.propTypes = {
...
+  disabled: PropTypes.bool.isRequired,
+};
+
+DynamicField.propTypes = {
+  field: PropTypes.shape({
+    id: PropTypes.number,
+  }).isRequired,
+  enabledFieldDataCallback: PropTypes.func.isRequired,
+  disabled: PropTypes.bool.isRequired,
+};
```
elohanlon, line 374: Thank you for adding these propTypes!  I want to clean up our remaining eslint errors and add an eslint check to the CI task.  Things like this really help. 👍

app/javascript/hyacinth_v1/components/projects/enabled_dynamic_fields/EnabledDynamicFieldForm.js
```diff
@@ -201,7 +205,7 @@ const DynamicFieldGroup = (props) => {
 };
 
 const DynamicFieldCategory = (props) => {
```
elohanlon, line 207:
```suggestion
const DynamicFieldCategory = ({ category, enabledFieldDataCallback, disabled }) => {
```
I've been getting to the pattern of prop destructuring lately.  This is a completely optional suggestion, but I wanted to show you an example of what I mean.

elohanlon:
GitHub isn't letting me add this as an inline comment because it's too far away from edited lines, but it would be good to also update this:
```
EnabledDynamicFieldForm.propTypes = {
  formType: PropTypes.string.isRequired,
  projectStringKey: PropTypes.string.isRequired,
  digitalObjectType: PropTypes.string.isRequired,
};
```
to this:
```
EnabledDynamicFieldForm.propTypes = {
  formType: PropTypes.oneOf(['edit', 'show']).isRequired,
  projectStringKey: PropTypes.string.isRequired,
  digitalObjectType: PropTypes.string.isRequired,
};
```

...OR...maybe even better, you could replace that formType prop with a boolean prop like:
```
readonly: PropTypes.bool,
```
And then have a default prop of false:

```
EnabledDynamicFieldForm.defaultProps = {
  readonly: false,
}
```

elohanlon (approved):

PR #334, opened 2021-08-03
Hyacinth 501 version preserved data

barmintor:
https://issues.cul.columbia.edu/browse/HYACINTH-501

> We need to preserve certain resource_attribute resources on preserve.  The resource_attribute method (used in classes like app/models/digital_object/asset.rb) should be updated to accept a second configuration hash argument that can optionally include `{preservable: true, versionable: true}` key-value pairs.
> 
> When `{preservable: true}` is present, that resource attribute's file content (if present) should be persisted by the preservation adapter.  When `{versionable: true}` is ALSO present, the resource should be versioned by the preservation system during each persist operation.
> 
> The resource_attribute configuration object should not accept `{versionable: true}` if `{preservable: true}` is not present.


lib/hyacinth/adapters/preservation_adapter/fedora3/text_resource_properties.rb
```diff
@@ -0,0 +1,55 @@
...
+
+        def to(fedora_obj)
+          return unless @hyacinth_obj.is_a? ::DigitalObject::Asset
+          ::DigitalObject::Asset::TEXT_RESOURCE_NAMES.each do |resource_name|
+            next unless @hyacinth_obj.send :"has_#{resource_name}_resource?"
+            resource = @hyacinth_obj.send :"#{resource_name}_resource"
+            next unless resource.preservable && resource.content_exists?
+            ensure_datastream(fedora_obj, resource_name, datastream_props_for(resource_name, resource))
+            datastream = fedora_obj.datastreams[resource_name]
+            content_digest = md5_digest(resource)
```
elohanlon, line 25: @barmintor The resource object has a `#checksum` method on it that you can use instead.

app/models/digital_object/asset.rb
```diff
@@ -28,6 +29,10 @@ class DigitalObject::Asset < DigitalObject
 
   attr_accessor :skip_resource_request_callbacks
 
+  def can_have_children?
```
elohanlon, line 32: Since this method is defined by the parent class with a return value of false, maybe no reason to override it here?

lib/hyacinth/adapters/preservation_adapter/fedora3/text_resource_properties.rb
```diff
@@ -0,0 +1,47 @@
...
+          new(hyacinth_obj)
+        end
+
+        def initialize(hyacinth_obj)
+          @hyacinth_obj = hyacinth_obj
+        end
+
+        def to(fedora_obj)
+          return unless @hyacinth_obj.is_a? ::DigitalObject::Asset
+          ::DigitalObject::Asset::TEXT_RESOURCE_NAMES.each do |resource_name|
```
elohanlon, line 19: @barmintor This implementation, only around text resources on assets, is more specific than I was imagining.  Can you make it more general so that the adapter iterates through all preservable resources on any digital object type, instead of just the ones that are hard-coded in the DigitalObject::Asset::TEXT_RESOURCE_NAMES constant?

lib/hyacinth/adapters/preservation_adapter/fedora3/text_resource_properties.rb
```diff
@@ -0,0 +1,47 @@
...
+            next unless @hyacinth_obj.send :"has_#{resource_name}_resource?"
+            resource = @hyacinth_obj.send :"#{resource_name}_resource"
+            resource_config = @hyacinth_obj.class.resource_attributes.fetch(resource_name.to_sym, {})
+            next unless resource_config[:preservable] && resource.content_exists?
+            ensure_datastream(fedora_obj, resource_name, datastream_props_for(resource_name, resource_config, resource))
+            datastream = fedora_obj.datastreams[resource_name]
+            # Only update content if it has changed
+            next if "sha256:#{datastream.checksum}" == resource.checksum
```
elohanlon, line 27: I remember you mentioning checksums on text resources vs larger ones on non-text resources.  This is making me think that we want to add a `verify_checksum: true|false` config option to resource_attributes as well, and only have fedora verify the checksum if that's marked as true (but always store the checksum as a RELS-INT property if present, regardless of verify_checksum property value).

lib/hyacinth/digital_object/resource_attributes.rb
```diff
@@ -11,51 +11,67 @@ def resource_attributes
...
-          @resource_attributes << resource_attribute_name.to_sym
+        def resource_attribute(resource_attribute_name, config = {})
+          config ||= {}
+          @resource_attributes ||= {}
+          @resource_attributes[resource_attribute_name.to_sym] = config
 
-          @resource_import_attributes ||= Set.new
-          @resource_import_attributes << resource_attribute_name.to_sym
+          @resource_import_attributes ||= {}
+          @resource_import_attributes[resource_attribute_name.to_sym] = config
```
elohanlon, line 74: I'm not sure that we benefit from also associating the resource_attribute config options with the resource_import_attributes.  This change may not be necessary?

app/models/digital_object/asset.rb
barmintor: I'm a little nervous of inheriting that behavior - I'm almost inclined to make the parent class method `raise` instead.

lib/hyacinth/adapters/preservation_adapter/fedora3/text_resource_properties.rb
barmintor: This is definitely the HYACINTH-501 implementation - I think the HYACINTH-814 commits will make it more general.

lib/hyacinth/adapters/preservation_adapter/fedora3/text_resource_properties.rb
barmintor: it also only makes sense for FCR-managed content, really

lib/hyacinth/digital_object/resource_attributes.rb
barmintor: If I remember correctly there's a place where the import `merge` the resource attribute config (or I thought there was). Maybe not necessary after all?

spec/hyacinth/adapters/preservation_adapter/fedora3/system_spec.rb
```diff
@@ -90,7 +90,7 @@
       end
     end
     context "basic rels-ext properties" do
-      let(:dsids) { ['structMetadata'] }
+      # let(:dsids) { ['structMetadata'] }
```
elohanlon, line 93: Should this comment line be deleted?

app/models/digital_object.rb
```diff
@@ -91,25 +91,16 @@ def number_of_children
     children.length
   end
 
-  def can_have_rights?
+  def can_have_children?
```
elohanlon, line 94: Even though it's a really simple method, we should probably add a test for this and the other DigitalObject subclasses to ensure that they all behave as expected.

spec/hyacinth/adapters/preservation_adapter/fedora3/system_spec.rb
barmintor: yes!

app/models/digital_object/asset.rb
elohanlon: I'm good with going with raise instead, but my other thought about this is: Should this actually be an instance method, or is static a better choice?  This is one of those cases where I wish ruby was more like java, and you could call static methods on instances.

lib/hyacinth/adapters/preservation_adapter/fedora3/text_resource_properties.rb
elohanlon: That works for me. 👍

lib/hyacinth/adapters/preservation_adapter/fedora3/text_resource_properties.rb
elohanlon: True.  I'll resolve this conversation since we talked about nesting the arguments under the `preserve` key.

lib/hyacinth/digital_object/resource_attributes.rb
elohanlon: I don't remember any merging of configs for the resource_import_attributes (but I could be forgetting what you're referring to).  Do you mean merging related to the superclass?  That was a case of merging Sets of strings.

lib/hyacinth/adapters/preservation_adapter/fedora3/preserved_resource_properties.rb
```diff
@@ -0,0 +1,53 @@
...
+            next unless @hyacinth_obj.send :"has_#{resource_name}_resource?"
+            resource = @hyacinth_obj.send :"#{resource_name}_resource"
+            resource_config = @hyacinth_obj.class.resource_attributes.fetch(resource_name.to_sym, {})
+            preservable_config = resource_config.fetch(:preservable, {})
+            next unless preservable_config.present? && resource.content_exists?
+            dsid = adapter.dsids_for_resources.fetch(resource_name, resource_name)
+            ensure_datastream(fedora_obj, dsid, datastream_props_for(resource_name, preservable_config, resource))
+            datastream = fedora_obj.datastreams[dsid]
+            # Only update content if it has changed
+            next if "sha256:#{datastream.checksum}" == resource.checksum
```
elohanlon, line 23: I'm not sure if this is a problem or not, but technically if you change a resource_attribute persist config from :copy to :reference (or vice versa), I think this line will block that update from happening if the resource's checksum hasn't changed.

lib/hyacinth/digital_object/resource_attributes.rb
```diff
@@ -11,51 +11,67 @@ def resource_attributes
...
           end
         end
 
+        def resource_attribute_names
+          resource_attributes.keys.dup.freeze
+        end
+
         def resource_import_attributes
-          @resource_import_attributes ||= Set.new # initialized here because it may not have been initialized in a subclass of the including class
+          @resource_import_attributes ||= {} # initialized here because it may not have been initialized in a subclass of the including class
```
elohanlon, line 60: For later: If we don't end up needing config for resource_import_attributes, this should go back to `Set.new`.

lib/hyacinth/adapters/preservation_adapter/fedora3/preserved_resource_properties.rb
barmintor: if you change a resource from `:copy` to `:reference` after it has been preserved, you will need to do a migration like the one we did with AC data recently!

elohanlon: Hopefully it'll be easier though, since the big difference between that migration and this situation is that Hyacinth will always be the owner of the original copy.  Fedora 3, as a preservation target, is just being told to track Hyacinth's copy for 'E' or make yet another copy for 'M', so I was thinking of it as different.

There may be complications related to Fedora 3 datastream versioning, but I imagined that a switch from :copy to :reference would involve telling Fedora to delete the managed copy and then track Hyacinth's file on the filesystem.  And the opposite would be even easier, since it would be telling Fedora to switch from tracking a file to managing a copy.

The concern in my original comment was just that a Hyacinth developer might change a resource_attribute configuration parameter and not realize that something had gone wrong (or that a manual migration was required) as long as the file checksum hadn't changed.  But we can document that.

app/models/digital_object/asset.rb
barmintor: I'm going to break out a separate ticket for refactoring and testing `can_have_children?` and `can_have_rights?` (which the implementation was copied after): HYACINTH-826

app/models/digital_object.rb
barmintor: Moving this to another ticket (HYACINTH-826)

lib/hyacinth/adapters/preservation_adapter/fedora3/text_resource_properties.rb
barmintor: Marking resolved (now using the sha256)

lib/hyacinth/adapters/preservation_adapter/fedora3/preserved_resource_properties.rb
barmintor: Another option would be to use the configuration of the datastream (which will match class property if new) to determine content assignment behavior, and then ticket a change to delete and replace if class property has changed.

elohanlon (approved):

PR #336, opened 2021-08-04
enable tested fields in spec for DigitalObject::DescriptiveFieldsValidator (HYACINTH-630)

barmintor:


barmintor:
Also changing factory call to `build` instead of `create` to prevent saving/indexing

elohanlon (approved):

PR #342, opened 2021-08-10
confirm call of .deindex after DigitalObject#destroy and .index only …

barmintor:
…on save commits (HYACINTH-796)

### DIFF
--- a/spec/hyacinth/adapters/digital_object_search_adapter/solr/system_spec.rb
+++ b/spec/hyacinth/adapters/digital_object_search_adapter/solr/system_spec.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require 'shared_examples/digital_object_search_adapter/shared_examples'
+
+# TODO: HYACINTH-840 Refactor this to permit configuration of in-memory impl for test
+# - create an independent instance of Hyacinth::Adapters::DigitalObjectSearchAdapter::Solr here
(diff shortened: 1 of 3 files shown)


spec/hyacinth/adapters/digital_object_search_adapter/solr/system_spec.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require 'shared_examples/digital_object_search_adapter/shared_examples'
+
+describe Hyacinth::Adapters::DigitalObjectSearchAdapter::Solr, solr: true do
+  let(:adapter) { Hyacinth::Config.digital_object_search_adapter }
```
elohanlon, line 7: If this test is testing the `Hyacinth::Adapters::DigitalObjectSearchAdapter::Solr` implementation, you'll want to instantiate a new solr adapter rather than relying on the adapter selected by `Hyacinth::Config.digital_object_search_adapter`.

barmintor: Added a comment here to refer to subsequent ticket - I agree, but there are some complications to deal with that seem out of scope of this ticket.

elohanlon: 👍

elohanlon (approved):

PR #343, opened 2021-08-11
HYACINTH-841 - Hyacinth setup/development tasks should automatically enable title fields for assets

elohanlon:


elohanlon:
Agreed!  Definitely a short-term fix.  Thank you for the review!

PR #347, opened 2021-08-12
Hyacinth 847 - Fix UserNew.js and add feature test for rendered page

barmintor:
Update dependencies to run Capybara with selenium headless
Correct an import bug in UserNew.js
Add spec support for feature specs and verify that /ui/v1/users/new renders


.github/workflows/ci.yml
```diff
@@ -44,6 +44,6 @@ jobs:
         with:
           node-version: ${{ matrix.node }}
       - run: npm install -g yarn
-      - run: yarn install
+      - run: yarn install --frozen-lockfile
```
elohanlon, line 47: Good thinking 👍

lib/tasks/hyacinth/ci.rake
```diff
@@ -29,7 +29,7 @@ namespace :hyacinth do
     ENV['INCLUDE_HYACINTH_SPECS'] = includes.join(',')
     rspec_system_exit_failure_exception = nil
 
-    task_stack = ['hyacinth:rspec']
+    task_stack = ['webpacker:clobber', 'webpacker:compile', 'hyacinth:rspec']
```
elohanlon, line 32: Does this make the ci run take longer when a developer runs it locally?

If it doesn't have an effect on the actual ci functionality (and if you just added it to verify that the webpack compile tasks runs), should it be added only as an ENV config option that we enable in a CI environment like GitHub Actions?

spec/rails_helper.rb
```diff
@@ -70,6 +76,7 @@
 
   # Include helpers
   config.include JsonSpec::Helpers
+  config.include AuthenticatedRequests, type: :feature
   config.include AuthenticatedRequests, type: :request
```
elohanlon, line 80:
```suggestion
  [:request, :feature].each do |type|
    config.include AuthenticatedRequests, type: type
  end
```

lib/tasks/hyacinth/ci.rake
elohanlon: Also, maybe I'm misunderstanding your intention here, but here's some info from something I just tested:

As you'll recall, our earlier commit 6e2f831d2c35c16aa62cee4fe5dd7c3fc756db85 had this error when we tried to deploy:
```
ERROR in ./app/javascript/packs/hyacinth_v1.js
Module build failed (from ./node_modules/babel-loader/lib/index.js):
Error: Cannot find module 'babel-plugin-transform-react-remove-prop-types'
...
...
```

I checked out that commit locally and ran:
```
bundle exec rake webpacker:clobber webpacker:compile
```
and that completed without error.

But when I ran this...
```
rm -rf node_modules; yarn install; rm -rf public/packs public/assets; mkdir public/packs; mkdir public/assets; NODE_ENV=production bundle exec rake assets:precompile
```
...it failed and printed that `Error: Cannot find module 'babel-plugin-transform-react-remove-prop-types'` error.

spec/features/ui/v1/users_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  describe 'GET /users/new' do
+    context 'when logged in user has appropriate permissions' do
+      context 'when string_key is valid' do
+        before do
+          sign_in_user_manager
+          visit "/ui/v1/users/new"
+        end
+
+        it 'returns correct response' do
+          expect(page).to have_content('Create New User')
```
elohanlon, line 15: Very glad to have this test (and soon, others like it)!  👍

lib/tasks/hyacinth/ci.rake
barmintor: I'm not sure what you're saying? In my case, the webpack compile was clearly executed at some point - but the script wasn't running (and the error is not available to a headless browser that I could see; blank screenshot outputs). When the tests ran with chrome instead of headless, then I could see the console (until the test timed out). That's where I was seeing this error.

barmintor: > Does this make the ci run take longer when a developer runs it locally?
> 
> If it doesn't have an effect on the actual ci functionality (and if you just added it to verify that the webpack compile tasks runs), should it be added only as an ENV config option that we enable in a CI environment like GitHub Actions?

I added some logic to make it optional (presuming it runs), but we need to acculturate the team to doing a test environment clobber/recompile if feature specs appear not to be executing the entry point js.

spec/rails_helper.rb
barmintor: I rearranged the `config.include` calls to group the three related to authentication and do something like this.

lib/tasks/hyacinth/ci.rake
elohanlon: I was saying that if your intention behind the addition of `webpacker:clobber` and `webpacker:compile` was to detect webpack build errors like the deployment one that we saw before, it wasn't working.  But it sounds like you're saying that it is not, and that it's only intended to detect a type of post-build runtime error that can appear in the chrome browser console -- is that right?

Btw, it may be good to clarify this intention with a comment above the `task_stack = ['webpacker:clobber', 'webpacker:compile', 'hyacinth:rspec']` line, since this seems like the thing that someone could forget down the line.

elohanlon: > I added some logic to make it optional (presuming it runs), but we need to acculturate the team to doing a test environment clobber/recompile if feature specs appear not to be executing the entry point js.

Thank you for that.  I just tried running the task locally and it adds an extra 25 seconds to every run, so it's nice making that optional.

spec/rails_helper.rb
elohanlon: 👍

PR #349, opened 2021-08-12
Remove webdrivers

barmintor:
If you want to try just removing webdrivers for now


elohanlon:
I found a solution!  No need to remove webdrivers after all.  See: https://github.com/cul/ldpd-hyacinth/pull/348/commits/02a1237cc25a751db7344a730426d5e15f201dc3 (second commit in https://github.com/cul/ldpd-hyacinth/pull/348).

It's good to keep it for both CI server and local testing, to ensure the latest driver version.

PR #350, opened 2021-08-13
Hyacinth 473 drag rearrange child objects

JackBlackLight:


JackBlackLight:
@elohanlon, there's actually one more piece to add on the frontend, but I wanted your input on the way I've arranged things. I found it easiest to place all the beautiful-dnd components in DigitalObjectList. This meant that I needed to pass a flag signaling whether drag and drop functionality should be enabled (for Children.js) or disabled (for DigitalObjectSearch). The last step is to share the state of the reordered list with Children.js for when the new list order is persisted - I was thinking of using window.setStorage to share these values between components, unless you think there's a better way. Given these questions, I've opened this as a draft PR.

elohanlon:
@JackBlackLight I tried this locally and realized that your code is behind master (by quite a lot of commits actually).  Could you please rebase when you have a chance?  There are some more changes coming soon and I don't want you to get more out of sync.

`window.sessionStorage` is good for sharing data across react-router pages within the same single-page-app load, but we generally don't want to use it for sharing data across components.  In general, if you find yourself wanting to write data to an object where writes don't automatically trigger a React re-render, you may be working against React rather than with it.  You probably want to update the DigitalObjectList component to take an `onRearrange` callback function (whose implementation is passed in by Children.js), write a new child-order-saving mutation, use a `useMutation` hook in Children.js and fire off the request as part of the `onRearrange` callback, and then when the request is done you'll want to call the Children.js refetch method (aliased as `refreshChildStructure`) that's returned by `useQuery(getChildStructureQuery)`.

Before you go all in on the above method though, I'm starting to think that we actually may not want to keep modifying the DigitalObjectList component for rendering this sortable list.  Between your addition of the disableDrag prop and my suggestion above to also add an `onRearrange` prop, it's starting to seem like DigitalObjectList isn't just a plain list of digital objects anymore.  Also, since it's primarily used for search results and is likely going to include extra info and actions down the line that won't be relevant on a child reordering page, that makes me think even more that it's not the right component for the job.  And (looking at the component more closely now) the DigitalObjectList component also has limit and offset props that aren't relevant for child display -- since you can't rearrange a big list of children if the results are paged.

We currently have some items in Hyacinth 2 with hundreds of children, so that's another reason to create a new component  -- one that's better suited to showing information in a more compact form.  Maybe good to call it something like `<DigitalObjectChildList>`?  It can show extremely minimal information for each child like we do in Hyacinth 2.  For now, I'd recommend just displaying title and uid.  Some time later we'll also add a thumbnail image, but that's definitely out of scope for this ticket.

So if you go with the new component, you can take my suggestion above and just pretend that I was talking about modifying the `<DigitalObjectChildList>` component rather than `<ListChildPages>`.

app/javascript/src/hyacinth_v1/components/digital_objects/DigitalObjectList.js
```diff
@@ -43,40 +44,87 @@ const storeSearchValues = (
 const DigitalObjectList = (props) => {
   const {
     digitalObjects, displayProjects, displayParentIds, orderBy, totalCount,
-    limit, offset, pageNumber, searchParams, path,
+    limit, offset, pageNumber, searchParams, path, disableDrag,
```
elohanlon, line 47: Can the drag logic be removed from this component now that you've introduced the other component?

JackBlackLight: Apologies for the confusion of leaving that in. I restored DigitalObjectList.js to the version in main, which appears to include some new formatting adjustments from that branch. DigitalObjectSearch.js has similar formatting adjustments, but I didn't add those - possibly a side effect of the git history cleanup we worked on?

Quick note: there's an update button on the child list page, but the onSubmit is not implemented. I'll add that as part of the HYACINTH-854 work, unless you'd prefer it be ticketed separately. Thanks!

elohanlon: @JackBlackLight Are you sure you restored those files from the latest version of main?  I'm not seeing those same changes in https://github.com/cul/ldpd-hyacinth/blob/main/app/javascript/src/hyacinth_v1/components/digital_objects/DigitalObjectList.js

See line 83 in that file, for example.  Your changes will end up reverting lines like that to an earlier state if we merge this in, since your commit will have a newer commit date.  The reason you're seeing the changes as a delta is because they differ from the recently rebased main branch.

Also, I tried your branch locally (to try adding a child object to an Item that didn't have any children yet) and the children page isn't rendering for me.  I'm getting a blank page and a `useState`-related error in the console.

One other note, just to clarify: Remember in my earlier comment when I said this?

> Since not all users will have permission to reorder child objects, this feature should also be implemented on a separate view, meaning that in DigitalObject.js under { routePath: ${path}/children, Component: Children, action: 'read_objects' }, you'll want to add a new route to something like { routePath: ${path}/edit_child_structure, Component: EditChildStructure, action: 'edit_objects' }, (and then create a new EditChildStructure page component).

What I meant was that you didn't need to change the current child view substantially (i.e. you can continue using the DigitalObjectList on the `/children` page) , but that it would be helpful to have the child editor on a separate page located at routePath `${path}/edit_child_structure`.  In this case, you could add a link above the child view that says "Reorder children" and that would bring users to the new page.  One big reason for a new page is that people who have permission to read an object won't always have permission to reorder children.  Adding a new page isn't usually too bad (and you'll see examples of this in `DigitalObject.js` already, along with permission restrictions via `DigitalObjectProtectedRoute`), but let me know if you run into any issues.

Although you could technically allow child reordering on the same page instead of creating a new one, that would involve adding some kind of  `editMode=true` state variable in Children.js and conditionally rendering DigitalObjectList or DigitalObjectChildList based on whether that had been switched on or off by a button -- but I wouldn't recommend it and it would also involve more permission-checking logic and bundling a little too much functionality into Children.js.  Since you already have a working DigitalObjectChildList component, I think you'd be better off with the separate page at this point.

Re: onSubmit button -- Totally fine to keep that non-functional and bundle it into HYACINTH-854 work.  That way we can get this code merged in before HYACINTH-854 is done.

PR #352, opened 2021-08-13
Hyacinth webpack analysis

barmintor:


config/webpack/development.js
```diff
@@ -1,5 +1,5 @@
-process.env.NODE_ENV = process.env.NODE_ENV || 'development'
+process.env.NODE_ENV = process.env.NODE_ENV || 'development';
 
-const environment = require('./environment')
+const environment = require('./environment');
```
elohanlon, line 3: Is there any value is also calling `environment.splitChunks();` in development.js?  Would it make the dynamic recompilation faster by recompiling a smaller chunk?

app/javascript/src/hyacinth_v1/utils/lazyFontAwesome.js
```diff
@@ -0,0 +1,19 @@
...
+LazyFontAwesomeIcon.propTypes = {
+  fallback: PropTypes.object,
+  icon: PropTypes.string.isRequired,
+};
+
+LazyFontAwesomeIcon.defaultProps = {
+  fallback: <span />,
+};
+
+export default LazyFontAwesomeIcon;
```
elohanlon, line 19: Very creative solution with this module (and lazyJSONInput.js). I like this pattern a lot better than having to use the magic comments in many different places.

app/views/layouts/hyacinth_v1.html.erb
```diff
@@ -8,10 +8,10 @@
     <%= csrf_meta_tags %>
     <%= csp_meta_tag %>
     <% # stylesheet_pack_tag is available as long as javascript_pack_tag includes at least one css file %>
-    <%= stylesheet_pack_tag 'hyacinth_v1' %>
+    <%= stylesheet_packs_with_chunks_tag 'hyacinth_v1' %>
```
elohanlon, line 11: Aha.  CSS in chunks too.  Nice!

config/webpacker.yml
```diff
@@ -77,7 +77,11 @@ development:
```
elohanlon, line 77: Reminder: You'll want to delete my comment above `extract_css: false` in the default section if we're keeping that value `false`.

elohanlon (approved):
These are some great changes!  We should deploy this to dev before we merge (if you haven't already) just to make sure that everything works as expected in the deployed environment, but I'm glad to approve the PR now.  👍

config/webpack/development.js
barmintor: I don't _think_ it has an effect unless the corresponding webpack config is set to compile.

app/javascript/src/hyacinth_v1/utils/lazyFontAwesome.js
barmintor: I'm not gonna lie, I was pretty pleased with this refactor.

barmintor:
Build failures - something about the ace editor on linux: https://gist.github.com/barmintor/294965a1088a3e891dca6a3a237a25bb

elohanlon:
@barmintor I get the same error when pulling down this branch and running `yarn install`.

elohanlon:
@barmintor Can you try out this page in your local copy? http://localhost:3000/ui/v1/field_export_profiles/new

When I'm typing in the latest version of the JSONInput component, I'm encountering a problem where every time I type a character the typing cursor jumps to the end of the entire input area.

elohanlon:
On the same page I mentioned above, I'm also seeing two js console errors:
```
Refused to execute script from 'http://localhost:3000/ui/v1/field_export_profiles/mode-json.js' because its MIME type ('text/html') is not executable, and strict MIME type checking is enabled.

Refused to execute script from 'http://localhost:3000/ui/v1/field_export_profiles/theme-textmate.js' because its MIME type ('text/html') is not executable, and strict MIME type checking is enabled.
```

But these warnings go away if you modify JSONInput.js and change this:
```
mode="ace/mode/json"
theme="ace/theme/textmate"
```
to this:
```
mode="json"
theme="textmate"
```

I thought these imports might have been necessary to add too:
```
import 'ace-builds/src-noconflict/mode-json';
import 'ace-builds/src-noconflict/theme-textmate';
```
But I don't see any errors if they're absent.

barmintor:
Pretty sure if you pop the last commit off branch it will run locally, that
was an attempt to fix server build per react-ace docs

elohanlon:
Ah, I see what you mean.  JS console errors are gone.  I hadn't looked at the latest commit delta.

The typing issue at `http://localhost:3000/ui/v1/field_export_profiles/new` is still there though.

barmintor:
Hmm, will look into it - I developed against the dynamic field select
options, so maybe I missed something about field export.

barmintor:
Some of this is a red herring - I saw this in the logs:
```bash
Trace: 
  Error: ENOTDIR: not a directory, lstat '/opt/passenger/ldpd/hyacinth_3_dev/releases/20210815165049/node_modules/react-ace/dist/react-ace.min.js/main.js'
```
... and sure enough, react-ace was minified on dev. So I'm trying to deploy the original branch with a completely fresh yarn install.

barmintor:
this worked for fixing the build - now I'm debugging what is some kind of configuration issue with the ACE editor (and that causes the cursor issue you mentioned - there's a null pointer in the EditSession for a prop called `bgTokenizer`).

barmintor:
Closing in favor of simpler PR for webpack chunks

PR #353, opened 2021-08-13
Hyacinth 852 form buttons history cleanup

barmintor:
use SuccessHandler functions to prevent state updates on unloaded components

### DIFF
--- a/app/javascript/src/hyacinth_v1/components/controlled_vocabularies/ControlledVocabularyForm.js
+++ b/app/javascript/src/hyacinth_v1/components/controlled_vocabularies/ControlledVocabularyForm.js
@@ -26,14 +26,20 @@ function ControlledVocabularyForm(props) {
   const [updateVocabulary, { error: updateError }] = useMutation(updateVocabularyMutation);
   const [deleteVocabulary, { error: deleteError }] = useMutation(deleteVocabularyMutation);
 
-  const onSubmitHandler = () => {
+  const onSuccessHandler = (result) => {
+    if (result.data.createVocabulary) {
+      history.push(`/controlled_vocabularies/${result.data.createVocabulary.vocabulary.stringKey}/edit`);
+    } else if (result.data.deleteVocabulary) {
+      history.push('/controlled_vocabularies');
+    }
+  };
+
+  const onSaveHandler = () => {
     const variables = { input: { stringKey, label, locked } };
 
     switch (formType) {
       case 'new':
-        return createVocabulary({ variables }).then((res) => {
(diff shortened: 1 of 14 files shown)


elohanlon:
Good fixes for the state change attempts after unmount.  A few general thoughts:

1) There's something I don't like about FormButtons' onSuccess and onDelete operations both triggering the onSuccess handler.  It makes it so the onSuccess handler needs to inspect the data from the result and then you end up mixing submit-success and delete-success logic in the same callback function.  Those seem like two things that shouldn't be tied together like that.  How would you feel about changing the double-duty `onSuccess` to `onSubmitSuccess` and `onDeleteSuccess` props?

2) Unrelated to your changes, I'm not a huge fan of how the DeleteButton conditionally shows itself based on the formType prop that's passed in.  I'd rather have the outer caller conditionally render it or not.  Maybe we can change that in the near future (in another PR).

3) Related to comment2, maybe it's worth replacing DeleteButton in the FormButtons component with a ProgressButton (but this can be done in another PR).  And maybe we could even replace other occurrences with the DeleteButton with ProgressButton.  It would be nice to have the same kind of status animation for deletions as we have for submissions.

barmintor:
My responses are out of order:

(2) Me neither, but it would be a bigger change - mostly wanted to get rid of the memory leaks (and class components) here.
(3) I like how this idea - it would clarify the successHandler issue more than breaking out a distinct save and delete, because...
(1) the current save success handler still has to inspect the data, to figure out if it was a create (push to history) or update (do nothing or back to show depending on form)

elohanlon:
I'm fine with inspecting the data to distinguish between create or update since those seem related (like handling the result of an upsert scenario).  Deletion just felt different.  In some ways, the design of FormButtons overall is actually kind of weird.  It provides save and delete buttons in an edit view context, meaning that you need to edit a record to delete it, despite edit and delete being different permissions.  It also means you need to scroll all the way down the edit form to delete something.

My concern was an attempt to look our for future-me, or future-any-of-us, since I found the grouping of save handling and delete handling to be kind of unintuitive.  I'm just imagining myself, several months from now, using the FormButtons component and thinking, "So the onSuccess handler is used for handling save success...but what callback do I use for successful deletion?  Oh, I guess it goes in there with my save handling code."

barmintor:
I know what you mean, but it's funny - I went into this branch assuming that delete data was passed to that handler, and it wasn't! If it seems counterintuitive, maybe splitting (or renaming) would help.

elohanlon:
I can definitely see why it would seem that way!  Tbh, it was just an incomplete implementation with no consideration for the result of the delete action.  I think splitting/renaming would help.

elohanlon (approved):

PR #355, opened 2021-08-17
HYACINTH-850 - Rails 6.1

elohanlon:


app/models/batch_import.rb
```diff
@@ -152,10 +152,10 @@ def ensure_imports_are_complete!
       imports_in_progress = import_count('in_progress').positive?
 
       if cancelled && imports_in_progress
-        errors[:base] << 'Cannot destroy cancelled batch import while imports are in_progress'
+        errors.add(:base, 'Cannot destroy cancelled batch import while imports are in_progress')
```
barmintor, line 155: https://edgeapi.rubyonrails.org/classes/ActiveModel/Errors.html suggests that this should be `errors.add(:base, message: 'Cannot destroy cancelled batch import while imports are in_progress')` ... does it work as expected with the two parameters?

config/initializers/filter_parameter_logging.rb
```diff
@@ -3,4 +3,6 @@
 # Be sure to restart your server when you modify this file.
 
 # Configure sensitive parameters which will be filtered from the log file.
-Rails.application.config.filter_parameters += [:password]
+Rails.application.config.filter_parameters += [
+  :passw, :secret, :token, :_key, :crypt, :salt, :certificate, :otp, :ssn
```
barmintor, line 7: is `passw` a type here? I can't find this change in the 6.1 notes

barmintor (commented):
Couple of things look a little suspicious to me, but generally not too surprising. Will need to make some changes to the language tag PR based on these when merged.

app/models/batch_import.rb
elohanlon: It does.  In the `add` method description here (https://edgeapi.rubyonrails.org/classes/ActiveModel/Errors.html#method-i-add) it says `If type is a string, it will be used as error message.`.

config/initializers/filter_parameter_logging.rb
elohanlon: It's correct.  This file was created by the rails `app:update` rake task.  You can see an example of `passw` here:

https://github.com/rails/rails/blob/main/railties/lib/rails/generators/rails/app/templates/config/initializers/filter_parameter_logging.rb.tt

elohanlon:
@barmintor Thank you for your review!  Just replied to your comments.  If things look good at this point, okay for me to merge?

barmintor (approved):

PR #356, opened 2021-08-20
HYACINTH-835 language tag subsystem and default language configuration

barmintor:


barmintor:
squashed in a removal of a typedef stub that should be in a branch for HYACINTH-833, not HYACINTH-835

lib/hyacinth/config.rb
```diff
@@ -58,5 +58,22 @@ def self.external_identifier_adapter
...
+    end
+
+    def self.allowed_lang_value?(tag_value)
+      ::Language::Tag.for(tag_value)
+    rescue
+      false
+    end
+
+    def self.load_default_subtags!
+      Rails.application.config_for(:lang).fetch(:default_lang_subtags, {}).each do |subtag, attributes|
```
elohanlon, line 73:
```suggestion
      LANG.fetch(:default_lang_subtags, {}).each do |subtag, attributes|
```
Okay to use LANG?

barmintor: It creates a dependency on initializer execution that I think might not be desirable here - I'd want to revisit and make sure tests run, etc. Can we ticket as a cleanup task?

elohanlon: Later cleanup task is fine, but aren't the initializers always going to be loaded before you use the Hyacinth::Config object?  You're already using LANG a couple of methods up in the `default_lang_value`.

app/models/language.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module Language
+  def self.table_name_prefix
+    'language_'
+  end
```
elohanlon, line 6: I didn't realize you could do this for inheriting the table prefix.  This isn't a change suggestion, but I saw this and wanted to give you a heads up in case you didn't know about a minor limitation:

https://stackoverflow.com/questions/27483859/nested-modules-in-rails-breaking-table-name-prefix

app/models/language/tag.rb
```diff
@@ -0,0 +1,179 @@
+# frozen_string_literal: true
+
+class Language::Tag < ApplicationRecord
+  has_many :tag_subtags
```
elohanlon, line 4: I did a search through the PR changes and I only see a reference to `tag_subtag` here, in a migration, and in the schema.  Is it used elsewhere?

barmintor: ah, I think this is a relic of `has_many through:` in an earlier iteration

barmintor: the migration supports the HABTM

lib/hyacinth/config.rb
barmintor: This method gets called in contexts that `default_lang_value` does not (eg specs)

app/models/language/tag.rb
```diff
@@ -0,0 +1,179 @@
...
+    first_subtag_of_type('extlang')
+  end
+
+  def tag_value(use_preferred = false)
+    TagValue.for_subtags(subtags, use_preferred)
+  rescue SubtagError => ex
+    self.errors.add :subtags, ex.message
+  end
+
+  def set_tag
```
elohanlon, line 41: I was a little thrown off by method names like `set_tag` and `set_preferred_value` that look like setter methods, but don't take any arguments.  Changing isn't absolutely necessary, but just giving my thoughts as I go through the PR.

app/models/language/tag.rb
```diff
@@ -0,0 +1,179 @@
...
+        if other_language
+          subtags.delete(duplicate)
+        else
+          subtags.delete(extlang)
+        end
+      end
+    end
+    subtags
+  end
+  class SubtagError < RuntimeError; end
```
elohanlon, line 99: I often subclass StandardError for custom error classes, but maybe I should be using RuntimeError instead like you've done here.

app/models/language/tag.rb
barmintor: I will admit I had flashbacks to talking about `set_X` and `assign_X` on the mezz. Definitely not wedded to the names.

lib/hyacinth/config.rb
elohanlon: Ah, I see what you mean.  Okay, I take back my original suggestion.  Here's what I'd recommend instead:  I don't think this method belongs in Hyacinth::Config.  The Hyacinth::Config class isn't supposed to actually _do_ things.  It's just supposed to _provide_ things.  Can we move this into a different module, along with the `allowed_lang_value?` method?  Maybe `Hyacinth::Language`?

app/models/language/tag.rb
barmintor: Rubocop's suggestion - it is just one subclass away!

config/initializers/lang.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+LANG = Rails.application.config_for(:lang)
+
+Hyacinth::Config.load_default_subtags! if ActiveRecord::Base.connection.table_exists? :language_subtags
```
elohanlon, line 5: So to verify, this is supposed to run every time the app starts in case someone changed the default subtags since the last application restart?

app/models/language/tag.rb
elohanlon: @barmintor and I talked this over.  We'll want to:

Rename methods to:
`set_tag_from_subtags!` and `set_preferred_tag_from_subtags!`

And also make them private, since there's no benefit to offering them to outside callers.

app/models/language/tag.rb
elohanlon: @barmintor and I talked this over and we think that `has_many :tag_subtags` can be deleted (assuming tests still pass).

spec/models/language/tag_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+      expect(lang_tag.tag).to eql tag_value
+      expect(lang_tag.suppressed_script&.subtag).to eql 'Latn'
+    end
+  end
+  context 'tag has a region' do
+    let(:iana_fixtures) { [iana_en_fixture] }
+    let(:tag_value) { 'en-US' }
+    it "associates a suppressed script without including it in the tag" do
+      expect(lang_tag.tag).to eql tag_value
+      expect(lang_tag.suppressed_script&.subtag).to eql 'Latn'
```
elohanlon, line 31: Why use the safe navigation operator after `suppressed_script` here if you're expecting a known test value?

elohanlon: Same question for three other tests in this file that use `&.`.

config/initializers/lang.rb
barmintor: Yes - and we don't plan on storing default tag data on the values.

lib/hyacinth/config.rb
barmintor: These two methods have been moved to `Hyacinth::Language`

app/models/language/tag.rb
barmintor: This association was extraneous and deleted.

app/models/language/tag.rb
barmintor: This change to private methods was implemented, although it did require use of a `send` in one instance (also inside this class).

elohanlon (approved):

PR #358, opened 2021-08-24
Hyacinth 463 title typedef

barmintor:


barmintor:
@elohanlon if we can get the initial language stuff merged I'll cherrypick the two commits added here into a new branch, and pick up the migration and front-end changes from there.

barmintor:
I have rebased with updates for the refactor to `title.value` and `title.value_lang`. I have also broken the spec support refactors allowing `rspec --tag ~@\solr --tag ~@\fedora` to run into a separate commit (the first in the current PR, not tagged with HYACINTH-463). I think we may want to consider merging it separately (and first), as it should review quickly and it will clean the diff from the actual HYACINTH-463 work up a lot.

For reference:

- `rake hyacinth:ci` -> 1654 examples; 96.68% coverage
- `rspec --tag ~@\solr --tag ~@\fedora` -> 1355 examples; 90.45% coverage

elohanlon:
@barmintor Ideally I'd like to implement a fast, in-memory search adapter at some later point so that we don't have to stub so much, and so that other tests that actually do rely on a real search adapter will run more quickly, but this is a good move for the tests you've modified.  Merging the commit separately sounds good too.  I've looked it over (https://github.com/cul/ldpd-hyacinth/pull/358/commits/8cb9d646a778cd62b76ac38aa2b09bdbbc031f72) and the only thing I caught was that you have one `solr: false` in there that I think should be omitted.

👍

barmintor:
@elohanlon I know exactly which one you are talking about, and I just ran out of energy on the refactor TBQH. It might not be necessary since if I remember right the "sibling" test that does require solr has a `solr: true` flag.

elohanlon:
@barmintor Given the position in the commit order, I'm fine with merging that commit as is and we can change the flag after this branch gets merged.  I'm all for keeping things simpler!

barmintor:
ok - not sure why the commit is still here, it is marked as in main with the same hash, but you can safely select the other commits as a range and have less noise about changed files.

app/javascript/src/hyacinth_v1/components/digital_objects/metadata/Field.js
```diff
@@ -10,17 +10,21 @@ import DateInput from '../../shared/forms/inputs/DateInput';
 import NumberInput from '../../shared/forms/inputs/NumberInput';
 import Checkbox from '../../shared/forms/inputs/Checkbox';
 
+const randomId = (prefix) => `${prefix}${Math.random()}`.replace('.','');
```
elohanlon, line 13: let uniqueFieldIdCounter = 0;

elohanlon, line 13:
```suggestion
let uniqueFieldIdCounter = 0;
```

app/javascript/src/hyacinth_v1/components/digital_objects/metadata/Field.js
```diff
@@ -10,17 +10,21 @@ import DateInput from '../../shared/forms/inputs/DateInput';
...
   const {
-    onChange, value, dynamicField, dynamicField: { displayLabel, fieldType }
+    inputName, onChange, value, dynamicField, dynamicField: { displayLabel, fieldType }
   } = props;
 
   let field = '';
 
-  const sharedProps = { onChange, value };
+  const sharedProps = { onChange, value, inputName };
+  sharedProps.inputName ||= randomId('input-');
```
elohanlon, line 23:
```suggestion
  sharedProps.inputName = sharedProps.inputName || `field-${uniqueCheckboxIdCounter++}`;
```
Two suggestions here:
1) Guaranteed unique id instead of a random one.
2) I didn't think `||=` works in js.  Was it working in your local copy?

elohanlon: Here's what threw me off:

```
let something ||= 'hello';
VM411:1 Uncaught SyntaxError: Unexpected token '||='
```
But:
```
let obj = {};
obj.something ||= 'hello';
"hello"
```

So I guess it only works for object properties.

elohanlon: fixed!

app/javascript/src/hyacinth_v1/components/digital_objects/rights/RightsEdit.js
```diff
@@ -33,14 +33,14 @@ function RightsEdit(props) {
     return (<GraphQLErrors errors={digitalObjectError || rightsFieldsError} />);
   }
 
-  const { digitalObject, digitalObject: { digitalObjectType } } = digitalObjectData;
+  const { digitalObject, digitalObject: { digitalObjectType }, digitalObject: { title } } = digitalObjectData;
```
elohanlon, line 36:
```suggestion
  const { digitalObject, digitalObject: { digitalObjectType, title } } = digitalObjectData;
```
I forgot -- does this work?

barmintor: I'm pretty sure nested destructures work, yes!

app/javascript/src/hyacinth_v1/components/digital_objects/rights/rights_form/subsections/DescriptiveMetadata.js
```diff
@@ -84,14 +85,10 @@ function DescriptiveMetadata(props) {
...
-          (descriptiveMetadata.title || [{}]).map((t, i) => (
-            <InputGroup key={i}>
-              <Label sm={4} align="right">Title</Label>
-              <ReadOnlyInput sm={8} value={t.sortPortion} />
-            </InputGroup>
-          ))
-        }
+        <InputGroup key="title_group">
+          <Label sm={4} align="right" for="rights_title_display">Title</Label>
+          <ReadOnlyInput sm={8} inputName="rights_title_display" value={(title || { value: {} }).value.sortPortion} />
```
elohanlon, line 90:
```suggestion
          <ReadOnlyInput sm={8} inputName="rights_title_display" value={title?.value?.sortPortion} />
```
I think this will work now that we have the optional chaining babel addition.

app/models/digital_object.rb
```diff
@@ -17,10 +17,9 @@ class DigitalObject < ApplicationRecord
   include DigitalObjectConcerns::ParentChildBehavior
   include DigitalObjectConcerns::ResourceImports
 
-  # Special structure for the title dynamic field
-  TITLE_DYNAMIC_FIELD_GROUP_NAME = 'title'
-  TITLE_SORT_PORTION_DYNAMIC_FIELD_NAME = 'sort_portion'
-  TITLE_NON_SORT_PORTION_DYNAMIC_FIELD_NAME = 'non_sort_portion'
+  # Access keys for the title attribute
+  TITLE_SORT_PORTION_KEY = 'sort_portion'
+  TITLE_NON_SORT_PORTION_KEY = 'non_sort_portion'
```
elohanlon, line 22: It seems like these constants aren't doing a lot for us.  They're only used in the generate_label method below.  Maybe it's worth getting rid of them?  Or if it's useful to have globally-accessible constants like these we should put them in the title TypeDef class?

Also, `'value'` is a hard-coded string in the `generate_label` method, so if these keys are treated as special constants then maybe that should be too.

But I'm also fine with just getting rid of the constants and using strings in the generate_label method if that simplifies things.

app/models/digital_object.rb
```diff
@@ -51,6 +50,8 @@ class DigitalObject < ApplicationRecord
     @mint_doi = false
   end
 
+  # Title
+  metadata_attribute :title, Hyacinth::DigitalObject::TypeDef::Title.new.validation(proc { |value| value&.dig('value').blank? || value.dig('value', 'sort_portion')&.strip.present? })
```
elohanlon, line 54: I just realized that the downside of doing validations via typedef (rather than regular model validation) is that you can't provide a helpful message about what went wrong.

app/models/digital_object/asset.rb
```diff
@@ -29,6 +29,15 @@ class DigitalObject::Asset < DigitalObject
 
   attr_accessor :skip_resource_request_callbacks
 
+  def generate_label
```
elohanlon, line 32:
```suggestion
  def generate_display_label
```
How would you feel about changing all occurrences of generate_label to generate_display_label?

lib/hyacinth/adapters/digital_object_search_adapter/solr/document_generator.rb
```diff
@@ -48,11 +48,18 @@ def merge_title_fields_for(digital_object, solr_document = {})
         end
 
         def merge_title_fields_for!(digital_object, solr_document = {})
-          solr_document['sort_title_ssi'] = digital_object.generate_title(true)
-          indexable_title = digital_object.generate_title
-          solr_document['title_ss'] = indexable_title
-          add_keywords(indexable_title, solr_document)
-          add_titles(indexable_title, solr_document)
+          display_label = digital_object.generate_label
```
elohanlon, line 51: Since label is often derived from the title, but not always, do you mind if we remove the display label stuff in here and add `'displayLabel_ss' => digital_object.generate_label` to the `merge_core_fields_for!` method?

elohanlon: Actually, I just noticed that display_label is also added to the set of indexed titles.  Hmm, that does complicate things a bit.  Here's a question that we haven't answered yet: will people think of the display label AS the title when a title hasn't been assigned to the object?

lib/hyacinth/adapters/preservation_adapter/fedora3/dc_properties.rb
```diff
@@ -22,7 +22,7 @@ def to(fedora_obj)
         def proposed_dc_properties(hyacinth_obj = @hyacinth_obj)
           properties = {}
           # set the title
-          properties[:title] = [get_title(hyacinth_obj.descriptive_metadata)]
+          properties[:title] = [get_title(hyacinth_obj)]
```
elohanlon, line 25: Should we be getting the display label instead of the title, in case there is no title?

lib/hyacinth/adapters/preservation_adapter/fedora3/object_properties.rb
```diff
@@ -8,7 +8,7 @@ class Fedora3::ObjectProperties
         include Fedora3::PropertyContextInitializers
 
         def to(fedora_obj)
-          fedora_obj.label = get_title(@hyacinth_obj.descriptive_metadata)
+          fedora_obj.label = get_title(@hyacinth_obj)
```
elohanlon, line 11: Same question here about using display_label instead of title.

lib/hyacinth/adapters/digital_object_search_adapter/solr/document_generator.rb
elohanlon: After looking at some other parts of the code below, I'm starting to think that the display_label is more of a title than the (possibly absent) title is.

app/javascript/src/hyacinth_v1/components/shared/forms/inputs/ReadOnlyInput.js
```diff
@@ -2,13 +2,16 @@ import React from 'react';
...
+
 class ReadOnlyInput extends React.PureComponent {
   render() {
-    const { value, ...rest } = this.props;
+    const { value, inputName, ...rest } = this.props;
 
     return (
       <Col sm={10} style={{ alignSelf: 'center' }} {...rest}>
         <Form.Control
+          id={inputName || `checkbox-${uniqueReadOnlyInputIdCounter++}`} // id is required for label to work with
```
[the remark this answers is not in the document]
barmintor: I prefer to do this in another commit, because there is at least one other instance of this message in code that was not modified in this branch.

lib/hyacinth/adapters/digital_object_search_adapter/solr/document_generator.rb
```diff
@@ -48,11 +48,18 @@ def merge_title_fields_for(digital_object, solr_document = {})
...
+          solr_document['displayLabel_ss'] = display_label
```
elohanlon, line 52:
```suggestion
          solr_document['display_label_ss'] = display_label
```
We've been going for snake case in solr keys to match the associated property names.

lib/hyacinth/adapters/digital_object_search_adapter/solr/document_generator.rb
```diff
@@ -48,11 +48,18 @@ def merge_title_fields_for(digital_object, solr_document = {})
...
+          if digital_object.title.present?
+            solr_document['title_sortPortion_ssi'] = digital_object.title.dig('value', 'sort_portion')
+            solr_document['sort_title_ssi'] = digital_object.title.dig('value', 'sort_portion')
+            solr_document['title_nonSortPortion_ssi'] = digital_object.title.dig('value', 'non_sort_portion')
```
elohanlon, line 56:
```suggestion
            solr_document['title_sort_portion_ssi'] = digital_object.title.dig('value', 'sort_portion')
            solr_document['sort_title_ssi'] = digital_object.title.dig('value', 'sort_portion')
            solr_document['title_non_sort_portion_ssi'] = digital_object.title.dig('value', 'non_sort_portion')
```
Suggesting snake_case for key consistency.

app/models/digital_object.rb
barmintor: This is true. We could probably refactor to be better about it, but out-of-scope here.

app/models/digital_object/asset.rb
barmintor: Done!

app/models/digital_object.rb
barmintor: Inlined!

app/javascript/src/hyacinth_v1/components/digital_objects/rights/rights_form/subsections/DescriptiveMetadata.js
barmintor: Done!

app/javascript/src/hyacinth_v1/components/digital_objects/rights/RightsEdit.js
barmintor: Done!

app/javascript/src/hyacinth_v1/components/digital_objects/metadata/Field.js
barmintor: Done!

app/javascript/src/hyacinth_v1/components/digital_objects/metadata/Field.js
barmintor: Done!

elohanlon (changes requested):
@barmintor Done with the review and this is great!  Thank you for all of your work on this.  I left some comments/suggestions.

lib/hyacinth/adapters/digital_object_search_adapter/solr/document_generator.rb
barmintor: It's not tested (!), but Hyacinth::DigitalObject::SolrKeys actually camelcases the string keys (as does the Rights indexing) - this ensures no unexpected collisions here. If we want to revisit that,, we should ticket separately. I'll add a test for the SolrKeys behavior, though!

app/javascript/src/hyacinth_v1/components/shared/forms/inputs/ReadOnlyInput.js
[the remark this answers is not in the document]
barmintor: Done!

app/models/digital_object.rb
elohanlon: 👍

lib/hyacinth/adapters/digital_object_search_adapter/solr/document_generator.rb
elohanlon: Oh, you're right!  Comment withdrawn for now then!  Maybe worth revisiting later.

elohanlon: And thank you for adding the test!

elohanlon (approved):
Thank you for the changes!

PR #359, opened 2021-08-24
Hyacinth 858 - fix and test identifier UI search

elohanlon:


Gemfile
```diff
@@ -57,7 +57,7 @@ group :development do
 end
 
 group :test do
-  gem 'capybara', '>= 2.15'
+  gem 'capybara', '~> 3.35'
```
elohanlon, line 60: We were already on capybara 3.35.3 (which is the latest version, so I updated the version here to make it clear that we're not on 2.x).

barmintor:
I'm going to review first thing Wednesday morning. Saw that tests failed - I am restarting in case it was a transient thing.

barmintor:
ok, see it failed again - likely suspect is the chrome install in this PR, right? Can we separate that into a different PR?

elohanlon:
@barmintor It was failing before, so the chrome install was an attempt to fix the failure (but no luck).  Tests pass locally, but not on GitHub Actions CI.  I'll try some other things and see if I can fix the CI run.

elohanlon:
@barmintor I changed the :system tests back to :feature tests and everything passes, so for the sake of development speed I'll keep them that way for now!  Feel free to review whenever is convenient!

app/javascript/src/hyacinth_v1/components/digital_objects/search/QueryForm.js
```diff
@@ -7,7 +7,7 @@ import SelectInput from '../../shared/forms/inputs/SelectInput';
 
 import SearchButton from '../../shared/buttons/SearchButton';
 
-const searchTypes = ['KEYWORD', 'TITLE', 'IDENTIFER'];
+const searchTypes = ['KEYWORD', 'TITLE', 'IDENTIFIER'];
```
barmintor, line 10: Well that's unfortunate.

barmintor (approved):
There are a couple of things here that I recommend doing in separate commits in the future (moving an unrelated test to a better location, improving a log message), as the commit history for the file will itself kind of violate the principle of least surprise - but this looks good to me.

elohanlon:
@barmintor Good suggestion!  I just pushed an update to split the changes up into multiple commits.

PR #362, opened 2021-08-25
HYACINTH-854 - add update child structure mutation

JackBlackLight:


elohanlon:
@JackBlackLight I was looking at this and got stuck on it for a bit too, but I believe I've found a way past it (and just pushed a commit)!

I renamed `Inputs::UpdateChildStructureInput` class to `Inputs::ChildStructureInput`.  It seems like there's some kind of name interference going on with the input class name and the the named `UpdateChildStructureInput!` input parameter in the usage of the mutation:

```
mutation ($input: UpdateChildStructureInput!) {
  ...
}
```

In addition to the rename fixing the problem, I also kind of prefer how `ChildStructureInput` describes what the data is rather than the context it's being used in.

elohanlon:
@JackBlackLight Also, here's the mutation + variables combo that I was using to test:

```
mutation ($input: UpdateChildStructureInput!) {
  updateChildStructure(input: $input) {
    parent {
      id
    }
  }
}
```

Variables:
```
{
  "input": {
    "parentId": "1",
    "orderedChildren": [
      {"uid": "abc-123", "sortOrder": 1},
      {"uid": "abc-456", "sortOrder": 0}
    ]
  }
}
```

Just mentioning because the variables in your earlier comment had a different structure (array brackets instead of object curly braces, and numeric uid values instead of string ones).

And now I get this error message when I run the query:

```
"message": "undefined method `ability_authorize!' for #<Mutations::DigitalObject::UpdateChildStructure:0x00007fcdada92360>"
```

But that's just because `ability_authorize!` needs to be changed to `ability.authorize!` in the mutation.

You're probably already aware of some/all of this because the work is in progress, but I just wanted to give some additional feedback while I was reviewing in case it helps!

JackBlackLight:
@elohanlon, how would you handle an exception within the resolve method of the update_child_structure mutation? right now, if a non-existent uid was provided for a child object, the code would call the .id method on a nil object. this would only occur due to an application error because the the user only rearranges the data provided and doesn't input any data.

elohanlon:
@JackBlackLight I would recommend using `find_by_uid!` instead of `find_by_uid`, so it'll raise a not-found exception that'll be caught by high level catch logic in HyacinthSchema.

As an example, see: https://github.com/cul/ldpd-hyacinth/blob/main/app/graphql/mutations/digital_object/delete_digital_object.rb#L12

Like you said, it would be an application error rather than a user error, so it's fine if we don't package up this error in a special way.  It'll just get turned into a top level graphql error.

elohanlon (approved):
Looks great!  I just merged in the related changes that you and I talked over and I'll merge this as soon as the CI run finishes.

PR #364, opened 2021-08-31
add a rubocop ban on capybara saving screenshots

barmintor:


lib/rubocop/cop/cul/capybara_screenshots.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+# TODO: when finished, run `rake generate_cops_documentation` to update the docs
```
elohanlon, line 3:
```suggestion
# TODO: when finished, run `rake generate_cops_documentation` to update the docs
```
TODO for later, or TODO already done?

elohanlon (approved):
Nicely done!  Only had one question about a TODO comment, but I'm approving anyway.  The `:new_cop` task seems very useful. 👍

lib/rubocop/cop/cul/capybara_screenshots.rb
elohanlon: Ben and I talked this over and this will remain as a TODO for later.

PR #369, opened 2021-09-01
HYACINTH-624 Send asynch resource creation requests to Triclops

barmintor:


barmintor:
I will clean up commits post-review, but I think this is ready to discuss otherwise. It does depend on a minor change to triclops to remove requirement of width and height params to create resource.

app/jobs/concerns/resource_request_jobs/iiif_job_behaviors.rb
```diff
@@ -0,0 +1,35 @@
...
+  module IiifJobBehaviors
+    extend ActiveSupport::Concern
+
+    class_methods do
+      def src_resource_for_digital_object(digital_object)
+        digital_object.asset_type == 'Image' ? digital_object.access_resource : digital_object.poster_resource
+      end
+
+      def eligible_object?(digital_object)
+        return false unless digital_object&.is_a?(::DigitalObject::Asset)
```
elohanlon, line 13:
```suggestion
        return false unless digital_object&.is_a?(::DigitalObject::Asset)
        return false unless digital_object.featured_thumbnail_region.present?
```
```suggestion
        return false unless digital_object&.is_a?(::DigitalObject::Asset)
```

elohanlon, line 13:
```suggestion
        return false unless digital_object&.is_a?(::DigitalObject::Asset)
        return false unless digital_object.featured_thumbnail_region.present?
```

app/jobs/concerns/resource_request_jobs/iiif_job_behaviors.rb
```diff
@@ -0,0 +1,35 @@
...
+        if digital_object.asset_type == 'Image'
+          digital_object.has_access_resource?
+        else
+          digital_object.has_poster_resource?
+        end
```
elohanlon, line 18: This check applies to the IiifRegistrationJob, but not the IiifDeregistrationJob.  In addition, we also want to check if `digital_object.featured_thumbnail_region.present?` in a IiifRegistrationJob, but don't need to for a IiifDeregistrationJob.

app/jobs/resource_requests/iiif_deregistration_job.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module ResourceRequests
+  class IiifDeregistrationJob < AbstractJob
+    include ResourceRequestJobs::IiifJobBehaviors
+
+    @queue = :image_syndication
```
elohanlon, line 7:
```suggestion
    @queue = :image_registration
```
Suggesting a rename in case this was missed, but also because I think we'll want registration (higher priority) to be on a different queue than deregistration (lower priority).

elohanlon, line 7:
```suggestion
    @queue = :image_deregistration
```
Suggesting a rename in case this was missed, but also because I think we'll want registration (higher priority) to be on a different queue than deregistration (lower priority).

app/jobs/resource_requests/iiif_registration_job.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module ResourceRequests
+  class IiifRegistrationJob < AbstractJob
+    include ResourceRequestJobs::IiifJobBehaviors
+
+    @queue = :image_syndication
```
elohanlon, line 7:
```suggestion
    @queue = :image_registration
```
Suggesting a rename in case this was missed, but also because I think we'll want registration (higher priority) to be on a different queue than deregistration (lower priority).

app/models/resource_request.rb
```diff
@@ -27,15 +33,9 @@ class ResourceRequest < ApplicationRecord
   # enqueue_derivativo_job ends up making requests to Derivativo which makes requests back to
   # Hyacinth and those requests fail when there's an existing lock on the squlite database.
   # See: https://flexport.engineering/how-to-safely-use-activerecords-after-save-efde2b52baa3
-  after_commit :enqueue_derivativo_job, on: :create
+  after_commit :run_create_callback, on: :create
```
elohanlon, line 36: I feel like the naming of `run_create_callback` is a little too literal, and could be confusing.  If feels similar to creating a method called `after_create_method` and binding it to the model's ActiveRecord `after_create` callback.  Also, I'm concerned that if someone on our team refers to the model's "create callback" that everyone else would assume they were talking about the ActiveRecord callbacks around create.

But I understand what `run_create_callback` is doing, and why you named it this way, and I am also finding it difficult to think of an alternate name.

You've introduced a cool way to dynamically add an after_create (or maybe more accurately, after_commit) hook on an instance of a ResourceRequest.  I'm going to think about this a little more and see if I can come up with an alternate name.

app/models/resource_request.rb
```diff
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
 class ResourceRequest < ApplicationRecord
+  attr_accessor :create_callback
```
elohanlon, line 4: As a follow up to my other comment about `create_callback`, the naming is also a little problematic because the attr name doesn't indicate whether it's before create or after create.  (But also, it's really after commit!)

app/models/resource_request.rb
```diff
@@ -27,15 +33,9 @@ class ResourceRequest < ApplicationRecord
   # enqueue_derivativo_job ends up making requests to Derivativo which makes requests back to
```
elohanlon, line 33: Just a note here that the old `enqueue_derivativo_job` still exists in this comment.

app/jobs/concerns/resource_request_jobs/derivativo_job_behaviors.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module ResourceRequestJobs
+  module DerivativoJobBehaviors
+    extend ActiveSupport::Concern
+
+    class_methods do
+      def create_callback(resource_request, _digital_object)
```
elohanlon, line 8: Sorry to be so picky about naming, but when I started reading this method I thought it might be a factory method that returns return a proc (i.e. I thought it would "create a callback").

elohanlon:
Proposed changes here in this PR: https://github.com/cul/ldpd-hyacinth/pull/375

elohanlon:
I like what you've done here!  I have some proposed changes here in this PR (https://github.com/cul/ldpd-hyacinth/pull/375), but we're very close to merging.

elohanlon (changes requested):
(Done with review.)

elohanlon (approved):
🚀

PR #371, opened 2021-09-07
HYACINTH-866: Digital Object structured child GraphQL update

elohanlon:


app/javascript/src/hyacinth_v1/components/digital_objects/children/Children.js
```diff
@@ -23,7 +23,7 @@ const Children = (props) => {
 
   if (childStructureLoading) return (<></>);
   if (childStructureError) return (<GraphQLErrors errors={childStructureError} />);
-  const { childStructure: { parent, structure } } = childStructureData;
+  const { digitalObject: parent, digitalObject: { childStructure: { structure } } } = childStructureData;
   return (
```
JackBlackLight, line 27: @elohanlon, could you explain the rationale for this change? Everything else looks great and is very clear, I just wasn't sure why this change is preferable (I'm not objecting - just asking for clarification). Since this is just an info request, I'm going to go ahead and approve this PR. Thanks!

JackBlackLight (approved):

app/javascript/src/hyacinth_v1/components/digital_objects/children/Children.js
elohanlon: @JackBlackLight Thank you for reviewing!  The reason for the above change is that now that the childStructure is a sub-property of a digital object (instead of, previously, being something that was returned by a dedicated child structure query), I needed to separate out selection of the digitalObject and the childStructure since both of those entities are used later in the component.

Technically I could have just selected the digitalObject alone and used digitalObject.childStructure later in the component though, and that might be clearer than the unusual destructuring above (where I refer to `digitalObject` twice).  I'll actually revise this and push an amended commit since I think it'll make things clearer.  I'm glad you asked about it!

elohanlon: @JackBlackLight I just pushed an update!  Hopefully that should look better.  I also updated some variable names so they were more similar to other Digital Object tabs.

PR #372, opened 2021-09-09
HYACINTH-798 gql and ui for enabled publish targets

barmintor:


app/graphql/mutations/update_project_publish_targets.rb
```diff
@@ -0,0 +1,35 @@
...
+    # This should be an all or nothing update
+    ActiveRecord::Base.transaction do
+      project = Project.find_by!(project.to_h)
+      # Ensure that the user initiating this update is allowed to do so for the given project
+      ability.authorize! :update, project
+
+      publish_targets = PublishTarget.where(string_key: publish_targets.map { |pt| pt['stringKey'] })
+      project.publish_targets = publish_targets.to_a
+      project.save!
+    end
```
elohanlon, line 25: Does this actually need to be in a transaction, since you're only performing one update query?

elohanlon: If for some reason the update does need to be in a transaction, I think you can at least move some of the earlier lines out of it to minimize transaction time.

elohanlon: Oh, maybe you're wrapping in a transaction because `project.publish_targets = publish_targets.to_a` and then `project.save!` updates the updated_at timestamp?  In that case, maybe only those two lines need to be in the transaction.

elohanlon: Oh, maybe you're wrapping in a transaction because `project.publish_targets = publish_targets.to_a` instantly performs a query and then `project.save!` updates the updated_at timestamp in a separate query?  In that case, I think only those two lines need to be in the transaction.

app/graphql/types/query_type.rb
```diff
@@ -34,6 +34,11 @@ class QueryType < Types::BaseObject
       description "List of all projects"
     end
 
+    field :projects_publish_targets, [ProjectsPublishTargetType], null: true do
```
elohanlon, line 37: Should this query be called "project_publish_targets" rather than "projects__publish_targets"?  And should it return a `[PublishTargetType]` rather than `[ProjectsPublishTargetType]` since there is no ProjectsPublishTarget model.

elohanlon:
@barmintor I haven't finished reviewing this PR, but felt like I should pause to note that it seems like you're building around an intermediate model between Projects and PublishTargets that doesn't currently exist -- especially with the `ProjectsPublishTargetType` GraphQL type, which has an "enabled" field.

It's not that I think this is a bad idea (since after all, we actually DO have an intermediate EnabledPublishTarget model in Hyacinth 2), but if we want to go this route then it seems like we should add the EnabledPublishTargetModel and change the HABTM relationship to a has_many :through.  And then I'd suggest renaming your `ProjectsPublishTargetType` to `EnabledPublishTargetType` because it would make things clearer.

How do you feel about that?

app/graphql/types/query_type.rb
elohanlon: (also see my follow-up comment in the Conversation tab)

barmintor: Remember the graphql API is not necessarily tied to the backing model - I don't think there's a need (or benefit, really) to having a model for the join entity if it has no other attributes. The `enabled` annotation here is just sugar to prevent the inspection from happening more than is necessary.

app/graphql/mutations/update_project_publish_targets.rb
barmintor: No, nothing like that - I'm adhering to what I think we probably want to encourage as a general pattern in the UI-driven update layer. You're right that we could shave lines off here - as you pointed out to me, the HABTM relationship automatically saves, so you could technically drop that `save!` line, too. But the resulting method would, in my view, reward high-context maintainers (able to infer the HABTM behavior, etc) and punish low-context maintainers ("why aren't they saving this record? how are these tests passing?"), and the latter (ie, team members who do not spend lots of contiguous time in this codebase) is I think the more likely audience. This isn't an absolute position, but it's something to consider: What balance of performance, legibility and maintainability do we want to arrive at? What should we fix with comments/documentation? The balance is much different for batch processing, I think, than in the gql layer of individual updates.

barmintor:
I prefer not to have the model - it really adds nothing that I can see - but I am sympathetic to (and honestly expected you to suggest) renaming that gql type that annotates a publish target record. I just don't have a good name - I was avoiding `Enabled-` and didn't think it should hold up opening the PR. `FlaggedPublishTargetType`? `Connected-`? Naming things is slower than implementing.

barmintor:
The gql type is essentially a left join result in the context of a project. So... it's not the `Enabled-` ones. It's all of the PublishTargets, with a flag indicating the ones that also have the association. `EnabledPublishTarget` with an attribute called `enabled` that can be false is kind of awkward.

app/graphql/mutations/update_project_publish_targets.rb
elohanlon: I understand what you mean, and it is easy to forget that has_many/habtm relationships are automatically added.  How about we just take the non-updating lines out of the transaction then?  That seems like a good practice regardless of rails-specific knowledge.

app/graphql/types/query_type.rb
elohanlon: I didn't forget.  And though a graphql API doesn't have to be tied to the backing model, we should be careful about building an API that seems to contradict the backing model.  I think this top level query will confuse developers, since I myself found it confusing!

When a Project has a PublishTarget on the backend, that means that the PublishTarget is enabled.  So why does this query, which returns a list of publish target information for the given Project, return a mixed list of enabled and disabled publish targets?

See what I'm getting at?  It all comes down to the naming here.  How about we call the type `AvailablePublishTargetType`?  I think that could help a lot!

Also, I wonder...does this even need to be a dedicated top level query?  Or could availablePublishTargets just be a field that we add to the ProjectType?

elohanlon:
> I prefer not to have the model - it really adds nothing that I can see - but I am sympathetic to (and honestly expected you to suggest) renaming that gql type that annotates a publish target record. I just don't have a good name - I was avoiding `Enabled-` and didn't think it should hold up opening the PR. `FlaggedPublishTargetType`? `Connected-`? Naming things is slower than implementing.

I understand your argument against the model.  In fact, that's the reason why I implemented the relationship with HABTM instead of `has_many :through`!  I was just suggesting it because your code made me think you were thinking along those lines, but I now see that was not the case!

elohanlon:
> The gql type is essentially a left join result in the context of a project. So... it's not the `Enabled-` ones. It's all of the PublishTargets, with a flag indicating the ones that also have the association. `EnabledPublishTarget` with an attribute called `enabled` that can be false is kind of awkward.

Please see my other comment where I suggested `AvailablePublishTargetType`.  Maybe a good option?  (Along with perhaps switching from top level query field to field within ProjectType.)

app/graphql/types/projects_publish_target_type.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module Types
+  class ProjectsPublishTargetType < Types::PublishTargetType
```
elohanlon, line 4: I cannot say this with great authority, but it seems like subclassing a graphql type in this manner may be working against the general graphql patterns I've seen.  I'm suspicious of having two almost identical types with only one field difference.  Should we instead be defining two fields in this type, `:enabled` and `:publish_target`, and then the client would just include `publishTarget {}` in a query?

app/javascript/src/hyacinth_v1/components/projects/publish_targets/PublishTargetsEdit.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import { startCase } from 'lodash';
+import produce from 'immer';
+
+import TabHeading from '../../shared/tabs/TabHeading';
+import { PublishTargetsForm } from './PublishTargetsForm';
+import { projects } from '../../../utils/hyacinthApi';
+import ProjectInterface from '../ProjectInterface';
+
+export default class PublishTargetsEdit extends React.Component {
```
elohanlon, line 10: We're trying to go with functional components for all new components.  Why go class-based for this one?  Is there something that didn't work in a functional context?

app/javascript/src/hyacinth_v1/components/projects/publish_targets/PublishTargetsEdit.js
```diff
@@ -0,0 +1,40 @@
...
+  state = {
+    project: {},
+  }
+
+  componentDidMount() {
+    const { match: { params: { projectStringKey } } } = this.props;
+
+    projects.get(projectStringKey)
```
elohanlon, line 18: We're really trying not to add more references to the old REST api in the app.  Is there a reason we can't use the project graphql query?  (see: `getProjectQuery` in `app/javascript/src/hyacinth_v1/graphql/projects.js`).

app/javascript/src/hyacinth_v1/components/projects/publish_targets/PublishTargetsForm.js
```diff
@@ -0,0 +1,158 @@
...
+  useEffect(() => {
+    if (projectsPublishTargetsLoading === false && projectsPublishTargetsData) {
+      const mappedQueryData = mapPublishTargetData(projectsPublishTargetsData);
+      setProjectPublishTargets(mappedQueryData);
+    }
+  }, [projectsPublishTargetsLoading, projectsPublishTargetsData]);
+
+  const [updateEnabledPublishTargets, { error: updateError }] = useMutation(updateProjectsPublishTargetsMutation);
+
+  if (projectsPublishTargetsLoading) return (<></>);
```
elohanlon, line 79: If you move `if (projectsPublishTargetsLoading) return (<></>);` above your useEffect, will that simplify things?  Then you wouldn't need to check if `projectsPublishTargetsLoading === false && projectsPublishTargetsData` inside the useEffect function.

app/javascript/src/hyacinth_v1/components/projects/publish_targets/PublishTargetsForm.js
```diff
@@ -0,0 +1,158 @@
...
+import {
+  Row, Col, Form, Badge,
+} from 'react-bootstrap';
+import { useHistory } from 'react-router-dom';
+import produce from 'immer';
+import GraphQLErrors from '../../shared/GraphQLErrors';
+import FormButtons from '../../shared/forms/FormButtons';
+import { getProjectsPublishTargetsQuery, updateProjectsPublishTargetsMutation } from '../../../graphql/projects/publishTargets';
+
+const PublishTarget = (props) => {
```
elohanlon, line 13:
```suggestion
const AvailablePublishTarget = (props) => {
```

This suggestion is related to my earlier renaming suggestion.

app/javascript/src/hyacinth_v1/components/projects/publish_targets/PublishTargetsForm.js
barmintor: I was working around an infinite rerender from the state modification using the pattern you describe, but maybe I missed something?

app/graphql/mutations/update_project_publish_targets.rb
barmintor: Sounds like a good compromise!

app/graphql/types/query_type.rb
barmintor: This was really easy to implement as a single list that you replace wholesale; it would be more complicated to move part of it around - I'm reading your suggestion to be that the project type has `publish_targets` and `available_publish_targets`. Or are you thinking that available publish targets acts entirely the way it does here, as a single list? I'd be fine with that, I think.

app/javascript/src/hyacinth_v1/components/projects/publish_targets/PublishTargetsEdit.js
barmintor: Just an oversight!

app/javascript/src/hyacinth_v1/components/projects/publish_targets/PublishTargetsEdit.js
barmintor: Nope, just expediency in implementation - the closest analog was a class-based component.

app/graphql/mutations/update_project_publish_targets.rb
barmintor: Moved reads/authorize outside of transaction.

app/javascript/src/hyacinth_v1/components/projects/publish_targets/PublishTargetsEdit.js
barmintor: Converted to functional component.

app/javascript/src/hyacinth_v1/components/projects/publish_targets/PublishTargetsEdit.js
barmintor: Done!

app/javascript/src/hyacinth_v1/components/projects/publish_targets/PublishTargetsForm.js
barmintor: Done!

app/graphql/types/query_type.rb
barmintor: There is not a selectable field called `available_publish_targets` on Project that handles this.

app/javascript/src/hyacinth_v1/components/projects/publish_targets/PublishTargetsForm.js
barmintor: Moving the `availablePublishTargets` query to a field on Project allowed me to roll the query up in the calling component, which removes the necessity of using state *and* query *and* mutation here, so the effect could be removed.

elohanlon (changes requested):
Thank you for all of the renaming changes.  I have the additional changes we talked about in a proposal PR here (https://github.com/cul/ldpd-hyacinth/pull/374), along with some extra test stuff I noticed along the way.

elohanlon (approved):

PR #374, opened 2021-09-14
Test refactoring, shared example renaming, and some related test cleanup

elohanlon:


app/graphql/types/query_type.rb
```diff
@@ -246,14 +246,6 @@ def projects
       Project.accessible_by(ability)
     end
 
-    def projects_publish_targets(params)
```
barmintor, line 249: did I just forget to remove this query?

barmintor (approved):

PR #379, opened 2021-09-20
Hyacinth 884 and hyacinth 885 title updates (and HYACINTH-886 sample data rake task refactoring/updates)

elohanlon:


lib/tasks/hyacinth/setup.rake
```diff
@@ -32,102 +32,6 @@ namespace :hyacinth do
         end
       end
     end
-
-    desc "Set up test projects"
-    task test_projects: :environment do
```
elohanlon, line 37: I removed this because I found that the test projects were just getting in the way most of the time, since all of the sample content was in the sample content project.  I can easily add a second sample project if useful though.  Or we can add it later on.

barmintor: Sounds good to me.

barmintor: The true test projects should be handled via factories anyway.

spec/models/digital_object_spec.rb
```diff
@@ -361,9 +361,11 @@
           instance.assign_attributes('title' => nil)
           expect(instance).to be_valid
         end
+        it 'is valid when the title only contains a blank non_sort_portion (because blank title fields are removed in before_validation callback)' do
```
barmintor, line 364: I found this spec description a little confusing - this may be one of those "too much information" situations, or it may just be a typo. I'll send a commit up with some recommended changes.

barmintor (approved):
This looks good - I appreciate the test/sample data fixups!

PR #382, opened 2021-09-20
Hyacinth 845 missing value field group array

barmintor:


elohanlon (approved):
This is great!  Thank you for fixing this.  I especially liked the spec.js test.  Looking forward to adding more of those in the future!

PR #383, opened 2021-09-21
HYACINTH-889 - jest setup plus CI

elohanlon:


elohanlon:
@barmintor I'm just opening as a draft PR for now, but would appreciate any thoughts.  The overall jest setup and JavaScript unit tests seem to be working well, but the js end-to-end browser tests are only running locally at the moment -- not passing on GitHub Actions.

This work on the e2e tests is making me reconsider whether we want to do browser feature testing with jest/javascript after all, since it depends on a running rails backend anyway, and we'd need to think through how to clear solr and the database between tests, etc.

.github/workflows/ci.yml
```diff
@@ -43,7 +43,68 @@ jobs:
...
+        run: yarn run test
+
+  ci-javascript-e2e:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: ['10', '12']
+    steps:
+      - uses: actions/checkout@v2
+      ### BEGIN Rails App Dependencies ###
```
elohanlon, line 75: FYI: This `### BEGIN Rails App Dependencies ###` (and corresponding `### END Rails App Dependencies ###`) is just here temporarily because these are an exact copy of the same steps in the earlier job.  I'd ideally like to find a way to share these common steps if we end up keeping the e2e tests (which we might not!).

barmintor: I think it would be wise to break the e2e actions out into a separate ticket/PR. But to your point about shared config: I don't see anything in the [action docs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions) that suggests named references aren't supported in the YML, so that should be an option (include all of a named array of step configs, and add additional inline).

app/javascript/src/hyacinth_v1/App.js
```diff
@@ -1,63 +1,14 @@
...
 import { BrowserRouter } from 'react-router-dom';
 import { QueryParamProvider } from 'use-query-params';
-import { ApolloProvider } from '@apollo/react-hooks';
-import axios from 'axios';
 
 import MainContent from './MainContent';
-import createApolloClient from './utils/createApolloClient';
 
 function App() {
-  const [loading, setLoading] = useState(true);
```
barmintor, line 11: Most of the below has been extracted into `app/javascript/src/hyacinth_v1/utils/getGraphqlSchemaTypes.js`, which is now called in an event handler added in the main entry file `app/javascript/packs/hyacinth_v1.js`

.github/workflows/ci.yml
barmintor: Oh, right: There's not a good way to merge sequences in yml. Hmm. Let me try some stuff.

barmintor: I think the best we could do would be to break the reusable tasks out into a separate list and have an anchor for each one - that would be some improvement.

barmintor: And we can't do that because github actions doesn't support anchors. So that's that then!

app/javascript/src/hyacinth_v1/App.js
elohanlon: 👍

barmintor (approved):
This looks great, and is a real step forward for the application!

PR #386, opened 2021-09-22
change string key of rights group restriction_on_access to asset_access_restrictions

barmintor:
see also HYACINTH-871

Deploying this change will require updating the database to change the relevant dynamic field group's string_key


elohanlon (changes requested):
Thanks for fixing this!  One minor note: For better or for worse, we generally use singular rather than plural names for dynamic fields/groups.  How would you feel about changing `asset_access_restrictions` to `asset_access_restriction` for consistency?

barmintor:
Don't mind at all.

elohanlon (approved):
Thanks!

PR #388, opened 2021-09-28
Proposal for renaming datacite rest API methods

elohanlon:
I thought it might be good to name the API methods for what they do rather than the REST verbs they use (like how in controllers we name the action methods "create" or "update" instead of "post" or "put").  Putting this out there as a change proposal.


docs/datacite.README.md
```diff
@@ -88,29 +88,28 @@ Note that the Api libraries are designed to operate against the external_identif
...
-2.6.4 :ZZZ > attributes[:titles] = [{ title: "A Great Book"}]
-2.6.4 :ZZZ > attributes[:creators] = ['Mouse, Katie', 'Mouse, John'].map {|v| { name: v } }
-2.6.4 :ZZZ > attributes[:publicationYear] = 2021
-2.6.4 :ZZZ > attributes[:publisher] = 'Columbia University'
-2.6.4 :ZZZ > attributes[:types] = { resourceTypeGeneral: 'Text' }
-2.6.4 :ZZZ > attributes[:url] = "https://www.columbia.edu"
-2.6.4 :003 > data = { type: :"dois", attributes: attributes }
-2.6.4 :003 > payload = JSON.generate(data: data)
-2.6.4 :012 > resp = api.put_dois('10.33555/6zy2-pw34', payload)
+2.6.4 :003 > attributes = { prefix: "10.33555" }
```
barmintor, line 103: 😆  thank you for fixing my line numbers!

barmintor (approved):
This change makes sense to me. I'm tempted to suggest dropping the `_doi` suffixes, but the Datacite REST API does have some different resource types, so maybe worth holding off on that.

barmintor:
I'm thinking a little more about this, and wondering if we might actually want to introduce new methods with these names that wrap the verbs, and hide the JSON response parsing from the client code. Does that make sense?

barmintor (commented):
I'm thinking a little more about this, and wondering if we might actually want to introduce new methods with these names that wrap the verbs, and hide the JSON response parsing from the client code. Does that make sense?

fcd1 (approved):

elohanlon:
Thank you both for the review!  As per our conversation, I've made some additional improvements!  I'll wait until the tests pass and then I'll merge the changes.

PR #392, opened 2021-10-01
preserve, publish behaviors should only acquire lock specific to func…

barmintor:
…tion

- this prevents having to release object lock to save within scope of function
- HYACINTH-903
- HYACINTH-904

### DIFF
--- a/app/models/concerns/digital_object_concerns/preserve_behavior.rb
+++ b/app/models/concerns/digital_object_concerns/preserve_behavior.rb
@@ -9,14 +9,14 @@ module DigitalObjectConcerns
       # No one should be preserving an object that has errors
       raise Hyacinth::Exceptions::InvalidPersistConditions, 'Cannot persist a DigitalObject that has errors' if self.errors.present?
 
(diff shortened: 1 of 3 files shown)


app/models/concerns/digital_object_concerns/preserve_behavior.rb
```diff
@@ -9,14 +9,14 @@ def preserve
...
-      Hyacinth::Config.lock_adapter.with_lock(self.uid) do
+      Hyacinth::Config.lock_adapter.with_lock("#{self.uid}-preserve") do |_lock_object|
         assign_and_save_doi_if_blank
 
         preservation_result, errors = Hyacinth::Config.preservation_persistence.preserve(self)
 
         unless preservation_result
           self.errors.add(:preservation, "Failed to preserve #{self.uid} due to the following errors: #{errors.join(', ')}")
-          return
+          break
```
elohanlon, line 19: If we break here instead of returning, that will lead to a preservation timestamp update even though the preservation failed.

elohanlon (approved):

elohanlon (approved):

PR #393, opened 2021-10-04
Add a graphql mutation to update the projects a DigitalObject is associated with

barmintor:
HYACINTH-392


app/graphql/mutations/digital_object/update_projects.rb
```diff
@@ -0,0 +1,68 @@
...
+        Project.find_by!(string_key: proposed_string_key) unless digital_object.primary_project.string_key.eql(proposed_string_key)
+      end
+
+      def authorize_change!(current_primary, proposed_primary)
+        if proposed_primary
+          ability.authorize! :delete_objects, current_primary
+          ability.authorize! :create_objects, proposed_primary
+        else
+          ability.authorize! :update_objects, current_primary
+        end
```
elohanlon, line 44: If a user is updating both the primary project and other_projects during the same request, this won't verify that they have update permission.  Update permission is required for modifying other_projects.

barmintor: I am hypothesizing that no one has `delete` but not `update` to save some work, but they are technically separate permissions.

barmintor: Updated to check permissions for `update` and `delete` if both the primary and other projects are changing.

spec/requests/graphql/mutations/digital_object/update_projects_spec.rb
```diff
@@ -0,0 +1,146 @@
...
+          id: authorized_item.uid,
+          primaryProject: {
+            stringKey: secondary_project.string_key
+          },
+          otherProjects: [
+            { stringKey: primary_project.string_key }
+          ]
+        }
+      }
+    end
```
elohanlon, line 81: If you're only updating the primary_project in this test, can we remove the otherProjects key-value pair from the variables?

app/graphql/mutations/digital_object/update_projects.rb
```diff
@@ -0,0 +1,82 @@
+# frozen_string_literal: true
+
+module Mutations
+  module DigitalObject
+    class UpdateProjects < Mutations::BaseMutation
+      argument :id, ID, required: true
+      argument :primary_project, Inputs::StringKey, required: true
```
elohanlon, line 7:
```suggestion
      argument :primary_project, Inputs::StringKey, required: false
```
Should this be `required: false` so users can use this mutation to update only the primary_project without having to supply other_projects?

app/graphql/mutations/digital_object/update_projects.rb
```diff
@@ -0,0 +1,82 @@
...
+      argument(
+        :other_projects,
+        [Inputs::StringKey],
+        "An array representing the new set of other projects for the given object, "\
+        "or an empty array to remove all projects but the primary.",
+        required: true
```
elohanlon, line 13:
```suggestion
        required: false
```
Should this be `required: false` so users can use this mutation to update only the other_projects without having to supply primary_project?

app/graphql/mutations/digital_object/update_projects.rb
```diff
@@ -0,0 +1,82 @@
...
+      )
+
+      field :digital_object, Types::DigitalObjectInterface, null: true
+      field :user_errors, [Types::Errors::FieldedInput], null: false
+
+      def resolve(id:, primary_project:, other_projects:)
```
elohanlon, line 19: If primary_project and other_projects are each optional for this mutation, the logic below would have to change to accommodate that.

spec/requests/graphql/mutations/digital_object/update_projects_spec.rb
```diff
@@ -0,0 +1,146 @@
...
+          id: authorized_item.uid,
+          primaryProject: {
+            stringKey: primary_project.string_key
+          },
+          otherProjects: [
+            { stringKey: tertiary_project.string_key }
+          ]
+        }
+      }
+    end
```
elohanlon, line 27: Can we remove the primaryProject key-value pair from the variables to focus on only updating other_projects?

spec/requests/graphql/mutations/digital_object/update_projects_spec.rb
elohanlon, line 81: Can we remove the otherProjects key-value pair from the variables to focus on only updating primaryProject?

spec/requests/graphql/mutations/digital_object/update_projects_spec.rb
barmintor: The three cases (update primary, update other, update both) are now separately tested.

spec/requests/graphql/mutations/digital_object/update_projects_spec.rb
barmintor: The three cases (update primary, update other, update both) are now separately tested.

app/validators/digital_object/projects_validator.rb
```diff
@@ -0,0 +1,12 @@
...
+
+class DigitalObject::ProjectsValidator < ActiveModel::Validator
+  def validate(digital_object)
+    primary_string_key = digital_object.primary_project&.string_key
+    if digital_object.other_projects&.detect { |proj| proj.string_key == primary_string_key }
+      digital_object.errors.add(:other_projects, "Other projects cannot also be primary: #{primary_string_key}")
+      return false
+    end
+    true
+  end
```
elohanlon, line 11:
```suggestion
  def validate(digital_object)
    primary_string_key = digital_object.primary_project&.string_key
    if digital_object.other_projects&.detect { |proj| proj.string_key == primary_string_key }
      digital_object.errors.add(:other_projects, "Other projects cannot also be primary: #{primary_string_key}")
      return
    end
  end
```
Looks like we don't need to return true or false from the validate method.  Adding or not adding errors is all that's required: https://api.rubyonrails.org/classes/ActiveModel/Validator.html

spec/models/concerns/digital_object_concerns/attribute_assignment/projects_spec.rb
```diff
@@ -21,14 +22,26 @@
...
+  describe "#assign_projects" do
     it "sets the projects each time it's called" do
       digital_object.assign_projects(digital_object_data_with_projects)
       expect(digital_object.projects).to eq([different_primary_project] + other_projects)
 
       digital_object.assign_projects(digital_object_data_with_different_projects)
       expect(digital_object.projects).to eq([another_different_primary_project] + other_projects)
     end
+    context "with a non-existent other project string_key" do
+      let(:expected_error) { Hyacinth::Exceptions::NotFound }
```
elohanlon, line 41: Completely optional suggestion, but maybe the `let(:expected_error)` doesn't buy you a lot here compared to just passing the error class directly to the `raise_error` method?

spec/support/authenticate_user.rb
```diff
@@ -20,7 +20,9 @@ def sign_in_project_contributor(**args)
   end
 
   def create_project_contributor(to:, project:)
-    permissions = Array.wrap(to).map { |action| Permission.new(action: action, subject: Project.to_s, subject_id: project.id) }
+    permissions = Array.wrap(to).product(Array.wrap(project)).map do |action, subject|
```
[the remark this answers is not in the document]
barmintor: do you prefer `(to:, for:)` or `(actions:, projects:)`?

spec/models/concerns/digital_object_concerns/attribute_assignment/projects_spec.rb
elohanlon: Sounds good!

spec/support/authenticate_user.rb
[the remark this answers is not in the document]
elohanlon: `(actions:, projects:)` would be much better.  Great suggestion!

PR #396, opened 2021-10-11
HYACINTH-821 adds ui and auth check for add/remove parent objects

JackBlackLight:


app/graphql/mutations/digital_object/remove_parent.rb
```diff
@@ -11,7 +11,7 @@ class RemoveParent < Mutations::BaseMutation
       def resolve(id:, parent_id:)
         digital_object = ::DigitalObject.find_by_uid!(id)
         parent_object = ::DigitalObject.find_by_uid!(parent_id)
-        ability.authorize! :update, parent_object, :read, digital_object
+        ability.authorize! :update, parent_object, :read, digital_object, :message, "Not authorized to remove parent object"
```
elohanlon, line 14: Same note here about needing to add a test to validate this authorization.

PR #398, opened 2021-10-14
Hyacinth 913

barmintor:
fix linter issues in components/digital_objects/DigitalObjects.js, components/digital_objects/DigitalObjectSearch.js
use consistent param parsing in React navigation search results (HYACINTH-913)
- separate more clearly internal search param names and URL query param names
- components/digital_objects/DigitalObjectInterface.js: move history hook and result paging behavior into components/shared/ResultsPagingBar.js
- components/digital_objects/DigitalObjectList.js: move sessionStorage of query and current result offset to reusable functions in utils/digitalObjectSearchParams.js
- components/digital_objects/DigitalObjectList.js: remove unused props
- components/digital_objects/DigitalObjectSearch.js: extract reusable and testable functions in utils/digitalObjectSearchParams.js
- components/digital_objects/DigitalObjectSearch.js: consolidate search param update logic
- components/digital_objects/DigitalObjectSearch.js: refactor useEffect history hook
- components/digital_objects/DigitalObjectSearch.js: remove unnecessary useState hooks
- components/digital_objects/DigitalObjectSearch.js: remove HOC-style wrapper for query params provider from default export
- utils/digitalObjectSearchParams.js: provide defaultParams from query string parsing configuration
- utils/encodeAndStringifySearch.js: deleted; moved relevant functions to utils/digitalObjectSearchParams.js
- utils/filterArrayParam.js: encode returns empty array if filters arg is blank
- SearchButton.js is given aria-label in components/digital_objects/search/QueryForm.js
- jest specs: utils/digitalObjectSearchParams.js
- feature specs: direct navigation to an item by UID does not link back to a search
- feature specs: using back button in digital object search after changing search terms
- feature specs: navigating to digital object from search presents link back to search

Shim missing Array.prototype.flat, Array.prototype.flatMap, Object.fromEntries for node 10 (HYACINTH-914)

### DIFF
--- a/app/javascript/src/hyacinth_v1/components/digital_objects/DigitalObjectSearch.js
+++ b/app/javascript/src/hyacinth_v1/components/digital_objects/DigitalObjectSearch.js
@@ -1,4 +1,4 @@
-import React, { useEffect } from 'react';
+import React from 'react';
 import { useHistory } from 'react-router-dom';
 import {
   Card, Col, Row,
@@ -8,7 +8,6 @@ import { useQueryParams } from 'use-query-params';
 import {
   decodedQueryParamstoSearchParams,
   encodeSessionSearchParams,
-  locationSearchToSearchParams,
   queryParamsConfig,
   searchParamsToLocationSearch,
 } from '../../utils/digitalObjectSearchParams';
@@ -44,7 +43,7 @@ const DigitalObjectSearch = () => {
   const [queryParams] = useQueryParams(queryParamsConfig);
   const searchParams = decodedQueryParamstoSearchParams({ ...queryParams });
   const {
-    loading, error, data, refetch,
+    loading, error, data,
   } = useQuery(
     getDigitalObjectsQuery, {
       variables: searchParamsToVariables(searchParams),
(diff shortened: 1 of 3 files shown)


barmintor:
Jest specs for the utility module are in-progress

barmintor:
ok, shims in for missing methods.

elohanlon:
@barmintor Go for it!

package.json
```diff
@@ -69,7 +70,10 @@
     "eslint-plugin-testing-library": "^4.12.3",
     "jest": "^27.2.1",
     "jest-fetch-mock": "^3.0.3",
+    "jest-puppeteer": "^5.0.4",
```
elohanlon, line 73: I just noticed that this was in here (added by me when I first set up jest stuff).  We can actually remove this (and the puppeteer module), since it's only relevant to the unmerged jest feature tests in https://github.com/cul/ldpd-hyacinth/tree/experimental-jest-e2e-testing.

barmintor: when I squashed I included `yarn remove` on the two puppeteer-related dependencies, thanks for catching this!

barmintor:
Things are looking ok with the shims - do we think this is mergeable?

elohanlon:
I'll take a look now.

app/javascript/src/hyacinth_v1/components/digital_objects/search/FacetSidebar.spec.js
```diff
@@ -0,0 +1,33 @@
+import renderer from 'react-test-renderer';
+import FacetDropdown from './facet_sidebar/FacetDropdown';
+import FacetSidebar from './FacetSidebar';
+
+jest.mock('./facet_sidebar/FacetDropdown', () => ({
+  __esModule: true, default: jest.fn(() => 'FacetDropdown'),
+}));
```
elohanlon, line 7: I might be wrong about this, but I _think_ this mock will leak across other tests/files.  The `jest.spyOn` method is a good alternative to `jest.mock` because it's easy to reset (and generally you'd want to call it in a beforeEach), but I don't think we can use it here because you're mocking the default export of a module rather than a function on an object.  Let's talk more about this tomorrow if you have a chance.

PR #401, opened 2021-10-21
Hyacinth 907: facet value search in slide out window

barmintor:


app/controllers/concerns/hyacinth/users/cas_authentication_behavior.rb
```diff
@@ -60,11 +60,12 @@ def identify_uni_user(user_uni, cas_logout_uri)
...
-                flash[:alert] = I18n.t 'devise.failure.inactive'
-              else
-                flash[:alert] = 'The UNI ' + user_uni + ' is not authorized to log into Hyacinth (no account exists with email ' + user_uni + '@columbia.edu).  If you believe that you should have access, please contact an application administrator.'
-              end
+              flash[:alert] =
+                if possible_user.present?
+                  I18n.t('devise.failure.inactive')
+                else
+                  'The UNI ' + user_uni + ' is not authorized to log into Hyacinth (no account exists with email ' + user_uni + '@columbia.edu).  If you believe that you should have access, please contact an application administrator.'
+                end
```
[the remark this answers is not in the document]
barmintor: I pulled this commit out into a separate PR - I'll remove it from this PR before requesting re-review.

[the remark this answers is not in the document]
elohanlon: Sounds good!  Thanks!

elohanlon:
I was testing this locally and noticed an issue: In the app sample data, with the Sample Language field, you get zero facet filter hits when searching for `en-CA` in both `starting with` and `containing` scenarios, even though there are several facets that start with `en-CA` (i.e. en-CA, en-CA-newfound, en-CA-newfound-unifon, and en-CA-unifon).

lib/solr/params.rb
```diff
@@ -74,6 +73,7 @@ def sort(field, direction)
       raise ArgumentError, "direction must be one of #{VALID_SORT_DIRECTION.join(', ')}, instead got '#{direction}'" unless VALID_SORT_DIRECTION.include?(direction)
 
       @parameters[:sort] = "#{field} #{direction}"
+      self
```
elohanlon, line 76: Although this comment is out of scope for this PR, it may also be good to return `self` in the above `raw_parameter` method.

lib/solr/params.rb
```diff
@@ -93,15 +93,32 @@ def with_statistics!
...
+        self
       end
 
       # TODO: Enable direction after Solr 8 upgrade
       def sort(field, _direction)
         @params.raw_parameter(:"f.#{@facet_name}.facet.sort", field)
+        self
+      end
+
+      # TODO: Build case and match function options around regex match after Solr 8
```
elohanlon, line 110: Thank you for adding notes like this about Solr 8.  👍

elohanlon:
I was testing some more in my local copy and I also found that for each letter you type in the facet starts-with/contains input, it results in two query requests -- one for the previous text input value (before key press) and one for the updated value (after key press).

app/graphql/types/query_type.rb
```diff
@@ -135,14 +136,14 @@ def digital_object(id:)
...
         solr_params.facet_on(arguments[:field_name]) do |facet_params|
-          facet_params.rows(arguments[:limit])
-          facet_params.start(arguments[:offset])
+          facet_params.rows(arguments[:limit].to_i + 1).start(arguments[:offset])
           facet_params.sort(arguments[:order_by][:field], arguments[:order_by][:direction]) if arguments[:order_by]
-          facet_params.with_statistics!
+          facet_params.filter(arguments[:facet_filter][:match_type], arguments[:facet_filter][:value]) if arguments[:facet_filter]
         end
       end
     end
```
elohanlon, line 149: With the updated functionality in this method, it would be good to update `facet_values_spec.rb` to test `facet_filter`/`match_type`?

elohanlon, line 149: With the updated functionality in this method, it would be good to update `spec/requests/graphql/query/facet_values_spec.rb` to test `facet_filter`/`match_type`.

barmintor: Definitely needed some test coverage - I added jest specs and unit tests for Solr::Params to cover param construction.

lib/solr/params.rb
barmintor: Good call - added it in a few places to make sure chaining works.

barmintor:
> I was testing some more in my local copy and I also found that for each letter you type in the facet starts-with/contains input, it results in two query requests -- one for the previous text input value (before key press) and one for the updated value (after key press).

Hmm. The filter setting method is called from an onChange handler, which should fire _less_ frequently than onInput. Setting a `defaultValue` counts as a change (I think) but it's not used in the Controls component. Is it used in `Form.Control`?

barmintor:
Noting these related issues, since we're on an old version of react-bootstrap:
https://github.com/ericgio/react-bootstrap-typeahead/issues/305
https://github.com/ericgio/react-bootstrap-typeahead/issues/320

barmintor:
There's no state on the Controls component, so I'm trying to think of ways this might be a double-render situation when the SlideOut state changes.

lib/solr/params.rb
elohanlon: Thanks!

elohanlon (approved):
Approved!  Thank you for all of your work on this, @barmintor !

PR #405, opened 2021-11-23
use I18n for flash messages in Hyacinth::Users::CasAuthenticationBehavior

barmintor:
Pulling this style conformance/message template work out of the HYACINTH-907 branch

### DIFF
--- a/app/controllers/concerns/hyacinth/users/cas_authentication_behavior.rb
+++ b/app/controllers/concerns/hyacinth/users/cas_authentication_behavior.rb
@@ -61,7 +61,7 @@ module Hyacinth
 
               redirect_to root_path, status: 302
             else
-              flash[:alert] = possible_user.present? ? I18n.t('devise.failure.inactive') : I18n.t('devise.failure.unauthorized', uid: user_uni, email: user_email)
+              flash[:alert] = possible_user.present? ? I18n.t('devise.failure.inactive') : I18n.t('devise.failure.unauthorized', uni: user_uni, email: user_email)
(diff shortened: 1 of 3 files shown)


app/controllers/concerns/hyacinth/users/cas_authentication_behavior.rb
```diff
@@ -51,26 +51,22 @@ def do_cas_login
...
 
               redirect_to root_path, status: 302
             else
-              if possible_user.present? && !possible_user.is_active
-                flash[:alert] = I18n.t 'devise.failure.inactive'
-              else
-                flash[:alert] = 'The UNI ' + user_uni + ' is not authorized to log into Hyacinth (no account exists with email ' + user_uni + '@columbia.edu).  If you believe that you should have access, please contact an application administrator.'
-              end
-
+              flash[:alert] = possible_user.present? ? I18n.t('devise.failure.inactive') : I18n.t('devise.failure.unauthorized', uid: user_uni, email: user_email)
```
elohanlon, line 64:
```suggestion
              flash[:alert] = possible_user.present? ? I18n.t('devise.failure.inactive') : I18n.t('devise.failure.unauthorized', uni: user_uni, email: user_email)
```
The User model has a uid field, but it doesn't hold the same value as the user's uni, so it might be best to call this key `uni` (and update the associated devise.en.yml entry).

elohanlon (approved):
Approved!  Just had a minor suggestion in `cas_authentication_behavior.rb`.

PR #409, opened 2021-12-09
Hyacinth 822 csv create object w uid

JackBlackLight:


spec/models/concerns/digital_object_concerns/validations_spec.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe DigitalObjectConcerns::Validations, type: :model do
```
elohanlon, line 5:
```suggestion
RSpec.describe DigitalObjectConcerns::Validations, type: :model do
  include_context 'with stubbed search adapters'
```
To allow this test to run without solr.

spec/models/concerns/digital_object_concerns/validations_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    it "rejects an invalid uid" do
+      expect { item_invalid_uid }.to raise_error ActiveRecord::RecordInvalid, 'Validation failed: Uid is invalid'
+    end
+    it "rejects a valid uid with uppercase letters" do
+      expect { item_uid_caps }.to raise_error ActiveRecord::RecordInvalid, 'Validation failed: Uid is invalid'
+    end
+    it "accepts a valid uid" do
+      expect(item_valid).to be_valid
+    end
+  end
```
elohanlon, line 20: Proposing a slightly different version of these tests:
```suggestion
  let(:valid_uid) { '2f4e2917-26f5-4d8f-968c-a4015b10e50f' }
  let(:invalid_uid) { 'FFF' }

  context "when a uid is provided on digital object create" do
    it "rejects an invalid uid" do
      expect { FactoryBot.create(:item, uid: invalid_uid) }.to raise_error ActiveRecord::RecordInvalid, 'Validation failed: Uid is invalid'
    end
    it "rejects a capitalized version of a valid uid" do
      expect { FactoryBot.create(:item, uid: valid_uid.upcase) }.to raise_error ActiveRecord::RecordInvalid, 'Validation failed: Uid is invalid'
    end
    it "accepts a valid uid" do
      FactoryBot.create(:item, uid: valid_uid)
    end
  end
```
Feel free to take what you think is useful in here (or not!).  Here's my thinking behind the proposal:
- This is personal preference, but when a creation operation is the thing that result in an error, I like to see the create call in the test rather than having to look at an earlier `let` to see what operation is raising an error.
- I thought that it might be illustrative to show the `upcase` call on the same `valid_uid` that passes in another test.

app/jobs/digital_object_import_processing_job.rb
```diff
@@ -44,13 +44,18 @@ def self.perform(digital_object_import_id)
...
-    return Hyacinth::Config.digital_object_types.key_to_class(digital_object_data['digital_object_type']).new if digital_object_data['digital_object_type'].present?
-
-    raise ArgumentError, 'Unable to find or create digital object because neither "uid" nor "digital_object_type" attributes were present in digital_object_data.'
+    if digital_object_data['assign_uid'].present?
+      raise ArgumentError, 'Either "uid" or "assign_uid" attribute may be present in digital_object_data, but not both.' if digital_object_data['uid'].present?
+      raise ArgumentError, 'Cannot create new digital object with "assign_uid"; object with same uid already exists' if DigitalObject.find_by_uid(digital_object_data['assign_uid'])
+      raise ArgumentError, 'Cannot create new digital object with "assign_uid"; digital object type is required' unless digital_object_data['digital_object_type'].present?
+      # assign_uid present, so we'll create a new object with type based on the digital_object_type value
+      Hyacinth::Config.digital_object_types.key_to_class(digital_object_data['digital_object_type']).new
```
elohanlon, line 52: If this line is inside of the `if digital_object_data['assign_uid'].present?` block, does that prevent creation of a new digital object in the case when assign_uid isn't present?  (i.e. the usual case of creating a digital object and letting the system auto-assign a new uid)

app/jobs/digital_object_import_processing_job.rb
```diff
@@ -44,13 +44,18 @@ def self.perform(digital_object_import_id)
...
-    return DigitalObject.find_by_uid!(digital_object_data['uid']) if digital_object_data['uid'].present?
-
-    # No uid present, so we'll create a new object with type based on the digital_object_type value
-    return Hyacinth::Config.digital_object_types.key_to_class(digital_object_data['digital_object_type']).new if digital_object_data['digital_object_type'].present?
-
-    raise ArgumentError, 'Unable to find or create digital object because neither "uid" nor "digital_object_type" attributes were present in digital_object_data.'
+    if digital_object_data['assign_uid'].present?
+      raise ArgumentError, 'Either "uid" or "assign_uid" attribute may be present in digital_object_data, but not both.' if digital_object_data['uid'].present?
+      raise ArgumentError, 'Cannot create new digital object with "assign_uid"; object with same uid already exists' if DigitalObject.find_by_uid(digital_object_data['assign_uid'])
+      raise ArgumentError, 'Cannot create new digital object with "assign_uid"; digital object type is required' unless digital_object_data['digital_object_type'].present?
```
elohanlon, line 50: The `digital_object_type` field is also required when creating a new digital object without a uid, so I don't think we want this check scoped to the `if digital_object_data['assign_uid'].present?` block.

app/jobs/digital_object_import_processing_job.rb
```diff
@@ -44,13 +44,18 @@ def self.perform(digital_object_import_id)
...
+      # assign_uid present, so we'll create a new object with type based on the digital_object_type value
+      Hyacinth::Config.digital_object_types.key_to_class(digital_object_data['digital_object_type']).new
+    elsif digital_object_data['uid'].present?
+      # If a uid is present in the digital_object_data, then this is an update operation
+      DigitalObject.find_by_uid!(digital_object_data['uid'])
+    else
+      raise ArgumentError, 'Unable to find or create digital object because "uid" or “assign_uid”/"digital_object_type" attributes were not present in digital_object_data.'
```
elohanlon, line 57: Related to the above comments, we still want to allow creation of a new digital object when neither uid nor assign_uid are present (but digital_object_type still needs to be present in that case).

spec/factories/digital_object_imports.rb
```diff
@@ -5,6 +5,7 @@
     index { 34 }
     digital_object_data do
       {
+        'assign_uid' => '2f4e2917-26f5-4d8f-968c-a4015b10e50f',
```
elohanlon, line 8: Rather than adding assign_uid to the base factory definition, it would be better to create a new trait like `:item_with_assign_uid` since lots of tests use this factory or the traits that inherit from it, and with this change those tests are no longer covering the default (and most common) auto-assign-new-uid case.

elohanlon, line 8: Rather than adding assign_uid to the base factory definition, it would be better to create a new trait like `:item_with_assign_uid` so that assign_uid doesn't appear as the default case (since, upcoming migration scenario aside, it will be infrequent).  I think this will also make it less necessary to modify existing text fixture files (e.g. sample-batch-import.csv).

spec/factories/digital_object_imports.rb
```diff
@@ -25,6 +26,7 @@
     trait(:asset) do
       digital_object_data do
         {
+          'assign_uid' => '2f4e2917-26f5-4d8f-968c-a4015b10e50f',
```
elohanlon, line 29: Along with my suggestion about creating a new, separate trait for the `assign_uid` case, no need to add `assign_uid` here on the asset.  Testing it on one digital object type should be enough, and I'd feel better keeping the asset digital_object_import uid-less so that we can continue to test that the default auto-uid-assignment scenario works.

spec/jobs/digital_object_import_processing_job_spec.rb
```diff
@@ -150,21 +156,53 @@
...
       expect(
         described_class.digital_object_for_digital_object_data(existing_digital_object_data).new_record?
       ).to eq(false)
     end
 
-    it 'instantiates a new digital object of the expected type when a uid is NOT provided in the digital_object_data' do
-      obj = described_class.digital_object_for_digital_object_data(new_digital_object_data)
-      expect(obj).to be_a(DigitalObject::Item)
-      expect(obj.new_record?).to eq(true)
+    it 'raises an error if assign_uid or uid are not provided in the digital_object_data' do
```
elohanlon, line 179: Related to my earlier comment, creating a new import with no `assign_uid` or `uid` actually is a valid scenario (for automatic uid generation by the system).

spec/jobs/digital_object_import_processing_job_spec.rb
```diff
@@ -150,21 +156,53 @@
...
+      }
+    end
+
     it 'finds an existing digital object when a uid is provided in the digital_object_data' do
       expect(
         described_class.digital_object_for_digital_object_data(existing_digital_object_data).new_record?
       ).to eq(false)
     end
 
-    it 'instantiates a new digital object of the expected type when a uid is NOT provided in the digital_object_data' do
```
elohanlon, line 159: This is an important test that we don't want to remove.  The best way to think about this is:
- Creating a digital object without a uid is the default scenario.
- The `assign_uid` field can be optionally added to a csv row/digital_object_import, and that just allows a user to provide the specific uid that they want to be assigned to the brand new digital object before it's saved for the first time.

spec/jobs/digital_object_import_processing_job_spec.rb
```diff
@@ -150,21 +156,53 @@
...
+      }.to raise_error(ArgumentError, "Unable to find or create digital object because \"uid\" or \“assign_uid\”/\"digital_object_type\" attributes were not present in digital_object_data.")
+    end
+
+    it 'raises an error if value for assign_uid is already assigned to a digital object' do
+      expect {
+        described_class.digital_object_for_digital_object_data(digital_object_data_attempt_to_create_existing_object)
+      }.to raise_error(ArgumentError, "Cannot create new digital object with \"assign_uid\"; object with same uid already exists")
     end
 
-    it 'raises an error if neither uid nor digital_object_type are keys in the digital_object_data' do
```
elohanlon, line 165: We want to keep this test too.

JackBlackLight:
Just using the official channel to let you know this is ready for rereview @elohanlon (even though you already know).

app/jobs/digital_object_import_processing_job.rb
```diff
@@ -44,13 +44,23 @@ def self.perform(digital_object_import_id)
...
-    return Hyacinth::Config.digital_object_types.key_to_class(digital_object_data['digital_object_type']).new if digital_object_data['digital_object_type'].present?
+    if digital_object_data['uid'].present? && digital_object_data['assign_uid'].present?
+      raise ArgumentError, 'Either "uid" or "assign_uid" attribute may be present in digital_object_data, but not both.'
+    end
+    if digital_object_data['uid'].present?
+      # If a uid is present in the digital_object_data, then this is an update operation
+      DigitalObject.find_by_uid!(digital_object_data['uid'])
+    elsif appropriate_create_data_present(digital_object_data)
+      Hyacinth::Config.digital_object_types.key_to_class(digital_object_data['digital_object_type']).new
+    end
```
elohanlon, line 55:
```suggestion
if digital_object_data['uid'].present? && digital_object_data['assign_uid'].present?
      raise ArgumentError, 'Either "uid" or "assign_uid" attribute may be present in digital_object_data, but not both.'
    end
    return DigitalObject.find_by_uid!(digital_object_data['uid']) if digital_object_data['uid'].present?

    ensure_appropriate_create_data_present!(digital_object_data)
    Hyacinth::Config.digital_object_types.key_to_class(digital_object_data['digital_object_type']).new
```

app/jobs/digital_object_import_processing_job.rb
```diff
@@ -44,13 +44,23 @@ def self.perform(digital_object_import_id)
...
+  end
 
-    raise ArgumentError, 'Unable to find or create digital object because neither "uid" nor "digital_object_type" attributes were present in digital_object_data.'
+  def self.appropriate_create_data_present(digital_object_data)
+    raise ArgumentError, 'Cannot create new digital object; digital object type is required' unless digital_object_data['digital_object_type'].present?
+    if digital_object_data['assign_uid'].present?
+      raise ArgumentError, 'Cannot create new digital object with "assign_uid"; object with same uid already exists' if DigitalObject.find_by_uid(digital_object_data['assign_uid'])
+    end
+    true
```
elohanlon, line 63:
```suggestion
  def self.ensure_appropriate_create_data_present!(digital_object_data)
    raise ArgumentError, 'Cannot create new digital object; digital object type is required' unless digital_object_data['digital_object_type'].present?
    if digital_object_data['assign_uid'].present? && DigitalObject.find_by_uid(digital_object_data['assign_uid'])
      raise ArgumentError, 'Cannot create new digital object with "assign_uid"; object with same uid already exists'
    end
    true
```

app/jobs/digital_object_import_processing_job.rb
elohanlon: Proposing a minor refactor (with another related comment below) because I thought it wasn't immediately obvious that the elsif condition check might raise an error.  The `ensure_appropriate_create_data_present!` method, ending with a `!` could be a good way to indicate this at a glance.

spec/fixtures/files/batch_import/sample-batch-import.csv
```diff
@@ -1,6 +1,6 @@
...
-,asset1,asset,2f4e2917-26f5-4d8f-968c-a4015b10e50f,,,great_project,New Asset 1,fixture_file,test.txt
-,asset2,asset,,item1,,great_project,New Asset 2,fixture_file,test.txt
-,asset3,asset,,item1,item2,great_project,New Asset 3,fixture_file,test.txt
-,item2,item,,,,great_project,New Item 2,,
+_assign_uid,_uid,_identifiers[0],_digital_object_type,_parents[0].uid,_parents[1].identifier,_parents[2].identifier,_primary_project.string_key,_title.value.sort_portion,_resource_imports.main.method,_resource_imports.main.location
+,2f4e2917-26f5-4d8f-968c-a4015b10e50f,item1,item,,,,great_project,Existing Item 1,,
+3f4e2917-26f5-4d8f-968c-a4015b10e50f,,asset1,asset,2f4e2917-26f5-4d8f-968c-a4015b10e50f,,,great_project,New Asset 1,fixture_file,test.txt
+4f4e2917-26f5-4d8f-968c-a4015b10e50f,,asset2,asset,,item1,,great_project,New Asset 2,fixture_file,test.txt
+5f4e2917-26f5-4d8f-968c-a4015b10e50f,,asset3,asset,,item1,item2,great_project,New Asset 3,fixture_file,test.txt
+6f4e2917-26f5-4d8f-968c-a4015b10e50f,,item2,item,,,,great_project,New Item 2,,
```
elohanlon, line 6: Did this change have any testing side effects?

JackBlackLight: i ran hyacinth:ci and all tests passed, so apparently not.

elohanlon: Also, I'd recommend not changing all of the assets here to have an _assigned_uid, just to make sure we have some examples of asset creation without assigning an id.  Maybe just add _assigned_uid it to one of them and leave blank for the rest?

elohanlon: After our conversation about those future csv pre-validation TODOs, I'll revise my earlier suggestion: It would still be good to have _assign_uid for only one of those rows, but it's fine not to worry about test side effects or needing to add a new test to cover the addition.  When we add csv validation improvements down the line, this will be a reminder to make sure that _assign_uid is handled appropriately during pre-validation.

Note about pre-validation: It's only meant to highlight user errors earlier in the csv processing process, but the real protections are on the existing Digital Object validations.

elohanlon (changes requested):
Thought I selected "Request changes" before, but actually doing it now!

spec/fixtures/files/batch_import/sample-batch-import.csv
JackBlackLight: I removed all but one assign_uid value since they were now unnecessary.

elohanlon (approved):
PR approved!  Thanks, @JackBlackLight !

PR #410, opened 2021-12-14
Publish UI and GraphQL (HYACINTH-623 and HYACINTH-932); Preserve UI and GraphQL (HYACINTH-940 and HYACINTH-941)

elohanlon:


spec/hyacinth/adapters/digital_object_search_adapter/solr_spec.rb
```diff
@@ -60,12 +60,10 @@
 
   context "#solr_document_for" do
     let(:authorized_object) { FactoryBot.build(:digital_object_test_subclass, :with_sample_data) }
-    let(:publication_adapter) { Hyacinth::Adapters::PublicationAdapter::Abstract.new }
     let!(:external_identifier_adapter) { Hyacinth::Adapters::ExternalIdentifierAdapter::Memory.new }
 
     before do
       authorized_object.send :uid=, 'dummy-uid'
-      allow(Hyacinth::Config).to receive(:publication_adapter).and_return(publication_adapter)
       allow(Hyacinth::Config).to receive(:external_identifier_adapter).and_return(external_identifier_adapter)
```
JackBlackLight, line 67: Hi, just wanted the background about why the publish_adapter is being replaced by the external_identifier_adapter here. Thanks.

app/models/concerns/digital_object_concerns/publish_behavior.rb
```diff
@@ -81,7 +88,10 @@ def perform_publish_changes(publish_to: [], unpublish_from: [], publishing_user:
             # Remove entries associated with successfull unpublishes AND successful publishes.
             # For successful publishes that need time/user updates, it's easier to do a mass deletion
             # than to find and update specific ones.
-            self.publish_entries.destroy_by(publish_target: successful_publish_targets + successful_unpublish_targets)
+            successful_additons_and_removals = successful_publish_targets + successful_unpublish_targets
```
JackBlackLight, line 91: Just flagging a typo here (and in the reference below it): should be successful_additions_and_removals

JackBlackLight (approved):
This looks good @elohanlon. The publish/unpublish/skip interface works well; there's a lot going on there, but it's pretty intuitive. 

I do wonder if it wouldn't be better to not to show the preserve/publish tab at all if the user doesn’t have permission, but maybe that's more trouble than it is worth (and could certainly be a separate ticket). 

In my  comments, I ask a general info question and caught a typo - nothing serious, so am approving now.

spec/hyacinth/adapters/digital_object_search_adapter/solr_spec.rb
elohanlon: Not a replacement in this case, just a deletion of the upper of the two lines.  With the publication adapter config change I made for the test environment, we no longer need to fake the `Hyacinth::Config.publication_adapter` method call here, so that's why I took the line out.

app/models/concerns/digital_object_concerns/publish_behavior.rb
elohanlon: Ah!  Thanks for catching that!  Fixed!

elohanlon:
@JackBlackLight Thank you for the review!  Question answered and typo fixed!

That's a good thought about whether or not we should show the preserve/publish tab, so that people without the right permission don't click on it only to see a "You do not have permission" message.  The `DigitalObjectInterface` component doesn't currently consider permissions when rendering those tabs, so right now it's showing up just because of a simple implementation, but it's worth reconsidering in the future. 👍

PR #411, opened 2021-12-23
HYACINTH-942: Switch Hyacinth 2.x to DataCite REST API

fcd1:


lib/hyacinth/datacite/api_session.rb
```diff
@@ -0,0 +1,117 @@
+# fcd1, 08/30/16: Original code was copied verbatim from
+# hypatia-new, and then modified. Modifications should be
+# kept to a minimum
+
+# This code interfaces with the EZID server
+# and therefore implements parts of the EZID server
+# API (http://ezid.cdlib.org/doc/apidoc.2.html)
+# EZID API, Version 2
```
elohanlon, line 8: Outdated comments?  Okay to remove?

lib/hyacinth/datacite/api_session.rb
```diff
@@ -0,0 +1,117 @@
...
+
+    private
+
+      def call_api(uri, request, attributes = nil)
+        data = { type: 'dois' }
+        data[:attributes] = attributes
+        request.body = { data: data }.to_json
+        # fcd1, 12/23/21: Following can be uncommented for in situ debugging
+        # Hyacinth::Utils::Logger.logger.error("URL: #{uri}")
+        # Hyacinth::Utils::Logger.logger.error("REQUEST BODY #{request.body}")
```
elohanlon, line 103: Lines like this make sense during active development, but I'd prefer to remove them once the code is production ready.

spec/models/concerns/digital_object/datacite_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+    test_do
+  end
+
+  context "#mint_and_store_doi:" do
+    it "mint_and_store_doi" do
+      # stub out method wrapping API call to EZID server, stub return Doi instance
+      allow_any_instance_of(Hyacinth::Datacite::ApiSession).to receive(:mint_identifier) do
+        Hyacinth::Datacite::Doi.new('10.33555/5x55-t644')
+      end
+      EZID[:user] = EZID[:ezid_test_user]
```
elohanlon, line 29: You should use `stub_const` instead of modifying the constant because I believe direct changes to this constant will leak into other tests.

https://relishapp.com/rspec/rspec-mocks/docs/mutating-constants

elohanlon (approved):
Thanks for backporting this feature!  I had a couple of minor comments, but since the tests pass and we have some time pressure to get this deployed I'll approve now.  If everything goes well after deployment, we can address those comments after the holiday break and then merge this PR into 2.x.

spec/unit/hyacinth/datacite/api_session_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+      allow_any_instance_of(Hyacinth::Datacite::ApiSession).to receive(:call_api).with(instance_of(URI::HTTPS),
+                                                                                       instance_of(Net::HTTP::Put),
+                                                                                       expected_attributes_parameter ).and_return response
+    end
+    it "modify identifier" do
+      subject.modify_identifier('10.33555/5x55-t644',
+                                :findable,
+                                sample_update_attributes,
+                                'https://www.columbia.edu')
+    end
```
elohanlon, line 85: Since this method can return a `Hyacinth::Datacite::Doi` or `nil` or `false`, it would be good to ensure that those three scenarios are tested.

spec/unit/hyacinth/datacite/api_session_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+  context "mint_identifier" do
+    let(:response) { instance_double(Net::HTTPResponse, body: sample_response_body_json) }
+    before do
+      allow_any_instance_of(Hyacinth::Datacite::ApiSession).to receive(:call_api).with(instance_of(URI::HTTPS),
+                                                                                       instance_of(Net::HTTP::Post),
+                                                                                       {prefix: "10.33555"} ).and_return response
+    end
+    it "mint identifier" do
+      subject.mint_identifier('10.33555',:draft)
+    end
```
elohanlon, line 70: Same comment as the other test in this file: since this method can return a `Hyacinth::Datacite::Doi` or `nil` or `false`, it would be good to ensure that those three scenarios are tested.

elohanlon (changes requested):
@fcd1 Now that the December 31st time pressure is off, would you be up for making some of the above requested changes before we merge this in?  I had a chance to take a closer look at a couple of methods and added two notes about test coverage.

fcd1:
@elohanlon We need to merge this into 2.x. We can open a new ticket for the minor outstanding change requests (remove comments, update tests) -- I'm currently concentrating on addressing errors appearing in the log file. Looking at /opt/passenger/ldpd/hyacinth_prod/current, it appears that 2.x was deployed on Feb 3, and 2.x does not have the DataCite REST API updates.

spec/models/concerns/digital_object/datacite_spec.rb
fcd1: Since the underlying code reads directly from the EZID hash, stub_const won't work without changes to the underlying code.
Furthermore, the original code, spec/models/concerns/digital_object/ezid_spec.rb, modifies the constant in the same way. Since one of the requirements of this implementation was to change as little code as possible, I suggest we leave it as it is.

spec/unit/hyacinth/datacite/api_session_spec.rb
fcd1: I would like to refactor the Doi class out of the implementation since it doesn't really serve a purpose. At that time, I can update the above spec.

spec/unit/hyacinth/datacite/api_session_spec.rb
fcd1: I would like to refactor the Doi class out of the implementation since it doesn't really serve a purpose. At that time, I can update the above spec

lib/hyacinth/datacite/api_session.rb
fcd1: Comments have been removed.

lib/hyacinth/datacite/api_session.rb
fcd1: Comments have been removed.

elohanlon:
@fcd1 Thanks for addressing some of the remaining changes today.  You're right that the rest can be addressed later.  I'll merge the PR now so we don't run into the deployment issue again.

spec/models/concerns/digital_object/datacite_spec.rb
elohanlon: @fcd1 If you want to use the "real" EZID hash as the source data for stub_const (and then merge in any test-specific changes) then I think it should still work.

I understand that you're trying to change as little source implementation code as possible.  In the above comment, I wasn't asking you to change the underlying code though, just the test.  Even if you'd prefer not to make the change, I hope my comment intention was clear/helpful?

spec/unit/hyacinth/datacite/api_session_spec.rb
elohanlon: Sounds good.  I'll ticket that.

spec/unit/hyacinth/datacite/api_session_spec.rb
elohanlon: Like the above comment, I'll ticket that too.

PR #413, opened 2022-01-19
hyacinth-938 controlled term error bug fix

JackBlackLight:


elohanlon:
@fcd1 Just following up on this -- did you have a chance to verify the fix in Jack's deployed copy?

fcd1:
@elohanlon The issue originally noted in the ticket description is fixed. However, I noticed another issue which I documented via a comment in the ticket.

elohanlon:
@fcd1 Thanks for documenting that!  In that case, I'll merge this PR and we can follow up with the other fix in a separate ticket, which will reference what you noted in HYACINTH-938.

elohanlon (approved):

PR #417, opened 2022-02-25
HYACINTH-956 - Update to Rails 7

elohanlon:
- Also updated react-bootstrap npm module from beta version to later stable release


app/models/ability.rb
```diff
@@ -93,7 +93,7 @@ def calculate_permissions(user)
 
   def to_list
     # Skipping rules that only contain a block.
-    rules.select { |r| !r.only_block? }.map do |rule|
+    rules.reject(&:only_block?).map do |rule|
       rule.only_block?
```
JackBlackLight, line 97: could you explain what this line (97) is for? thanks

JackBlackLight (approved):
@elohanlon, i just had a few questions for my own understanding. Otherwise, everything looks good.

app/controllers/concerns/hyacinth/users/cas_authentication_behavior.rb
```diff
@@ -20,7 +20,7 @@ def do_cas_login
           # If ticket is NOT set, this means that the user hasn't gotten to the uni/password login page yet.  Let's send them there.
           # After they log in, they'll be redirected to this page and they'll continue with the authentication.
 
-          redirect_to(cas_login_uri + '?service=' + URI.encode_www_form_component(request.protocol + request.host_with_port + '/users/do_cas_login'))
+          redirect_to(cas_login_uri + '?service=' + URI.encode_www_form_component(request.protocol + request.host_with_port + '/users/do_cas_login'), allow_other_host: true)
```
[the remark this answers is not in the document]
elohanlon: After the Rails 7 update, I got an error that said `redirect_to` with an external domain (in this case, the CAS domain) required an explicit `allow_other_host` flag.  It seems to be a security measure that makes you opt into allowing your app to redirect to another domain.

app/graphql/mutations/update_project_permissions.rb
```diff
@@ -46,8 +46,7 @@ def project_permissions_response(project_permissions_update, users, projects)
 
   def apply_new_permission_actions(project, user, permission_actions)
     # Perform update by clearing out all old permissions for this user+project combo
-    # TODO: In Rails 6, change operation below Permission.delete_by(...)
-    Permission.where(user: user, subject: 'Project', subject_id: project.id).delete_all
+    Permission.delete_by(user: user, subject: 'Project', subject_id: project.id)
```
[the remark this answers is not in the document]
elohanlon: SO GOOD!  Haha.

app/models/ability.rb
elohanlon: That is an excellent question.  That line appears to do...absolutely nothing!  Haha

I looked back at the commit that originally added it (from a couple of years ago) and it looks like it was accidentally left in.  I'll remove it and verify that the tests still pass.

Thank you for catching that!

elohanlon: Tests passed!

elohanlon:
@JackBlackLight Thank you for reviewing and approving!

PR #422, opened 2022-06-27
Webpack 5 and shakapacker

elohanlon:
We can move forward with merging this because we're working with LIT to get deployment set up on the new AlmaLinux hosts, which are a requirement for running the newer version of node.

I'm planning on looking into pack chunking later on, but this is just the initial, working migration from webpacker 4 to shakapacker 6.


elohanlon:
Thanks, @JackBlackLight !

PR #423, opened 2022-07-08
Docker for external development services and CI

elohanlon:
Keeping this as a draft for now.  I temporarily have the PR targeting merge into the `webpack-5-and-shakapacker` branch, but I'll eventually change it to the `main` branch after that webpack 5 branch is merged.  I'm targeting `webpack-5-and-shakapacker` to provide a more useful diff view here in the PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,13 +138,10 @@ If you have an IDE that supports jsconfig.json files (e.g. Visual Studio Code), 
 {
   "compilerOptions": {
     "baseUrl": "app/javascript",
-    "module": "commonjs",
-    "target": "es2017",
     "jsx": "react",
-    "checkJs": true,
   },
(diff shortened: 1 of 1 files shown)


elohanlon:
@barmintor @JackBlackLight @fcd1 I've added all three of you as reviewers here so you can see what was required for the docker setup, in case we want to do this for any other apps in the near future.

elohanlon:
Since this these changes are fairly isolated from the core app functionality, and they only affect development and testing, and because I'd like to build off them for upcoming work, and because we've also had success testing them in @JackBlackLight 's development environment, I'm going to go ahead and merge -- but feel free to add any comments at your convenience and I'll be glad to answer/address/improve later!

PR #429, opened 2023-03-07
Switch non-synchronizer media player to VideoJS

elohanlon:


elohanlon:
Note: CI not working because of GitHub intentional Ubuntu 18 brownout, but all tests pass in local CI runs.

PR #435, opened 2023-03-27
DLC-985 poll access derivative generation for republish

barmintor:


elohanlon:
Looks good!  Similar to the other PR, would you mind rebasing off 2.x to make sure that everything passes here on Ruby 2.6, and in the later version of Ubuntu?  Also like the other PR, I don't expect any issues, but would be nice to verify before the merge.

elohanlon (approved):

elohanlon:
Thanks!

PR #436, opened 2023-04-02
2.x allow controlled vocabularies to be configured not to allow temp term creation

barmintor:


elohanlon:
@barmintor Great changes, thank you!

One question and one request:

Question: What kind of feedback does a user get during UI and CSV import scenarios, if they attempt to create a temp term in a vocabulary where it's not allowed?  (I forgot what happens when an error is raised at the level where you're raising the error.)  Do they get enough information to know what they need to fix?

Request: Could you please rebase off 2.x to make sure that everything passes here on Ruby 2.6, and in the later version of Ubuntu?  I don't expect any issues, but would be nice to verify before the merge.

db/migrate/20230327200000_add_prohibit_temp_terms_to_controlled_vocabularies.rb
```diff
@@ -0,0 +1,7 @@
+class AddProhibitTempTermsToControlledVocabularies < ActiveRecord::Migration
+  def change
+    change_table(:controlled_vocabularies) do |t|
+      t.boolean :prohibit_temp_terms, default: false, null: false, index: true
```
elohanlon, line 4:
```suggestion
      t.boolean :prohibit_temp_terms, default: false, null: false
```
Maybe no need for an index on this field?

elohanlon (approved):
Looks good!  I left one optional suggestion.

PR #456, opened 2025-05-05
Initial thoughts as pseudo-code

fcd1:


app/models/concerns/digital_object/persistence.rb
```diff
@@ -42,6 +42,9 @@ def save(manual_commit_to_solr = true)
   def before_save
     # TODO: rewrite with ActiveRecord::Callbacks
     # To be overridden by subclasses
+    # PSEUDO CODE
+    # begin - rescue around the following call, rescue Hyacinth::Exceptions::DataciteError
+    # Question: What to do if above exception is recued/caught?
     mint_and_store_doi(Hyacinth::Datacite::Doi::IDENTIFIER_STATUS[:draft]) if @mint_reserved_doi_before_save || @publish_after_save
```
elohanlon, line 48: Let's plan to add a `rescue` to this method that will add to the @errors object like this:

```
def before_safe
  ...
  ...
rescue
  @errors.add(:datacite, "This is a helpful error message with information that will be shown to the user.")
end
```

elohanlon:
Adding this from our conversation:

In `def change_doi_status_to_unavailable` method:

We agreed that `return false if @doi.nil?` should be changed to an exception instead of returning false.  The exception message could be something like `Hyacinth::Exceptions::MissingDoi` with a message like "Cannot change status of doi that is not present."

app/models/digital_object/publish_target.rb
```diff
@@ -132,6 +132,9 @@ def unpublish_digital_object(digital_object, do_ezid_update)
     )
     if do_ezid_update && !digital_object.doi.blank?
       # We need to change the state of this ezid to :unavailable
+      # PSEUDO CODE
+      # begin - rescue around the following call, rescue Hyacinth::Exceptions::DataciteError
+      # Question: What to do if above exception is recued/caught?
       success = digital_object.change_doi_status_to_unavailable
       digital_object.errors.add(:ezid_response, "An error occurred while attempting to set this digital object's ezid status to 'unavailable'.") unless success
```
elohanlon, line 139: This is pseudocode, but you probably want something like this:

```suggestion
      begin
        digital_object.change_doi_status_to_unavailable
      rescue YourExceptionType1, YourExceptionType2, etc. => e
        digital_object.errors.add(:datacite, e.message)
      end
      status to 'unavailable'.") unless success
```

elohanlon:
Also adding this from our conversation:

Similar to the comment above, we agreed that in `def update_doi_metadata` return false if @doi.nil? should be changed to an exception instead of returning false. The exception message could be something like Hyacinth::Exceptions::MissingDoi with a message like "Cannot update doi metadata because doi is not present."

Also:
If `def update_doi_metadata` is changed so that it raises an exception and no longer returns truthy/false values, you'll need to make a small update in lib/tasks/hyacinth/publish.rake in the `hyacinth:publish:dois` rake task.

fcd1:
In lib/hyacinth/datacite/api_session.rb, when rescuing  Net::ReadTimeout and SocketError, create new Exception (possibly Hyacinth::Exception::DataciteConnectionError) and raise this new exception.

PR #457, opened 2025-05-07
HYACINTH-1087 - Handle errors returned by the DataCite REST API

fcd1:


app/models/concerns/digital_object/datacite.rb
```diff
@@ -4,7 +4,7 @@ module DigitalObject::Datacite
   # or nil if minting did not go through
```
elohanlon, line 4: I think this method description needs to be updated, since this method doesn't return nil anymore?

app/models/concerns/digital_object/datacite.rb
```diff
@@ -48,7 +33,7 @@ def mint_and_store_doi(identifier_status, target_url = nil)
   # returns true if status change was successful
```
elohanlon, line 33: Does this method ever return false anymore?  Or does the comment need to be updated?

app/models/concerns/digital_object/datacite.rb
```diff
@@ -71,7 +57,7 @@ def change_doi_status_to_unavailable
   # Uses the modify identifier API call, returns true if metadata update was successful
   # if not, returns false
```
elohanlon, line 58: Does this method ever return false anymore?  Or does the comment need to be updated?

app/models/concerns/digital_object/persistence.rb
```diff
@@ -43,6 +43,8 @@ def before_save
     # TODO: rewrite with ActiveRecord::Callbacks
     # To be overridden by subclasses
     mint_and_store_doi(Hyacinth::Datacite::Doi::IDENTIFIER_STATUS[:draft]) if @mint_reserved_doi_before_save || @publish_after_save
+  rescue DataciteErrorResponse, DataciteConnectionError, HyacinthError::DoiExists => e
```
elohanlon, line 46: I see you raising `Hyacinth::Exceptions::DataciteErrorResponse` and `Hyacinth::Exceptions::DataciteConnectionError` in one part of the code, but not catching that fully namespaced error classes here.  Is this a mistake?  If the tests are all passing, it seems like you may be missing some tests for this update.

app/models/digital_object/publish_target.rb
```diff
@@ -147,8 +149,10 @@ def publish_digital_object(digital_object, do_ezid_update)
     if do_ezid_update && response.code == 200 && response.headers[:location].present?
       # By this point, all records should have an ezid. Let's update the status of
       # that ezid to :public, and send the latest published_object_url in case it changed.
-      unless digital_object.update_doi_metadata(response.headers[:location])
-        @errors.add(:ezid_response, "An error occurred while attempting to updated the ezid doi for this object.")
+      begin
+        digital_object.update_doi_metadata(response.headers[:location])
+      rescue DataciteErrorResponse, DataciteConnectionError, HyacinthError::DoiMissing => e
```
elohanlon, line 154: I see you raising `Hyacinth::Exceptions::DataciteErrorResponse` and `Hyacinth::Exceptions::DataciteConnectionError` in one part of the code, but not catching those fully namespaced error classes here.  Is this a mistake?  If the tests are all passing, it seems like you may be missing some tests for this update.

app/models/concerns/digital_object/datacite.rb
```diff
@@ -103,15 +90,15 @@ def update_doi_metadata(target_url = nil)
...
```
elohanlon, line 91: Does this method ever return false anymore? Or does the comment need to be updated?

elohanlon (changes requested):
@fcd1 Thank you for these updates.  I think a couple of comments need to be updated to reflect the new method behavior, but that's minor.  However, I think there are some incorrect references to error classes that need to be fixed (which may reveal some gaps in testing), and we'd want to fix those before merging.

elohanlon:
You'll also want to rebase this branch off `2.6.x` to pick up some changes that I pushed, which preserve errors set in the `before_save` method.

app/models/concerns/digital_object/datacite.rb
fcd1: Comment updated as suggested.

app/models/concerns/digital_object/datacite.rb
fcd1: Comment updated as suggested.

app/models/concerns/digital_object/datacite.rb
fcd1: Comment updated as suggested.

app/models/concerns/digital_object/datacite.rb
fcd1: Comment updated as suggested.

lib/tasks/hyacinth/publish.rake
```diff
@@ -134,7 +134,13 @@ namespace :hyacinth do
             puts "must have _pid and _doi_target; skipping : #{row.headers} #{row}"
           else
             obj = DigitalObject::Base.find(pid)
-            result = (ENV['include_metadata'] != 'TRUE') ? obj.update_doi_target_url(target) : obj.update_doi_metadata(target)
+            begin
+              result = (ENV['include_metadata'] != 'TRUE') ? obj.update_doi_target_url(target) : obj.update_doi_metadata(target)
+            rescue Hyacinth::Exceptions::DataciteErrorResponse,
+                   Hyacinth::Exceptions::DataciteConnectionError,
+                   Hyacinth::Exceptions::HyacinthError::MissingDoi => e
```
elohanlon, line 141: @fcd1 This `Hyacinth::Exceptions::HyacinthError::MissingDoi` should probably be `Hyacinth::Exceptions::MissingDoi`.

app/models/digital_object/publish_target.rb
```diff
@@ -131,9 +131,11 @@ def unpublish_digital_object(digital_object, do_ezid_update)
       Authorization: "Token token=#{publish_target_field('api_key')}"
     )
     if do_ezid_update && !digital_object.doi.blank?
-      # We need to change the state of this ezid to :unavailable
-      success = digital_object.change_doi_status_to_unavailable
-      digital_object.errors.add(:ezid_response, "An error occurred while attempting to set this digital object's ezid status to 'unavailable'.") unless success
+      begin
+        digital_object.change_doi_status_to_unavailable
+      rescue DataciteErrorResponse, DataciteConnectionError, HyacinthError::DoiMissing => e
```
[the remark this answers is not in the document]
fcd1: I did not find existing specs for publish_target.rb. Assuming there are no issues, I will create publish_target_spec.rb in spec/models/digital_object.

app/models/concerns/digital_object/persistence.rb
fcd1: I did not find existing specs for persistence.rb. Assuming there are no issues, I will create persistence_spec.rb in spec/models/concerns/digital_object/

elohanlon: @fcd1 I think it might be useful to talk this over in real time and I sent you a message a couple of hours ago in gchat.  Feel free to reply over there any let me know when you're available.  I don't think that what you need to test here is too bad.  Let's work through it together, and talk through some existing test examples.

elohanlon (approved):

PR #459, opened 2025-06-27
add rake tasks to reset all user credentials

barmintor:
add rake tasks to reset all user credentials from command line or assign individually

- UNIX-9927

### DIFF
--- a/lib/tasks/hyacinth/users.rake
+++ b/lib/tasks/hyacinth/users.rake
@@ -1,0 +1,36 @@
+require 'base64'
+require 'io/console'
+
+namespace :hyacinth do
+  namespace :users do
+    # this is ported from the logic in users.js, with a longer default and b64 encoding
+    def random_password(len = 24)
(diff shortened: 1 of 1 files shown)


lib/tasks/hyacinth/users.rake
```diff
@@ -0,0 +1,36 @@
...
+        user.update!(password: random_password)
+      end
+    end
+
+    desc "reset a specific user credential to an input value"
+    task reset_user_creds: :environment do
+      user_id = ENV['id']&.to_i
+      unless user_id
+        puts "call this task with id=INTEGERID"
+        break
```
wtomaszewska, line 26: It might be better to use "abort" or "raise" to terminate this task early. Break causes `LocalJumpError: break from proc-closure (LocalJumpError)` and is semantically most often used in loops.

barmintor: This is a good point. I'm always of two minds in rake tasks because they look like methods but they're blocks. When you're returning early from a block you might use `next`?

PR #465, opened 2025-07-24
Ruby 3 removes URI.escape (HYACINTH-1110)

barmintor:


app/controllers/users_controller.rb
```diff
@@ -6,6 +6,23 @@ class UsersController < ApplicationController
   # Note: To log a specific user in manually, use:
   # sign_in User.where(email: 'durst@library.columbia.edu').first, :bypass => true
 
+
+  # We'll validate the ticket against the wind server
```
barmintor, line 10: @elohanlon I pulled this out into a method for testability purposes.

elohanlon (approved):
Looks great!  Thank you for the fix and tests!

PR #489, opened 2026-04-09
Update to Yarn 4 and Node 22

elohanlon:
This PR updates Hyacinth to use Node 22 and Yarn 4.

The node_modules directory no longer needs to be symlinked during a deployment because we're using the default pnp nodeLinker, which does not create a local node_modules directory.

We are git ignoring additional Yarn-related directories that were introduced in the newer version of Yarn.  These are regenerated on a remote deployment server during deployment.

Our FontAwesome token is now read from a .yarnrc.yml, so the ci.yml GitHub Actions configuration has been updated to construct a .yarnrc.yml file during the CI run.  The `corepack enable` command is also now run during CI to enable corepack's automatic installation and use of the yarn version specified in package.json `packageManager` property.


goldsmithb (approved):

wtomaszewska (approved):
Looks great! Not a blocker but it would be helpful to also update the README.md with instructions on how to set up `yarnrc.yml` (in the "First Time Setup" section).

elohanlon:
@wtomaszewska Good call about the README.md update!  Done!

PR #490, opened 2026-04-17
Switch from FontAwesome Pro to FontAwesome Free

elohanlon:
Creating this PR as a proof of concept, to show what changes would be required to switch from FontAwesome Pro to FontAwesome Free.  Not sure whether or not we'll end up merging it in, but putting it out there as a possibility that we can talk over.


wtomaszewska (approved):
🎉

PR #492, opened 2026-06-30
HYACINTH-1190: Import Jobs and DOIs - New API and UI

wtomaszewska:
# Ticket [HYACINTH-1190](https://columbiauniversitylibraries.atlassian.net/browse/HYACINTH-1190)

## Overview
This PR adds a new API and UI for Import Jobs and their underlying Digital Object Imports. As with previous tickets in this migration, the old API and UI are preserved and the new UI is only visible under the `/ui/v2` route, where React is mounted.

## Backend
As with previous PRs in this migration, the new controllers use JSON: incoming request keys are transformed to snake_case and response keys are converted to camelCase. The index and show responses differ slightly to avoid sending data the UI doesn't use. The more significant changes are listed below.

### Import Jobs Controller
1. Enhanced permissions - Previously, non-admins saw only their own import jobs in the list but they could still view the details of other users' import jobs and the digital object imports belonging to those jobs.
2. Separated the validate-only and submit actions - Previously the mode was determined by a param (e.g. `params[:submit] == 'validate'`); they are now two distinct endpoints. This also fits the new `restoreArchivedS3ObjectsForNewAssets` value, which isn't passed for the validate-only action.
3. Split queue status out of the index action - Import job queue data now lives in its own endpoint so it can be cached and retrieved independently. Since the index action is paginated, keeping the two coupled meant fetching the queue status more often than necessary.

### Digital Object Imports Controller
1. Scoped lookups to the parent import job - Retrieving a digital object import now uses the import job ID as well, rather than the import's ID alone.
2. Enhanced permissions - Before returning a digital object import, we now also verify that the user can access its associated import job.

### Pagination
Both index actions use Kaminari to build the `pagination_data` object.

### New routes
1. `ui/v2/import-jobs` - renders a list of all import jobs (for admins) or the import jobs created by the current non-admin user.
2. `ui/v2/import-jobs/new` - renders a form to upload a CSV file with the option to either submit or validate it.
3. `ui/v2/import-jobs/:importJobId` - renders import job details, CSV download links and counts of pending/successful/failed row imports, with a link to view all object imports.
    1. `ui/v2/import-jobs/:importJobId/digital-object-imports` - displays a paginated list of all digital object imports for the import job, with the option to filter by status.
    2. `ui/v2/import-jobs/:importJobId/digital-object-imports/:digitalObjectImportId` - displays digital object import details and lists validation errors, if present.

### Notable changes
1. Modified the API client to handle file uploads via form data (previously it handled JSON only) and added an `ApiError` class (extends `Error`) that carries the response status and JSON body for better type handling.
2. The digital object import view now uses the Monaco editor in read-only mode to display the digital object data (previously rendered as a plain string) and includes a copy-to-clipboard button.
3. Added optional server-side pagination to the reusable `TableBuilder` component. Because sorting would only apply to the current page, I disabled it for paginated tables as a temporary workaround. We can discuss later whether any columns on the import jobs or digital object imports pages should be sortable.
4. The CSV download helpers don't use the shared API client, since they return files rather than JSON. Downloads currently happen on a single page, so it's best to keep the API client limited to JSON and form data.

### Prettier
I included a `settings.json` with the Prettier configuration we use for ATC. Formatting the entire repo would add too many changes to an already large PR so I haven't formatted any files outside of those relevant to the migration.

### Tests
Tests will be added separately once the API and UI are finalized.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -177,6 +177,8 @@ Rails.application.routes.draw do
       resources :publish_targets, only: [:index, :show, :create, :update, :destroy], param: :string_key
 
       resources :xml_datastreams, only: [:index, :show, :create, :update], param: :string_key
+
+      resources :import_jobs, only: [:create]
     end
   end
 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -178,7 +178,7 @@ Rails.application.routes.draw do
 
       resources :xml_datastreams, only: [:index, :show, :create, :update], param: :string_key
 
-      resources :import_jobs, only: [:create]
+      resources :import_jobs, only: [:index, :create, :show], param: :id
     end
   end
 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -185,6 +185,11 @@ Rails.application.routes.draw do
           get 'queue_activity'
         end
 
+        member do
+          get 'download_original_csv'
+          get 'download_csv_without_successful_rows'
+        end
+
         resources :digital_object_imports, only: [:index, :show]
       end
     end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -178,9 +178,10 @@ Rails.application.routes.draw do
 
       resources :xml_datastreams, only: [:index, :show, :create, :update], param: :string_key
 
-      resources :import_jobs, only: [:index, :create, :show], param: :id do
+      resources :import_jobs, only: [:index, :create, :show, :destroy], param: :id do
         collection do
           post 'validate'
+          get 'queue_activity'
         end
       end
     end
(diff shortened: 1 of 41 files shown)


config/routes.rb
```diff
@@ -44,6 +44,7 @@
     resources :digital_object_imports, only: [:index, :show]
   end
 
+  # ? Should this be deleted since it's the same resource as the one nested under import_jobs?
```
elohanlon, line 47: I think you're right!  I'll ticket this and we can address separately.  Good catch!

elohanlon: Ticketed in HYACINTH-1212.

elohanlon (approved):
Looks great! 🎉

PR #493, opened 2026-07-02
HYACINTH-1209: CSV Export Jobs - new API and UI

wtomaszewska:
# Ticket [HYACINTH-1209](https://columbiauniversitylibraries.atlassian.net/browse/HYACINTH-1209)

## Overview
This PR migrates the Export Jobs feature to React and adds a new API v2 controller with updated actions.

### Backend
The export jobs controller was simplified to only the `index`, `destroy` and `download` actions used by the new UI. A `create` action was also added but is not used by the new UI - the actual export is triggered from a non-React page that still uses the old controller.

### Frontend
The export jobs feature has a single route (`ui/v2/export-jobs`) that displays a paginated list of all CSV exports, with options to download and delete each entry.

Improvements:
1. Since this is the second table using server-side pagination, the logic for reading the current page, managing pagination state, and building the pagination change handler was the same across both. It's now extracted into a shared `useTablePagination` hook, used by both the import and export job tables.
2. The delete confirmation modal was similar to the one used in the import jobs table, so it's been refactored into a shared `ConfirmDeleteModal` component that works across different resources.


elohanlon (approved):
🎉

